Issue #9, opened 2019-12-13
Circuit breaker logic


jngz-es:
Add circuit breaker logic in AnomalyResultTransportAction and RCFResultTransportAction. Currently we don't control ThresholdResultTransportAction due to few memory consumption.

Add UTs.

PR #10, opened 2019-12-13
Circuit breaker

jngz-es:
#9

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/breaker/ADCircuitBreakerService.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/breaker/ADCircuitBreakerService.java
@@ -77,4 +77,14 @@ public class ADCircuitBreakerService {
 
         return this;
     }
+
+    public Boolean isOpen() {
+        for (CircuitBreaker breaker : breakers.values()) {
+            if (breaker.isOpen()) {
+                return true;
+            }
+        }
+
+        return false;
+    }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/breaker/ADCircuitBreakerService.java
```diff
@@ -77,4 +77,14 @@ public ADCircuitBreakerService init() {
 
         return this;
     }
+
+    public Boolean isOpen() {
+        for (CircuitBreaker breaker : breakers.values()) {
```
wnbts: Suggestion. This can also be written as `stream().anyMatch(CircuitBreaker::isOpen)` to be clearer.

src/main/java/com/amazon/opendistroforelasticsearch/ad/breaker/ADCircuitBreakerService.java
```diff
@@ -77,4 +77,14 @@ public ADCircuitBreakerService init() {
 
         return this;
     }
+
+    public Boolean isOpen() {
```
wnbts: Suggestion. This can also be written as `stream().anyMatch(CircuitBreaker::isOpen)` to be shorter and clearer.

jngz-es: Good suggestion, will rewrite it in next pr. Thanks Lai!

PR #39, opened 2020-01-30
adjust preview configuration for more data

wnbts:
Currently, for preview requests, up to 60 data points are sampled and up to 200 results are returned to the client to reduce latency from processing and network. To improve visualization for human eyes, the changes increase the number of samples to match the requested time range, the upper limit of samples, and the number of returned results to clients.

preview

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -64,6 +64,7 @@ public class FeatureManager {
     private final int shingleSize;
     private final int maxMissingPoints;
     private final int maxNeighborDistance;
+    private final double previewSampleRate;
     private final int maxPreviewSamples;
     private final Duration featureBufferTtl;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -317,7 +321,8 @@ public void getPreviewFeatures(AnomalyDetector detector, long startMilli, long e
         long end = truncateToMinute(endMilli);
         long bucketSize = ((IntervalTimeConfiguration) detector.getDetectionInterval()).toDuration().toMillis();
         int numBuckets = (int) Math.floor((end - start) / (double) bucketSize);
-        int stride = (int) Math.max(1, Math.floor((double) numBuckets / maxPreviewSamples));
+        int numSamples = (int) Math.min(numBuckets * previewSampleRate, maxPreviewSamples);
```
jngz-es: Suggestion, 
int numSamples = Math.max(1, (int) Math.min(numBuckets * previewSampleRate, maxPreviewSamples));
to avoid numSamples being 0.

wnbts: Changed. Thank you!

Issue #33, opened 2020-01-27
Do not start a new detection before previous detection finishes


zhanghg08:
Since currently anomaly-detection cannot abort query in elasticsearch , we don't want to start multiple queries at the same time. This may eventually exhaust resources within the system.

PR #40, opened 2020-01-30
Adding negative cache to throttle extra request https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/33

zhanghg08:
Description:
Adding negative cache to throttle extra request https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/33

Updated on Feb 05:
* Re-structured the code for a better dependency injection
* Now there is a new class Throttler handle the negative cache stuff.
* ClientUtil will inject the dependency of Throttler instead of ADStateManager.

Test:
* ./gradlew build
* Created a single node domain with NAB data. Created one AD detector and trigger the run. Verified the negative cache work as expected.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/HourlyCron.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/HourlyCron.java
@@ -43,6 +43,8 @@ public class HourlyCron implements Runnable {
     public void run() {
         DiscoveryNode[] dataNodes = clusterService.state().nodes().getDataNodes().values().toArray(DiscoveryNode.class);
 
+        // we also add the cancel query function here based on query text from the negative cache.
+
         CronRequest modelDeleteRequest = new CronRequest(dataNodes);
         client.execute(CronAction.INSTANCE, modelDeleteRequest, ActionListener.wrap(response -> {
             if (response.hasFailures()) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -35,7 +35,6 @@ import java.util.stream.Stream;
 
 import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
 import com.amazon.opendistroforelasticsearch.ad.model.IntervalTimeConfiguration;
-import com.amazon.opendistroforelasticsearch.ad.transport.ADStateManager;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.action.ActionListener;
@@ -116,16 +115,15 @@ public class FeatureManager {
      * @param detector anomaly detector for which the features are returned
      * @param startTime start time of the data point in epoch milliseconds
      * @param endTime end time of the data point in epoch milliseconds
-     * @param stateManager ADStateManager
      * @return unprocessed features and processed features for the current data point
      */
     @Deprecated
-    public SinglePointFeatures getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime, ADStateManager stateManager) {
+    public SinglePointFeatures getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime) {
         double[][] currentPoints = null;
         Deque<Entry<Long, double[]>> shingle = detectorIdsToTimeShingles
             .computeIfAbsent(detector.getDetectorId(), id -> new ArrayDeque<Entry<Long, double[]>>(shingleSize));
         if (shingle.isEmpty() || shingle.getLast().getKey() < endTime) {
-            Optional<double[]> point = searchFeatureDao.getFeaturesForPeriod(detector, startTime, endTime, stateManager);
+            Optional<double[]> point = searchFeatureDao.getFeaturesForPeriod(detector, startTime, endTime);
             if (point.isPresent()) {
                 if (shingle.size() == shingleSize) {
                     shingle.remove();
@@ -176,16 +174,13 @@ public class FeatureManager {
      * in dimension via shingling.
      *
      * @param detector contains data info (indices, documents, etc)
-     * @param stateManager ADStateManager
      * @return data for cold-start training, or empty if unavailable
      */
     @Deprecated
-    public Optional<double[][]> getColdStartData(AnomalyDetector detector, ADStateManager stateManager) {
+    public Optional<double[][]> getColdStartData(AnomalyDetector detector) {
         return searchFeatureDao
             .getLatestDataTime(detector)
-            .flatMap(
-                latest -> searchFeatureDao.getFeaturesForSampledPeriods(detector, maxTrainSamples, maxSampleStride, latest, stateManager)
-            )
+            .flatMap(latest -> searchFeatureDao.getFeaturesForSampledPeriods(detector, maxTrainSamples, maxSampleStride, latest))
             .map(
                 samples -> transpose(
                     interpolator.interpolate(transpose(samples.getKey()), samples.getValue() * (samples.getKey().length - 1) + 1)
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -35,6 +35,7 @@ import java.util.stream.Stream;
 
 import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
 import com.amazon.opendistroforelasticsearch.ad.model.IntervalTimeConfiguration;
+import com.amazon.opendistroforelasticsearch.ad.transport.ADStateManager;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.action.ActionListener;
@@ -115,15 +116,16 @@ public class FeatureManager {
      * @param detector anomaly detector for which the features are returned
      * @param startTime start time of the data point in epoch milliseconds
      * @param endTime end time of the data point in epoch milliseconds
+     * @param stateManager ADStateManager
      * @return unprocessed features and processed features for the current data point
      */
     @Deprecated
-    public SinglePointFeatures getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime) {
+    public SinglePointFeatures getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime, ADStateManager stateManager) {
         double[][] currentPoints = null;
         Deque<Entry<Long, double[]>> shingle = detectorIdsToTimeShingles
             .computeIfAbsent(detector.getDetectorId(), id -> new ArrayDeque<Entry<Long, double[]>>(shingleSize));
         if (shingle.isEmpty() || shingle.getLast().getKey() < endTime) {
-            Optional<double[]> point = searchFeatureDao.getFeaturesForPeriod(detector, startTime, endTime);
+            Optional<double[]> point = searchFeatureDao.getFeaturesForPeriod(detector, startTime, endTime, stateManager);
             if (point.isPresent()) {
                 if (shingle.size() == shingleSize) {
                     shingle.remove();
@@ -174,13 +176,16 @@ public class FeatureManager {
      * in dimension via shingling.
      *
      * @param detector contains data info (indices, documents, etc)
+     * @param stateManager ADStateManager
      * @return data for cold-start training, or empty if unavailable
      */
     @Deprecated
-    public Optional<double[][]> getColdStartData(AnomalyDetector detector) {
+    public Optional<double[][]> getColdStartData(AnomalyDetector detector, ADStateManager stateManager) {
         return searchFeatureDao
             .getLatestDataTime(detector)
-            .flatMap(latest -> searchFeatureDao.getFeaturesForSampledPeriods(detector, maxTrainSamples, maxSampleStride, latest))
+            .flatMap(
+                latest -> searchFeatureDao.getFeaturesForSampledPeriods(detector, maxTrainSamples, maxSampleStride, latest, stateManager)
+            )
             .map(
                 samples -> transpose(
                     interpolator.interpolate(transpose(samples.getKey()), samples.getValue() * (samples.getKey().length - 1) + 1)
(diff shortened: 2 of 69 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/HourlyCron.java
```diff
@@ -43,6 +43,8 @@ public HourlyCron(ClusterService clusterService, Client client) {
     public void run() {
         DiscoveryNode[] dataNodes = clusterService.state().nodes().getDataNodes().values().toArray(DiscoveryNode.class);
 
+        // we also add the cancel query function here based on query text from the negative cache.
```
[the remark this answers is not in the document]
zhanghg08: Good catch. That's true. Currently anomaly-detection cannot abort running query in elasticsearch. If a query running longer than expected, anomaly-detection will not wait for that even though the query is still running. To solve this issue, we will 1) stop accepting new query if this case happen which is [#33 ](https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/33) 2) daily cron clean up running query using es task management API(https://www.elastic.co/guide/en/elasticsearch/reference/current/tasks.html)

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -115,15 +116,16 @@ public FeatureManager(
      * @param detector anomaly detector for which the features are returned
      * @param startTime start time of the data point in epoch milliseconds
      * @param endTime end time of the data point in epoch milliseconds
+     * @param stateManager ADStateManager
      * @return unprocessed features and processed features for the current data point
      */
     @Deprecated
-    public SinglePointFeatures getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime) {
+    public SinglePointFeatures getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime, ADStateManager stateManager) {
```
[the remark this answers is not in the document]
ylwu-amzn: Agree. We should not pass state manager around.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -249,6 +251,12 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
                 return;
             }
             AnomalyDetector anomalyDetector = detector.get();
+            Optional<Entry<SearchRequest, Instant>> queryEntry = stateManager.getFilteredQuery(anomalyDetector);
+            if (queryEntry.isPresent()) {
+                LOG.info("There is one query running for detectorId: {}", anomalyDetector.getDetectorId());
+                listener.onResponse(new AnomalyResultResponse(Double.NaN, Double.NaN, new ArrayList<FeatureData>()));
```
ylwu-amzn: Client can't distinguish this case with "No data in current detection window"(line 295) as both return empty result. How about we throw an exception like line243

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -249,6 +251,12 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
                 return;
             }
             AnomalyDetector anomalyDetector = detector.get();
+            Optional<Entry<SearchRequest, Instant>> queryEntry = stateManager.getFilteredQuery(anomalyDetector);
+            if (queryEntry.isPresent()) {
+                LOG.info("There is one query running for detectorId: {}", anomalyDetector.getDetectorId());
```
ylwu-amzn: If a query stuck for a long time, that will impact AD realtime detection. How about we cancel the stuck/running query and run the new coming query?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
zhanghg08: Good catch! Yes, we should throw exception in this case. Will update in next revision.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -119,6 +124,47 @@ public int getPartitionNumber(String adID) throws InterruptedException {
...
+     * Insert the negative cache entry for given detector
+     * If detectorId is null, do nothing
+     * @param detector AnomalyDetector
+     * @param searchRequest ES search request
+     */
+    public void insertFilteredQuery(AnomalyDetector detector, SearchRequest searchRequest) {
+        if (detector.getDetectorId() == null) {
+            return;
+        }
+        negativeCache.putIfAbsent(detector.getDetectorId(), new SimpleEntry<>(searchRequest, clock.instant()));
```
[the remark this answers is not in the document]
zhanghg08: Yes, I think put is better than putIfAbsent here. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -209,4 +209,13 @@ public void addPressure(String nodeId) {
     public void resetBackpressureCounter(String nodeId) {
         backpressureMuter.remove(nodeId);
     }
+
+    /**
+     * Check if there is running query on given detector
+     * @param detector Anomaly Detector
+     * @return boolean
```
wnbts: Minor. The doc is incomplete.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -114,7 +114,9 @@ public SearchFeatureDao(
     }
 
     /**
-     * Gets features for the given time period.
+     * Gets features for the given time period. This function also add given detector to negative cache before sending es request.
```
wnbts: Minor. Adds.

Also, use descriptive language instead of prescriptive. See [java doc](https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html).

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -152,4 +158,62 @@ public ClientUtil(Settings setting, Client client) {
...
+     * @throws IllegalStateException when the waiting thread is interrupted
+     */
+    public <Request extends ActionRequest, Response extends ActionResponse> Optional<Response> throttledTimedRequest(
+        Request request,
+        Logger LOG,
+        BiConsumer<Request, ActionListener<Response>> consumer,
+        AnomalyDetector detector
+    ) {
+        try {
+            throttler.insertFilteredQuery(detector, request);
```
wnbts: What should happen if the detector is already in cache?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -152,4 +158,62 @@ public ClientUtil(Settings setting, Client client) {
...
+        }
+    }
+
+    /**
+     * Check if there is running query on given detector
+     * @param detector Anomaly Detector
+     * @return boolean
+     */
+    public boolean hasRunningQuery(AnomalyDetector detector) {
+        Optional<Map.Entry<ActionRequest, Instant>> queryEntry = throttler.getFilteredQuery(detector);
```
wnbts: Minor. The method can be simplified to one line. ```return throttler.getFilteredQuery(detector).isPresent()```

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.time.Clock;
+import java.time.Instant;
+import java.util.AbstractMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import org.elasticsearch.action.ActionRequest;
+
+public class Throttler {
```
wnbts: Minor. The class java doc is missing.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Get negative cache value(ActionRequest, Instant) for given detector
+     * @param detector AnomalyDetector
+     * @return negative cache value(ActionRequest, Instant)
+     */
+    public Optional<Map.Entry<ActionRequest, Instant>> getFilteredQuery(AnomalyDetector detector) {
+        if (negativeCache.containsKey(detector.getDetectorId())) {
+            return Optional.of(negativeCache.get(detector.getDetectorId()));
```
wnbts: MInor. The whole method can be simplified to this line  ```return Optional.of(negativeCache.get(detector.getDetectorId()));```

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Insert the negative cache entry for given detector
+     * If detectorId is null, do nothing
```
wnbts: Minor. The java doc is outdated.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: It won't happen here since if a detector is already in cache, we will throw exception in AnomalyResultTransportAction

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
zhanghg08: will update, thanks for the suggestion

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
zhanghg08: will update

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: will update

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
zhanghg08: will update

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
zhanghg08: will update

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
zhanghg08: will update

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
ylwu-amzn: Should be `Optional.ofNullable`

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/Throttler.java
```diff
@@ -41,15 +44,11 @@ public Throttler(Clock clock) {
      * @return negative cache value(ActionRequest, Instant)
      */
     public Optional<Map.Entry<ActionRequest, Instant>> getFilteredQuery(AnomalyDetector detector) {
-        if (negativeCache.containsKey(detector.getDetectorId())) {
-            return Optional.of(negativeCache.get(detector.getDetectorId()));
-        }
-        return Optional.empty();
+        return Optional.of(negativeCache.get(detector.getDetectorId()));
```
ylwu-amzn: `Optional.ofNullable`

zhanghg08: will update

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
zhanghg08: You are correct. Canceling the stuck query whenever new query comes is one way. However it can be expensive since right now we can only use string compare(query.getDescription()) to find which query to kill. Also killing query using task management API is not effective instantly. It will only happen when moving to next segment.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -152,4 +157,62 @@ public ClientUtil(Settings setting, Client client) {
...
+        Request request,
+        Logger LOG,
+        BiConsumer<Request, ActionListener<Response>> consumer,
+        AnomalyDetector detector
+    ) {
+        try {
+            // if key already exist, reject the request and throws exception
+            if (!throttler.insertFilteredQuery(detector.getDetectorId(), request)) {
+                LOG.error("There is one query running for detectorId: {}", detector.getDetectorId());
+                throw new EndRunException(detector.getDetectorId(), "There is one query running on AnomalyDetector", true);
```
wnbts: MInor. This exception should be documented for client.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -152,4 +157,62 @@ public ClientUtil(Settings setting, Client client) {
...
+                LOG.error("Cannot get response for request {}, error: {}", request, exception);
+            }), latch));
+
+            if (!latch.await(requestTimeout.getSeconds(), TimeUnit.SECONDS)) {
+                throw new ElasticsearchTimeoutException("Cannot get response within time limit: " + request.toString());
+            }
+            return Optional.ofNullable(respReference.get());
+        } catch (InterruptedException e1) {
+            LOG.error(CommonErrorMessages.WAIT_ERR_MSG);
+            throw new IllegalStateException(e1);
```
ylwu-amzn: If exception happens, should clear negative cache for this detector

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: added.

PR #45, opened 2020-02-20
Send back error response when failing to stop detector

kaituo:
Previously, a stop detector api request hangs until timeout if there is a failure.

Testing done:
* verified the issue is fixed

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestDeleteAnomalyDetectorAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestDeleteAnomalyDetectorAction.java
@@ -133,6 +133,7 @@ public class RestDeleteAnomalyDetectorAction extends BaseRestHandler {
             @Override
             public void onFailure(Exception e) {
                 logger.error("Failed to delete AD model for detector " + detectorId, e);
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestDeleteAnomalyDetectorAction.java
```diff
@@ -133,6 +133,7 @@ public void onResponse(StopDetectorResponse stopDetectorResponse) {
             @Override
             public void onFailure(Exception e) {
                 logger.error("Failed to delete AD model for detector " + detectorId, e);
+                channel.sendResponse(new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, "Failed to delete AD model"));
```
ylwu-amzn: Thanks for this fix.
We will add AD job and the stop detector API will not be called when delete monitor. Have moved stop detector logic to stop AD job API in this PR: https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/44

kaituo: I see.  Will close this PR and comment on your PR.

PR #54, opened 2020-03-05
Cancel query if given detector already have one

zhanghg08:
*Issue #55, if available:*
Cancel running query if given detector already has one running.

*Description of changes:*
1) ./gradlew build
2) start my own cluster and tested with a 20sec delay query as long running query. If one detector is running a query, a second query request will cancel the previous one and start itself.


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -199,7 +199,7 @@ private static Void initGson() {
         Settings settings = environment.settings();
```
kaituo: Is there a unit/integration test for the cancel mechanism?  If not, I strongly suggest we add one. 

You can add an ESIntegTestCase where we
1) define a SearchOperationListener such that index operations are delayed to simulate long running queries;
2) create a fake plugin to use listener defined in 1)
3) add AD plugin and the fake plugin together
4) ... automate what you did on manual testing ..

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -162,11 +180,11 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
      * Send a nonblocking request with a timeout and return response. The request will first be put into
      * the negative cache. Once the request complete, it will be removed from the negative cache.
```
ylwu-amzn: Please add more description about cancel request process

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -179,32 +197,36 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
             } catch (Exception e) {
-                LOG.error("Failed to process the request for detectorId: {}.", detector.getDetectorId());
-                throttler.clearFilteredQuery(detector.getDetectorId());
+                LOG.error("Failed to process the request for detectorId: {}.", detectorId);
+                throttler.clearFilteredQuery(detectorId);
                 throw e;
             }
 
             if (!latch.await(requestTimeout.getSeconds(), TimeUnit.SECONDS)) {
+
```
ylwu-amzn: Remove empty line

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -222,4 +244,80 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+        }
+        // case 1: given detectorId is not in current task list
+        if (matchedTask == null) {
+            // log and then clear negative cache
+            LOG.info("Couldn't find task for detectorId: {}. Clean this entry from Throttler", detectorId);
+            throttler.clearFilteredQuery(detectorId);
+            return;
+        }
+        // case 2: we can find the task for given detectorId
+        TaskId parentTaskId = matchedTask.getParentTaskId().isSet() ? matchedTask.getParentTaskId() : matchedTask.getTaskId();
```
ylwu-amzn: Is it possible the parent task has parent too?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -222,4 +244,80 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+     * Helper function to handle CancelTasksResponse
+     * @param cancelTasksResponse CancelTasksResponse
+     * @param detectorId Anomaly Detector Id
+     * @param LOG Logger
+     */
+    private void onCancelTaskResponse(CancelTasksResponse cancelTasksResponse, String detectorId, Logger LOG) {
+        List<ElasticsearchException> nodeFailures = cancelTasksResponse.getNodeFailures();
+        List<TaskOperationFailure> taskFailures = cancelTasksResponse.getTaskFailures();
+        if (nodeFailures.isEmpty() && taskFailures.isEmpty()) {
+            LOG.info("Cancelling query for detectorId: {} succeeds. Clear entry from Throttler", detectorId);
```
ylwu-amzn: Better to add some retry for these failed tasks.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -179,32 +197,36 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+                cancelRunningQuery(client, detectorId, LOG);
             }
             AtomicReference<Response> respReference = new AtomicReference<>();
             final CountDownLatch latch = new CountDownLatch(1);
 
-            try {
+            try (ThreadContext.StoredContext context = threadPool.getThreadContext().stashContext()) {
+                assert context != null;
+                threadPool.getThreadContext().putHeader(Task.X_OPAQUE_ID, CommonName.ANOMALY_DETECTOR + ":" + detectorId);
                 consumer.accept(request, new LatchedActionListener<Response>(ActionListener.wrap(response -> {
```
ylwu-amzn: It's possible the `cancelRunningQuery` in progress or fail when start a new request. If `cancelRunningQuery` is not time consuming, better to start a new request when we get respond of `cancelRunningQuery`. If it's heavy action, may need to monitor the cancelation status and retry if failed; so we can terminate unnecessary AD query to protect cluster performance. It's ok to add some todo&comments here and refactor it later if you think the change will be big.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -179,32 +197,36 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+                LOG.info("There is one query running for detectorId: {}. Trying to cancel the long running query", detectorId);
+                cancelRunningQuery(client, detectorId, LOG);
             }
             AtomicReference<Response> respReference = new AtomicReference<>();
             final CountDownLatch latch = new CountDownLatch(1);
 
-            try {
+            try (ThreadContext.StoredContext context = threadPool.getThreadContext().stashContext()) {
+                assert context != null;
+                threadPool.getThreadContext().putHeader(Task.X_OPAQUE_ID, CommonName.ANOMALY_DETECTOR + ":" + detectorId);
```
ylwu-amzn: Just asking, will the `X_OPAQUE_ID` header be passed to child tasks?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -222,4 +244,80 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+        if (matchedTask == null) {
+            // log and then clear negative cache
+            LOG.info("Couldn't find task for detectorId: {}. Clean this entry from Throttler", detectorId);
+            throttler.clearFilteredQuery(detectorId);
+            return;
+        }
+        // case 2: we can find the task for given detectorId
+        TaskId parentTaskId = matchedTask.getParentTaskId().isSet() ? matchedTask.getParentTaskId() : matchedTask.getTaskId();
+        CancelTasksRequest cancelTaskRequest = new CancelTasksRequest();
+        cancelTaskRequest.setParentTaskId(parentTaskId);
```
ylwu-amzn: From line292, if `matchedTask.getParentTaskId().isSet()` is false, will get `matchedTask.getTaskId()`  as `parentTaskId`.  For this case, `cancelTaskRequest.setParentTaskId(parentTaskId)` will cancel tasks which has parent task id as `matchedTask.getTaskId()`. Is it possible the `matchedTask` has no child tasks? If it's possible, will `cancelTaskRequest.setParentTaskId(parentTaskId)` throw exception or cancel `matchedTask` ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -222,4 +244,80 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+     */
+    private void onCancelTaskResponse(CancelTasksResponse cancelTasksResponse, String detectorId, Logger LOG) {
+        List<ElasticsearchException> nodeFailures = cancelTasksResponse.getNodeFailures();
+        List<TaskOperationFailure> taskFailures = cancelTasksResponse.getTaskFailures();
+        if (nodeFailures.isEmpty() && taskFailures.isEmpty()) {
+            LOG.info("Cancelling query for detectorId: {} succeeds. Clear entry from Throttler", detectorId);
+            throttler.clearFilteredQuery(detectorId);
+            return;
+        }
+        throw new InternalFailure(detectorId, "Failed to cancel current tasks due to node or task failures");
```
ylwu-amzn: log failures?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -222,4 +244,80 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+     * @param listTasksResponse ListTasksResponse
+     * @param detectorId Anomaly Detector Id
+     * @param LOG Logger
+     */
+    private void onListTaskResponse(ListTasksResponse listTasksResponse, String detectorId, Logger LOG) {
+        List<TaskInfo> tasks = listTasksResponse.getTasks();
+        TaskInfo matchedTask = null;
+        for (TaskInfo task : tasks) {
+            if (!task.getHeaders().isEmpty() && task.getHeaders().get(Task.X_OPAQUE_ID) != null) {
+                if (task.getHeaders().get(Task.X_OPAQUE_ID).contains(detectorId)) {
```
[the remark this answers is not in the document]
zhanghg08: Thanks for the suggestion, will change it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -179,32 +197,36 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
+
         try {
-            // if key already exist, reject the request and throws exception
-            if (!throttler.insertFilteredQuery(detector.getDetectorId(), request)) {
-                LOG.error("There is one query running for detectorId: {}", detector.getDetectorId());
-                throw new EndRunException(detector.getDetectorId(), "There is one query running on AnomalyDetector", true);
+            String detectorId = detector.getDetectorId();
+            if (!throttler.insertFilteredQuery(detectorId, request)) {
+                LOG.info("There is one query running for detectorId: {}. Trying to cancel the long running query", detectorId);
+                cancelRunningQuery(client, detectorId, LOG);
```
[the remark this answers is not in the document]
zhanghg08: Agree. Return after cancelling will be more safer. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
```diff
@@ -222,4 +244,80 @@ public ClientUtil(Settings setting, Client client, Throttler throttler) {
...
     }
+
+    /**
+     * Cancel long running query for given detectorId
+     * @param client Elasticsearch client
+     * @param detectorId Anomaly Detector Id
+     * @param LOG Logger
+     */
+    private void cancelRunningQuery(Client client, String detectorId, Logger LOG) {
+        ListTasksRequest listTasksRequest = new ListTasksRequest();
```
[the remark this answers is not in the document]
zhanghg08: Thanks for the advice. For the group_by=parents, it's a little weird. For the api it supports this feature(https://www.elastic.co/guide/en/elasticsearch/reference/current/tasks.html#_task_grouping). However when it comes to java api, it's not supported as far as I see.

For actions=*search, I will added it as "actions=*search*" since I can see some child query has something like "indices:data/read/search[phase/query]"

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: added.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: For our search query, there is only two-level parent-child relationship.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: Will add a todo comment for now.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: I think this is similar with Kaituo's comments. For safety concern, I will not start new request if we need to cancel the running one, just in case the cancel failed somehow and we keep adding new requests. We can revisit it later.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: Yes, it will. I can see both parent and children tasks have the same header from the manual test.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
zhanghg08: I got your point. To avoid this corner case, I will go through the entire tasks list(previously it will early terminate once found matched). If there is only one task(no parent), we need to setTaskId, otherwise setParentTaskId

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
zhanghg08: This is also one of my concern. I noticed we don't have any unit test for clientUtil and tried to add one but it's too complicated. When manual testing, I use similar listener which will delay the search to make up the long running query. Not sure if that can be done in integration test, I will sync up with you offline.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ClientUtil.java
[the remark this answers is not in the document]
kaituo: you meant we need to use "actions=*search*", right?  Yes, please do that.  Your current code uses "*search".

PR #56, opened 2020-03-06
Add threadpool prefix and change threadpool name

kaituo:
When we run AD and SQL plugin together, ES fails to start with registering a setting twice error. AD's threadpool setting prefix is null. Since SQL plugin uses null as well, we would have a register setting conflict because we cannot register setting [null.size] twice. The fix is to give our threadpool a setting name prefix instead of null

Also, this threadpool would be used by other components besides AD job.  Rename it to make it not ad job specific.

Testing done:
- run all unit and integration tests
- ES can start with AD and SQL plugin enabled

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -128,7 +128,7 @@ import com.amazon.opendistroforelasticsearch.ad.dataprocessor.LinearUniformInter
 import com.amazon.opendistroforelasticsearch.ad.dataprocessor.SingleFeatureLinearUniformInterpolator;
 import com.amazon.randomcutforest.serialize.RandomCutForestSerDe;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -346,10 +346,10 @@ private static Void initGson() {
             .singletonList(
                 new FixedExecutorBuilder(
                     settings,
-                    AD_JOB_THREAD_POOL_NAME,
+                    AD_THREAD_POOL_NAME,
                     Math.max(1, EsExecutors.numberOfProcessors(settings) / 4),
-                    AD_JOB_THEAD_POOL_QUEUE_SIZE,
-                    null
+                    AD_THEAD_POOL_QUEUE_SIZE,
+                    "opendistro.ad.ad_thread_pool"
```
ylwu-amzn: How about we use "opendistro.ad."+ AD_THREAD_POOL_NAME ? So we can know exactly which threadpool it maps to.

kaituo: good idea.  Changed.

Issue #64, opened 2020-03-16
Preview says not enough data points even though we do


kaituo:
The command

```
POST /_opendistro/_anomaly_detection/detectors/C53sinABEBZckSLeM76O/_preview
{
    "period_start":1582447622314,
    "period_end":1582879622315
}
```

returns empty results:

But we have enough data (1440 data points) in the range:

```
GET /nab_art_daily_jumpsup/_doc/_search?pretty
{
    "query": {
        "range" : {
            "timestamp" : {
                "gte" : 1582447622314,
                "lte" : 1582879622315
            }
        }
    }
}

{
  "took" : 3,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 1440,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "nab_art_daily_jumpsup",
        "_type" : "_doc",
        "_id" : "s53jinABEBZckSLeIa54",
        "_score" : 1.0,
        "_source" : {
          "timestamp" : 1582448503867,
          "value" : 19.3434172634
        }
      },
      {
        "_index" : "nab_art_daily_jumpsup",
        "_type" : "_doc",
        "_id" : "tp3jinABEBZckSLeIa54",
        "_score" : 1.0,
        "_source" : {
          "timestamp" : 1582449403867,
          "value" : 18.1971208058
        }
      },
      {
        "_index" : "nab_art_daily_jumpsup",
        "_type" : "_doc",
        "_id" : "vJ3jinABEBZckSLeIa54",
        "_score" : 1.0,
        "_source" : {
          "timestamp" : 1582451203867,
          "value" : 67.7731278131
        }
      },
      {
        "_index" : "nab_art_daily_jumpsup",
        "_type" : "_doc",
        "_id" : "vZ3jinABEBZckSLeIa54",
        "_score" : 1.0,
        "_source" : {
          "timestamp" : 1582451503867,
          "value" : 69.0056563819
        }
      },
      {
        "_index" : "nab_art_daily_jumpsup",
        "_type" : "_doc",
        "_id" : "wJ3jinABEBZckSLeIa54",
        "_score" : 1.0,
        "_source" : {
          "timestamp" : 1582452403867,
          "value" : 68.9065226554
        }
      },
      {
[...]

Issue #63, opened 2020-03-16
Performance comparison of multisearch and date range bucket aggregation


kaituo:
Performance comparison of multisearch and date range bucket aggregation
Currently, we use multisearch in SearchFeatureDao

kaituo:
running on two benchmarks with multisearch and time range query.  date_range query is 2~10 times faster than multisearch.  Every time I ran a query, I restarted ec2 nodes to invalidate OS and shard cache.

date_range query:

```
POST nab_art_daily_jumpsup/_search
 {
   "size": 0,
   "aggs": {
     "range": {
       "date_range": { 
         "field": "timestamp",
         "format": "epoch_millis",
         "ranges": [
           {"from" : 1583391275000, "to": 1583391575000},
           {"from" : 1583392475000, "to": 1583392775000},
           ...
        },
       "aggs": {
         "avg_value": {
           "sum": {
             "field": "value"
           }
         }
       }
     }
   }
 } 
```

multisearch query:

```
GET /nab_art_daily_jumpsup/_msearch?pretty
{"index" : "nab_art_daily_jumpsup"}
{"query":{"bool":{"must":[{"range":{"timestamp":{"from":1583391275000,"to":1583391575000,"include_lower":true,"include_upper":true,"boost":1}}},{"match_all":{"boost":1}}],"adjust_pure_negative":true,"boost":1}},"aggregations":{"7isajm0Bs-gslJuCOan7":{"sum":{"field":"value"}}}}
{"index" : "nab_art_daily_jumpsup"}
...
```

Environment:
* single node cluster
* cluster is not loaded (no other workload besides my queries)

Two benchmarks:
1. nab_art_daily_jumpsup, sparse data, every 5 minutes one matching document.  So each multisearch and date_time query match one doc.  I did 3 runs.  The following is the recorded time (millisecond)

| multisearch | date_time |
|-------------|-----------|
| 844         | 75       |
| 973         | 70       |
| 1040         | 59      |

2. server metrics, dense data, every 5 minutes multiple matching documents (e.g., 105).  So each multisearch and date_time query match multiple docs.  I did 3 runs.  The following is the recorded time (millisecond)

| multisearch | date_time |
|-------------|-----------|
| 663         | 205       |
| 716         | 181       |
| 439         | 224       |

PR #65, opened 2020-03-16
Fix empty preview result due to insufficient sample

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/64

*Description of changes:*
Preview does not use all data in the given time range as it is costly. Previously, we sample data by issuing multiple queries on shard 0 data. The purpose of the shard 0 query restriction is to reduce system costs. The nab_art_daily_jumpsup data set has one doc in each interval, and the doc is spread out in 5 shards. Even though we issue 360 queries, we only get 70~80 samples back by querying shard 0. Together with interpolated data points, the preview run misses significant portions of data required to train models (400 is the minimum) and thus returns empty preview results. This PR fixes the issue by removing the shard 0 search restriction.

Previously, the preview API issues multiple queries encapsulated in a multisearch request (the request can contain 360 search queries at most). The same result could be obtained via a date range query with multiple range buckets. We show a date range query is 2~10 times faster than a multisearch request (https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/63). This PR replaces the multisearch request with a date range query.

This PR also fixes a bug in query generation. We generate aggregation query twice: once with filter query, once separately.  

This PR also removes unused field scriptService in SearchFeatureDao.

Testing done:
- Previous preview unit tests pass.
- Manually verified date range queries results are correctly processed by cross checking intermediate logs.
- Manually verified preview results with multisearch (removing shard 0 search restriction) and date range implementation are the same
- Manually verified preview don't show empty results with the nab_art_daily_jumpsup data set after the fix

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.ad.feature;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayDeque;
 import java.util.HashMap;
@@ -41,10 +42,11 @@ import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.search.aggregations.AggregationBuilders;
+import org.elasticsearch.search.aggregations.Aggregations;
+import org.elasticsearch.search.aggregations.bucket.range.InternalDateRange;
 import org.elasticsearch.search.aggregations.metrics.Max;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.elasticsearch.common.xcontent.NamedXContentRegistry;
-import org.elasticsearch.script.ScriptService;
 import org.elasticsearch.search.aggregations.Aggregation;
 import org.elasticsearch.search.aggregations.metrics.NumericMetricsAggregation.SingleValue;
 import org.elasticsearch.search.aggregations.metrics.InternalTDigestPercentiles;
@@ -66,7 +68,6 @@ public class SearchFeatureDao {
 
     // Dependencies
     private final Client client;
-    private final ScriptService scriptService;
     private final NamedXContentRegistry xContent;
     private final Interpolator interpolator;
     private final ClientUtil clientUtil;
@@ -75,20 +76,12 @@ public class SearchFeatureDao {
      * Constructor injection.
      *
      * @param client ES client for queries
-     * @param scriptService ES ScriptService
      * @param xContent ES XContentRegistry
      * @param interpolator interpolator for missing values
      * @param clientUtil utility for ES client
      */
-    public SearchFeatureDao(
-        Client client,
-        ScriptService scriptService,
-        NamedXContentRegistry xContent,
-        Interpolator interpolator,
-        ClientUtil clientUtil
-    ) {
+    public SearchFeatureDao(Client client, NamedXContentRegistry xContent, Interpolator interpolator, ClientUtil clientUtil) {
         this.client = client;
-        this.scriptService = scriptService;
         this.xContent = xContent;
         this.interpolator = interpolator;
         this.clientUtil = clientUtil;
@@ -207,34 +200,25 @@ public class SearchFeatureDao {
         List<Entry<Long, Long>> ranges,
         ActionListener<List<Optional<double[]>>> listener
     ) {
-        MultiSearchRequest multiSearchRequest = new MultiSearchRequest();
-        ranges
-            .stream()
-            .map(range -> createFeatureSearchRequest(detector, range.getKey(), range.getValue(), Optional.of(FEATURE_SAMPLE_PREFERENCE)))
-            .forEachOrdered(request -> multiSearchRequest.add(request));
+        SearchRequest request = createPreviewSearchRequest(detector, ranges, Optional.empty());
 
-        client
-            .multiSearch(
-                multiSearchRequest,
-                ActionListener
-                    .wrap(
-                        response -> listener
-                            .onResponse(
-                                Optional
-                                    .of(response)
-                                    .map(Stream::of)
-                                    .orElseGet(Stream::empty)
-                                    .flatMap(multiSearchResp -> Arrays.stream(multiSearchResp.getResponses()))
-                                    .map(
-                                        item -> Optional
-                                            .ofNullable(item.getResponse())
-                                            .flatMap(r -> parseResponse(r, detector.getEnabledFeatureIds()))
-                                    )
-                                    .collect(Collectors.toList())
-                            ),
-                        listener::onFailure
-                    )
-            );
+        client.search(request, ActionListener.wrap(response -> {
+            Aggregations aggs = response.getAggregations();
+            if (aggs == null) {
+                listener.onResponse(Collections.emptyList());
+            }
+
+            listener
+                .onResponse(
+                    aggs
+                        .asList()
+                        .stream()
+                        .filter(InternalDateRange.class::isInstance)
+                        .flatMap(agg -> ((InternalDateRange) agg).getBuckets().stream())
+                        .map(bucket -> parseBucket(bucket, detector.getEnabledFeatureIds()))
+                        .collect(Collectors.toList())
+                );
+        }, listener::onFailure));
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -207,34 +200,25 @@ public void getFeatureSamplesForPeriods(
         List<Entry<Long, Long>> ranges,
         ActionListener<List<Optional<double[]>>> listener
     ) {
-        MultiSearchRequest multiSearchRequest = new MultiSearchRequest();
-        ranges
-            .stream()
-            .map(range -> createFeatureSearchRequest(detector, range.getKey(), range.getValue(), Optional.of(FEATURE_SAMPLE_PREFERENCE)))
```
wnbts: Minor. FEATURE_SAMPLE_PREFERENCE can be removed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -336,4 +320,32 @@ private SearchRequest createFeatureSearchRequest(AnomalyDetector detector, long
...
+        try {
+            SearchSourceBuilder searchSourceBuilder = ParseUtils.generatePreviewQuery(detector, ranges, xContent);
+            return new SearchRequest(detector.getIndices().toArray(new String[0]), searchSourceBuilder).preference(preference.orElse(null));
+        } catch (IOException e) {
+            logger.warn("Failed to create feature search request for " + detector + " for preview", e);
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private Optional<double[]> parseBucket(InternalDateRange.Bucket bucket, List<String> featureIds) {
```
wnbts: Minor. Some parts might be refactored with existing response parsing logic.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -336,4 +320,32 @@ private SearchRequest createFeatureSearchRequest(AnomalyDetector detector, long
             throw new IllegalStateException(e);
         }
     }
+
+    private SearchRequest createPreviewSearchRequest(
+        AnomalyDetector detector,
+        List<Entry<Long, Long>> ranges,
+        Optional<String> preference
```
wnbts: Minor. Since this parameter is not used, it can be removed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: removed

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: removed

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: refactored.

kaituo: refactored

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -336,4 +311,30 @@ private SearchRequest createFeatureSearchRequest(AnomalyDetector detector, long
             throw new IllegalStateException(e);
         }
     }
+
+    private SearchRequest createPreviewSearchRequest(AnomalyDetector detector, List<Entry<Long, Long>> ranges) {
+        try {
+            SearchSourceBuilder searchSourceBuilder = ParseUtils.generatePreviewQuery(detector, ranges, xContent);
+            return new SearchRequest(detector.getIndices().toArray(new String[0]), searchSourceBuilder);
+        } catch (IOException e) {
+            logger.warn("Failed to create feature search request for " + detector + " for preview", e);
```
ylwu-amzn: log `detector.getDetectorId()`

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -336,4 +311,30 @@ private SearchRequest createFeatureSearchRequest(AnomalyDetector detector, long
...
+            throw new IllegalStateException(e);
```
ylwu-amzn: Why wrap IOException with IllegalStateException?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -207,34 +191,25 @@ public void getFeatureSamplesForPeriods(
...
+            if (aggs == null) {
+                listener.onResponse(Collections.emptyList());
+            }
+
+            listener
+                .onResponse(
+                    aggs
+                        .asList()
+                        .stream()
+                        .filter(InternalDateRange.class::isInstance)
```
ylwu-amzn: Is it possible ES return different date range class other than `InternalDateRange`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -207,34 +191,25 @@ public void getFeatureSamplesForPeriods(
...
-                                    )
-                                    .collect(Collectors.toList())
-                            ),
-                        listener::onFailure
-                    )
-            );
+        client.search(request, ActionListener.wrap(response -> {
+            Aggregations aggs = response.getAggregations();
+            if (aggs == null) {
+                listener.onResponse(Collections.emptyList());
```
yizheliu-amazon: Should we do `return` here? Otherwise it will continue to line 202

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -336,4 +311,30 @@ private SearchRequest createFeatureSearchRequest(AnomalyDetector detector, long
             throw new IllegalStateException(e);
```
yizheliu-amazon: same as Yaliang's comment.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: for 7.4, it is not possible.  For other versions, it is possible.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: This is Lai's code I guess.  I guess IOException is checked exception and the method and and its callers have to declare the exception before the exception is handled.  Lai prefers to use unchecked exception like IllegalStateException. 

I prefer checked exceptions as they serve as documentation on what a method can throw and remind the caller to handle it.

Do you have any preference on when to use checked and unchecked exceptions?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: good catch.  fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: I am following Lai's style in this class.  He prefers unchecked exception like IllegalStateException. IOException is checked exception and the method and and its callers have to declare the exception before the exception is handled. 

I prefer checked exceptions as they serve as documentation on what a method can throw and remind the caller to handle it.

Do you have any preference on when to use checked and unchecked exceptions?

wnbts: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-exceptions.html

i don't mind the team making their own decisions.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
ylwu-amzn: Got it. So for other version, the `InternalDateRange.class` is still what we want?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
ylwu-amzn: From https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html

> Generally speaking, do not throw a RuntimeException or create a subclass of RuntimeException simply because you don't want to be bothered with specifying the exceptions your methods can throw.
> Here's the bottom line guideline: If a client can reasonably be expected to recover from an exception, make it a checked exception. If a client cannot do anything to recover from the exception, make it an unchecked exception.

For this case, `generateInternalFeatureQuery` method may throw `IOException` which caused by user's wrong query input. It's possible to be recovered from this for user to fix the query. And we just wrap the IOException without adding more info like custom error message; and no special handling logic for this `IllegalStateException `. So prefer to use checked exception here.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: other versions is untested.  Will do that when we release new versions.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: Thanks for your analysis.  fixed it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -184,23 +184,24 @@ protected void processResponse(GetResponse response) throws Exception {
...
                         channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
-                    } catch (IOException e) {
-                        logger.error("Fail to send back exception message" + detector.getDetectorId(), exception);
-                    }
-                }));
+                    }, exception -> {
+                        logger.error("Unexpected error running anomaly detector " + detector.getDetectorId(), exception);
+                        try {
+                            XContentBuilder builder = channel.newBuilder().startObject().field(ANOMALY_DETECTOR, detector).endObject();
+                            channel.sendResponse(new BytesRestResponse(RestStatus.OK, builder));
```
yizheliu-amazon: question: why send OK response if there is an exception while running AD?

kaituo: Good catch.  I changed to RestStatus.INTERNAL_SERVER_ERROR.

Issue #73, opened 2020-03-23
preview exits with null pointer exception


kaituo:
[2020-03-23T00:15:33,597][ERROR][c.a.o.a.r.RestExecuteAnomalyDetectorAction] [9801a7a977f1.ant.amazon.com] Unexpected error running anomaly detector jLajBXEBpozDUoJ-TiW2
java.lang.NullPointerException: null
	at com.amazon.opendistroforelasticsearch.ad.model.AnomalyResult.toXContent(AnomalyResult.java:94) ~[?:?]
	at org.elasticsearch.common.xcontent.XContentBuilder.value(XContentBuilder.java:857) ~[elasticsearch-x-content-7.4.2.jar:7.4.2]
	at org.elasticsearch.common.xcontent.XContentBuilder.value(XContentBuilder.java:850) ~[elasticsearch-x-content-7.4.2.jar:7.4.2]
	at org.elasticsearch.common.xcontent.XContentBuilder.unknownValue(XContentBuilder.java:828) ~[elasticsearch-x-content-7.4.2.jar:7.4.2]
	at org.elasticsearch.common.xcontent.XContentBuilder.value(XContentBuilder.java:915) ~[elasticsearch-x-content-7.4.2.jar:7.4.2]
	at org.elasticsearch.common.xcontent.XContentBuilder.unknownValue(XContentBuilder.java:824) ~[elasticsearch-x-content-7.4.2.jar:7.4.2]
	at org.elasticsearch.common.xcontent.XContentBuilder.value(XContentBuilder.java:804) ~[elasticsearch-x-content-7.4.2.jar:7.4.2]
	at org.elasticsearch.common.xcontent.XContentBuilder.field(XContentBuilder.java:895) ~[elasticsearch-x-content-7.4.2.jar:7.4.2]
	at com.amazon.opendistroforelasticsearch.ad.rest.RestExecuteAnomalyDetectorAction$1.lambda$processResponse$0(RestExecuteAnomalyDetectorAction.java:192) ~[?:?]
	at org.elasticsearch.action.ActionListener$1.onResponse(ActionListener.java:62) [elasticsearch-7.4.2.jar:7.4.2]
	at com.amazon.opendistroforelasticsearch.ad.AnomalyDetectorRunner.lambda$executeDetector$0(AnomalyDetectorRunner.java:67) [opendistro-anomaly-detection-1.4.0.0.jar:1.4.0.0]
[...]

PR #74, opened 2020-03-23
Fix null pointer exception during preview

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/73

*Description of changes:*
Recently, we changed the anomaly result index’s schema and toXContent method by introducing new fields. During the preview, some of the fields in anomaly result index are null. Thus, toXContent fails with a null pointer exception.

This PR fixes the issue by protecting a field’s serialization with a null check.

This PR also removes an unused local variable in ModelManager.

Testing done:
- Manually verified the issue is gone
- gradle build


kaituo:
> unit tests can be used to prevent regressions.

We may need integration test to find these errors as this requires basically all of the preview components to run to expose this bug.  Will create integration test in this package later.

Issue #79, opened 2020-04-07
Error during build: downloading resources\job-scheduler\opendistro-job-scheduler-1.4.0.0-SNAPSHOT.zip


amirmuminovic:
I tried following the instructions to get a running instance of anomaly detection.
I cloned the repository and set the JDK version to be 12.0.2 for environment variable JAVA_HOME.

The build fails with a cryptic message, as follows:

Downloading file://C:\**\anomaly-detection\src\test\resources\job-scheduler\opendistro-job-scheduler-1.4.0.0-SNAPSHOT.zip
Exception in thread "main" java.net.UnknownHostException: C
        at java.base/sun.nio.ch.NioSocketImpl.connect(NioSocketImpl.java:567)
        at java.base/java.net.Socket.connect(Socket.java:585)
        at java.base/sun.net.ftp.impl.FtpClient.doConnect(FtpClient.java:961)
        at java.base/sun.net.ftp.impl.FtpClient.tryConnect(FtpClient.java:923)
        at java.base/sun.net.ftp.impl.FtpClient.connect(FtpClient.java:1018)
        at java.base/sun.net.ftp.impl.FtpClient.connect(FtpClient.java:1004)
        at java.base/sun.net.www.protocol.ftp.FtpURLConnection.connect(FtpURLConnection.java:312)
        at java.base/sun.net.www.protocol.ftp.FtpURLConnection.getInputStream(FtpURLConnection.java:418)
        at org.elasticsearch.plugins.InstallPluginCommand.downloadZip(InstallPluginCommand.java:381)
        at org.elasticsearch.plugins.InstallPluginCommand.download(InstallPluginCommand.java:279)
        at org.elasticsearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:230)
        at org.elasticsearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:217)
        at org.elasticsearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:86)
        at org.elasticsearch.cli.Command.mainWithoutErrorHandling(Command.java:125)
        at org.elasticsearch.cli.MultiCommand.execute(MultiCommand.java:77)
        at org.elasticsearch.cli.Command.mainWithoutErrorHandling(Command.java:125)
        at org.elasticsearch.cli.Command.main(Command.java:90)
        at org.elasticsearch.plugins.PluginCli.main(PluginCli.java:47)

[...]

PR #82, opened 2020-04-09
Added URL for jb_scheduler-plugin_zip instead of local file path

amirmuminovic:
*Issue #79 *

*Description of changes:*
The build process failed on Windows because of a UNIX-specific path. During the build, there would be an attempt to download the file from the local file system which would result in an error, as shown in issue #79 . 

The change included adding a remote URL from which the job-scheduler plugin would be downloaded instead of a local file reference.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ repositories {
 ext {
     opendistroVersion = '1.4.0'
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
-    job_scheduler_plugin_zip = "file://${fileTree("src/test/resources/job-scheduler").getSingleFile().absolutePath}"
+    job_scheduler_plugin_zip = "https://github.com/opendistro-for-elasticsearch/job-scheduler/releases/download/v1.4.0.0/opendistro-job-scheduler-1.4.0.0.zip"
 }
 
 version = "${opendistroVersion}.0"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -44,7 +44,7 @@ repositories {
 ext {
     opendistroVersion = '1.4.0'
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
-    job_scheduler_plugin_zip = "file://${fileTree("src/test/resources/job-scheduler").getSingleFile().absolutePath}"
+    job_scheduler_plugin_zip = "https://github.com/opendistro-for-elasticsearch/job-scheduler/releases/download/v1.4.0.0/opendistro-job-scheduler-1.4.0.0.zip"
```
kaituo, line 47: This adds a dependency on github when we build.  Also, we need to upload a new zip to github in the future when we upgrade job-scheduler.  I suggest detecting windows in gradle and use different paths instead.  See https://stackoverflow.com/questions/11235614/how-to-detect-the-current-os-from-gradle/31443955

amirmuminovic: @kaituo I had a closer inspection of the issue. It was like @wnbts suggested, an error with the path on Windows. My initial solution was treating the symptom, not the root cause of the error.

I did some research and found the issue is that Linux requires two slashes when specifying the path whilst Windows requires three. I have made the adjustment and we should have no additional dependencies after this fix :D

kaituo: Thank you very much for the change.

Issue #78, opened 2020-04-07
Migrate to async method


ylwu-amzn:
## Problem
See some timeout exception when `RCFResultTransportAction` get checkout points:

## Cause
`ElasticsearchTimeoutException[Cannot get response within time limit: get [.opendistro-anomaly-checkpoints][_doc][kay-UXEBjVN7EqCB9SOS_model_rcf_1]: routing [null]]`

Seems [`RCFResultTransportAction`](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/development/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/RCFResultTransportAction.java#L60) is using deprecated method `manager.getRcfResult(request.getAdID(), request.getModelID(), request.getFeatures())` 

## Action
We need to migrate to async method.

Please check if we are using such deprecated methods at other places.

ylwu-amzn:
Same for [ThresholdResultTransportAction](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/1637128473cea66db96b87b20a4c3ff7eee5a150/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ThresholdResultTransportAction.java#L45)

ylwu-amzn:
[ADStateManager](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/development/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java#L131)

`ElasticsearchTimeoutException[Cannot get response within time limit: get [.opendistro-anomaly-detectors][_doc][WQXnUXEB44dtcje-PgWE]: routing [null]]`

PR #83, opened 2020-04-09
Use callbacks and bug fix

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/78

*Description of changes:*
This PR includes the following changes:

1. remove classes that are not needed in jacocoExclusions since we have enough coverage for those classes.
2. Use ClientUtil instead of Elasticsearch’s client in AD job runner
3. Use one function to get the number of partitioned forests. Previously, we have redundant code in both ModelManager and ADStateManager.
4. Change ADStateManager.getAnomalyDetector to use callback.
5. Change AnomalyResultTransportAction to use callback to get features.
6. Add in AnomalyResultTransportAction to handle the case where all features have been disabled, and users' index does not exist.
7. Change get RCF and threshold result methods to use callback and add exception handling of IndexNotFoundException due to the change. Previously, getting RCF and threshold result methods won’t throw IndexNotFoundException.
8. Remove unused fields in StopDetectorTransportAction and AnomalyResultTransportAction
9. Unwrap EsRejectedExecutionException as it can be nested inside RemoteTransportException. Previously, we would not recognize EsRejectedExecutionException and thus miss anomaly results write retrying.
10. Add error in anomaly result schema.
11. Fix broken tests due to my changes.

Testing done:
1. unit/integration tests pass
2. do end-to-end testing and make sure my fix achieves the purpose 
   * timeout issue is gone 
   * when all features have been disabled or index does not exist, we will retry a few more times and disable AD jobs.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
@@ -619,6 +619,116 @@ public class ModelManager {
     }
 
     /**
+    * Trains and saves cold-start AD models.
+    *
+    * This implementations splits RCF models and trains them all.
+    * As all model partitions have the same size, the scores from RCF models are merged by averaging.
+    * Since RCF outputs 0 until it is ready, initial 0 scores are meaningless and therefore filtered out.
+    * Filtered (non-zero) RCF scores are the training data for a single thresholding model.
+    * All trained models are serialized and persisted to be hosted.
+    *
+    * @param anomalyDetector the detector for which models are trained
+    * @param dataPoints M, N shape, where M is the number of samples for training and N is the number of features
+    * @param listener onResponse is called with null when this operation is completed
+    *                 onFailure is called IllegalArgumentException when training data is invalid
+    *                 onFailure is called LimitExceededException when a limit for training is exceeded
+    */
+    public void trainModel(AnomalyDetector anomalyDetector, double[][] dataPoints, ActionListener<Void> listener) {
+        if (dataPoints.length == 0 || dataPoints[0].length == 0) {
+            listener.onFailure(new IllegalArgumentException("Data points must not be empty."));
+        } else {
+            int rcfNumFeatures = dataPoints[0].length;
+            // creates partitioned RCF models
+            try {
+                Entry<Integer, Integer> partitionResults = getPartitionedForestSizes(
+                    RandomCutForest
+                        .builder()
+                        .dimensions(rcfNumFeatures)
+                        .sampleSize(rcfNumSamplesInTree)
+                        .numberOfTrees(rcfNumTrees)
+                        .outputAfter(rcfNumSamplesInTree)
+                        .parallelExecutionEnabled(false)
+                        .build(),
+                    anomalyDetector.getDetectorId()
+                );
+                int numForests = partitionResults.getKey();
+                int forestSize = partitionResults.getValue();
+                double[] scores = new double[dataPoints.length];
+                Arrays.fill(scores, 0.);
+                trainModelForStep(anomalyDetector, dataPoints, rcfNumFeatures, numForests, forestSize, scores, 0, listener);
+            } catch (LimitExceededException e) {
+                listener.onFailure(e);
+            }
+        }
+    }
+
+    private void trainModelForStep(
+        AnomalyDetector detector,
+        double[][] dataPoints,
+        int rcfNumFeatures,
+        int numForests,
+        int forestSize,
+        final double[] scores,
+        int step,
+        ActionListener<Void> listener
+    ) {
+        if (step < numForests) {
+            RandomCutForest rcf = RandomCutForest
+                .builder()
+                .dimensions(rcfNumFeatures)
+                .sampleSize(rcfNumSamplesInTree)
+                .numberOfTrees(forestSize)
+                .lambda(rcfTimeDecay)
+                .outputAfter(rcfNumSamplesInTree)
+                .parallelExecutionEnabled(false)
+                .build();
+            for (int j = 0; j < dataPoints.length; j++) {
+                scores[j] += rcf.getAnomalyScore(dataPoints[j]);
+                rcf.update(dataPoints[j]);
+            }
+            String modelId = getRcfModelId(detector.getDetectorId(), step);
+            String checkpoint = AccessController.doPrivileged((PrivilegedAction<String>) () -> rcfSerde.toJson(rcf));
+            checkpointDao
+                .putModelCheckpoint(
+                    modelId,
+                    checkpoint,
+                    ActionListener
+                        .wrap(
+                            r -> trainModelForStep(
+                                detector,
+                                dataPoints,
+                                rcfNumFeatures,
+                                numForests,
+                                forestSize,
+                                scores,
+                                step + 1,
+                                listener
+                            ),
+                            listener::onFailure
+                        )
+                );
+        } else {
+            double[] rcfScores = DoubleStream.of(scores).filter(score -> score > 0).map(score -> score / numForests).toArray();
+
+            // Train thresholding model
+            ThresholdingModel threshold = new HybridThresholdingModel(
+                thresholdMinPvalue,
+                thresholdMaxRankError,
+                thresholdMaxScore,
+                thresholdNumLogNormalQuantiles,
+                thresholdDownsamples,
+                thresholdMaxSamples
+            );
+            threshold.train(rcfScores);
+
+            // Persist thresholding model
+            String modelId = getThresholdModelId(detector.getDetectorId());
+            String checkpoint = AccessController.doPrivileged((PrivilegedAction<String>) () -> gson.toJson(threshold));
+            checkpointDao.putModelCheckpoint(modelId, checkpoint, ActionListener.wrap(r -> listener.onResponse(null), listener::onFailure));
+        }
+    }
+
+    /**
      * Returns the model ID for the RCF model partition.
      *
      * @param detectorId ID of the detector for which the RCF model is trained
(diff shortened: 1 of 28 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -225,9 +213,20 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
...
+
+    private ActionListener<Optional<AnomalyDetector>> onGetDetector(
+        ActionListener<AnomalyResultResponse> listener,
+        String adID,
+        AnomalyResultRequest request
+    ) {
+        return ActionListener.wrap(detector -> {
             if (!detector.isPresent()) {
-                listener.onFailure(new EndRunException(adID, "AnomalyDetector is not available.", true));
+                listener.onFailure(new EndRunException(adID, "AnomalyDetector is not available.", false));
```
ylwu-amzn, line 229: Why not end run immediately if we can't find detector?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -342,9 +358,19 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
...
-            handleExecuteException(ex, listener, adID);
-        }
+        }, exception -> {
+            LOG.warn(exception);
+            if (exception instanceof IndexNotFoundException) {
+                listener.onFailure(new EndRunException(adID, "Having trouble querying data: " + exception.getMessage(), false));
+            } else if (exception instanceof IllegalArgumentException) {
+                listener
+                    .onFailure(
+                        new EndRunException(adID, "Having trouble querying data. Maybe all of your features have been disabled.", false)
```
ylwu-amzn, line 368: How about we check feature list and give user a definite answer? If that change takes time, please add some todo here.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -342,9 +358,19 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
                         new ActionListenerResponseHandler<>(rcfListener, RCFResultResponse::new)
                     );
             }
-        } catch (Exception ex) {
-            handleExecuteException(ex, listener, adID);
-        }
+        }, exception -> {
+            LOG.warn(exception);
+            if (exception instanceof IndexNotFoundException) {
+                listener.onFailure(new EndRunException(adID, "Having trouble querying data: " + exception.getMessage(), false));
```
ylwu-amzn, line 364: How about we change to "Can't find index XXX"? So user can know clearly the trouble is missing index, rather than others like wrong query or network latency.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -272,6 +272,35 @@ public String getDetectorIdForModelId(String modelId) {
...
+     * If there are more partitions than the number of nodes in the cluster,
+     * the model is partitioned by the number of nodes and verified to
+     * ensure the size of a partition does not exceed the max size limit based on heap.
+     *
+     * @param detectorId ID of the detector with no effects on partitioning
+     * @param rcfNumFeatures the number of features
+     * @return a pair of number of partitions and size of a parition (number of trees)
+     * @throws LimitExceededException when there is no sufficient resouce available
+     */
+    public Entry<Integer, Integer> getPartitionedForestSizes(String detectorId, int rcfNumFeatures) {
```
wnbts, line 290: if we are going to refactor this method, I suggest the new api just takes a detector object, which contains all the needed info and simpler to use.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -91,67 +87,63 @@ public ADStateManager(
     /**
      * Get the number of RCF model's partition number for detector adID
      * @param adID detector id
+     * @param detector object
      * @return the number of RCF model's partition number for adID
      * @throws InterruptedException when we cannot get anomaly detector object for adID before timeout
      * @throws LimitExceededException when there is no sufficient resource available
      */
-    public int getPartitionNumber(String adID) throws InterruptedException {
+    public int getPartitionNumber(String adID, Optional<AnomalyDetector> detector) throws InterruptedException {
```
wnbts, line 95: Minor. Why not validate detector first and just pass a detector afterwards? saving all the repetitive and unlikely handling of a non-existent detector.

src/main/resources/mappings/anomaly-results.json
```diff
@@ -45,6 +45,9 @@
     "execution_end_time": {
       "type": "date",
       "format": "strict_date_time||epoch_millis"
+    },
+    "error": {
+      "type": "text"
```
wnbts, line 50: question. does error message need to be searched for?

ylwu-amzn: Yes, the error message is not standardized, that means it could be any string, better to make it searchable for easy operation.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: It is possible the system undergoes performance issue and get detector requests gets dropped somehow.  We should retry.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: good suggestion.  Done.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: exception.getMessage would return such information.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: We are only use use detector id as part of error message. Don't need other detector information.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: good point.  Done.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
wnbts: if model manager takes a detector, it can compute the feature dimensions and partitioning so that will be only input needed and that will save client the work to provide a second rcfNumFeatures input. that's why i suggest doing that.

kaituo: done

kaituo: please take a look at the recent commit: https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/83/commits/d8ea9cf5ec5c46f09b1b6fbcd8f5e725e0fa50b6

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -91,67 +86,59 @@ public ADStateManager(
     /**
      * Get the number of RCF model's partition number for detector adID
      * @param adID detector id
+     * @param detector object
      * @return the number of RCF model's partition number for adID
      * @throws InterruptedException when we cannot get anomaly detector object for adID before timeout
      * @throws LimitExceededException when there is no sufficient resource available
      */
-    public int getPartitionNumber(String adID) throws InterruptedException {
+    public int getPartitionNumber(String adID, AnomalyDetector detector) throws InterruptedException {
```
wnbts, line 94: Minor. Is this exception still possible?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -598,6 +634,116 @@ public void trainModel(AnomalyDetector anomalyDetector, double[][] dataPoints) {
...
+    * Filtered (non-zero) RCF scores are the training data for a single thresholding model.
+    * All trained models are serialized and persisted to be hosted.
+    *
+    * @param anomalyDetector the detector for which models are trained
+    * @param dataPoints M, N shape, where M is the number of samples for training and N is the number of features
+    * @param listener onResponse is called with null when this operation is completed
+    *                 onFailure is called IllegalArgumentException when training data is invalid
+    *                 onFailure is called LimitExceededException when a limit for training is exceeded
+    */
+    public void trainModel(AnomalyDetector anomalyDetector, double[][] dataPoints, ActionListener<Void> listener) {
```
wnbts, line 652: suggestion. it's clearer to rebase and force push to avoid showing pulled changes that are checked in.

kaituo: yes, will try next time

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: no, it is impossible now.  Fixed it.

PR #84, opened 2020-04-10
Add state and error to profile API

kaituo:
*Issue #, if available:*

*Description of changes:*

We want to make it easy for customers and oncalls to identify a detector’s state and error if any. This PR adds such information to our new profile API.

We expect three kinds of states:
-Disabled: if get ad job api says the job is disabled;
-Init: if anomaly score after the last update time of the detector is larger than 0
-Running: if neither of the above applies and no exceptions.

Error is populated if error of the latest anomaly result is not empty.

Example:

```
curl -X GET "localhost:9200/_opendistro/_anomaly_detection/detectors/t6gGRXEBNjeafFFiEvhk/_profile"
{"state":"INIT","error":"No full shingle in current detection window"}

curl -X GET "localhost:9200/_opendistro/_anomaly_detection/detectors/t6gGRXEBNjeafFFiEvhk/_profile/state"
{"state":"INIT"}
```

Testing done:
-manual testing during a detector’s life cycle: not created, created but not started, started, during initialization, after initialization, stopped, restarted
-added unit tests to cover above scenario

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad;
+
+import static com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector.ANOMALY_DETECTORS_INDEX;
+import static com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX;
+import static org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
+
+import java.io.IOException;
+import java.util.Set;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.get.GetRequest;
+import org.elasticsearch.action.get.GetResponse;
+import org.elasticsearch.action.search.SearchRequest;
+import org.elasticsearch.action.search.SearchResponse;
+import org.elasticsearch.client.Client;
+import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
+import org.elasticsearch.common.xcontent.NamedXContentRegistry;
+import org.elasticsearch.common.xcontent.XContentParseException;
+import org.elasticsearch.common.xcontent.XContentParser;
+import org.elasticsearch.common.xcontent.XContentType;
+import org.elasticsearch.index.IndexNotFoundException;
+import org.elasticsearch.index.query.BoolQueryBuilder;
+import org.elasticsearch.index.query.QueryBuilders;
+import org.elasticsearch.search.SearchHit;
+import org.elasticsearch.search.SearchHits;
+import org.elasticsearch.search.builder.SearchSourceBuilder;
+import org.elasticsearch.search.sort.FieldSortBuilder;
+import org.elasticsearch.search.sort.SortOrder;
+
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetectorJob;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyResult;
+import com.amazon.opendistroforelasticsearch.ad.model.DetectorProfile;
+import com.amazon.opendistroforelasticsearch.ad.model.DetectorState;
+import com.amazon.opendistroforelasticsearch.ad.model.ProfileName;
+import com.amazon.opendistroforelasticsearch.ad.util.DelegateActionListener;
+
+public class AnomalyDetectorProfileRunner {
+    private final Logger logger = LogManager.getLogger(AnomalyDetectorProfileRunner.class);
+    private Client client;
+    private NamedXContentRegistry xContentRegistry;
+    static String FAIL_TO_FIND_DETECTOR_MSG = "Fail to find detector with id: ";
+
+    public AnomalyDetectorProfileRunner(Client client, NamedXContentRegistry xContentRegistry) {
+        this.client = client;
+        this.xContentRegistry = xContentRegistry;
+    }
+
+    public void profile(String detectorId, ActionListener<DetectorProfile> listener, Set<String> profiles) {
+        DelegateActionListener<DetectorProfile> delegateListener = new DelegateActionListener<DetectorProfile>(
+            listener,
+            profiles.size(),
+            "Fail to fetch profile for " + detectorId
+        );
+
+        if (profiles.isEmpty()) {
+            listener.onFailure(new RuntimeException("Unsupported profile types."));
+            return;
+        }
+
+        if (profiles.contains(ProfileName.STATE.getName()) || profiles.contains(ProfileName.ERROR.getName())) {
+            prepareProfileStateNError(detectorId, delegateListener, profiles);
+        }
+    }
+
+    private void prepareProfileStateNError(String detectorId, DelegateActionListener<DetectorProfile> listener, Set<String> profiles) {
+        GetRequest getDetectorRequest = new GetRequest(ANOMALY_DETECTORS_INDEX, detectorId);
+        client.get(getDetectorRequest, onGetDetectorResponse(listener, detectorId, profiles));
+    }
+
+    private ActionListener<GetResponse> onGetDetectorResponse(
+        DelegateActionListener<DetectorProfile> listener,
+        String detectorId,
+        Set<String> profiles
+    ) {
+        return ActionListener.wrap(getResponse -> {
+            if (getResponse != null && getResponse.isExists()) {
+                try (
+                    XContentParser parser = XContentType.JSON
+                        .xContent()
+                        .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, getResponse.getSourceAsString())
+                ) {
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser::getTokenLocation);
+                    AnomalyDetector detector = parser.namedObject(AnomalyDetector.class, AnomalyDetector.PARSE_FIELD_NAME, null);
+                    long lastUpdateTimeMs = detector.getLastUpdateTime().toEpochMilli();
+
+                    if (profiles.contains(ProfileName.STATE.getName())) {
+                        profileState(detectorId, lastUpdateTimeMs, listener);
+                    }
+                    if (profiles.contains(ProfileName.ERROR.getName())) {
+                        profileError(detectorId, lastUpdateTimeMs, listener);
+                    }
+
+                } catch (IOException | XContentParseException | NullPointerException e) {
+                    logger.error(e);
+                    listener.failImmediately(new RuntimeException(FAIL_TO_FIND_DETECTOR_MSG + detectorId, e));
+                }
+            } else {
+                listener.failImmediately(FAIL_TO_FIND_DETECTOR_MSG + detectorId);
+            }
+        }, exception -> { listener.failImmediately(FAIL_TO_FIND_DETECTOR_MSG + detectorId, exception); });
+    }
+
+    /**
+     * We expect three kinds of states:
+     *  -Disabled: if get ad job api says the job is disabled;
+     *  -Init: if anomaly score after the last update time of the detector is larger than 0
+     *  -Running: if neither of the above applies and no exceptions.
+     * @param detectorId detector id
+     * @param lastUpdateTimeMs last update time of the detector in milliseconds
+     * @param listener listener to process the returned state or exception
+     */
+    private void profileState(String detectorId, long lastUpdateTimeMs, DelegateActionListener<DetectorProfile> listener) {
+        GetRequest getRequest = new GetRequest(ANOMALY_DETECTOR_JOB_INDEX, detectorId);
+        client.get(getRequest, ActionListener.wrap(getResponse -> {
+            if (getResponse.isExists()) {
+                try (
+                    XContentParser parser = XContentType.JSON
+                        .xContent()
+                        .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, getResponse.getSourceAsString())
+                ) {
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser::getTokenLocation);
+                    AnomalyDetectorJob job = AnomalyDetectorJob.parse(parser);
+                    if (job.isEnabled()) {
+                        SearchRequest searchLatestResult = createInittedEverRequest(detectorId, lastUpdateTimeMs);
+                        client.search(searchLatestResult, onInittedEver(listener, detectorId, lastUpdateTimeMs));
+                    } else {
+                        DetectorProfile profile = new DetectorProfile();
+                        profile.setState(DetectorState.DISABLED);
+                        listener.onResponse(profile);
+                    }
+                } catch (IOException | XContentParseException e) {
+                    String error = "Fail to parse detector with id: " + detectorId;
+                    logger.error(error);
+                    listener.onFailure(new RuntimeException(error, e));
+                }
+            } else {
+                DetectorProfile profile = new DetectorProfile();
+                profile.setState(DetectorState.DISABLED);
+                listener.onResponse(profile);
+            }
+        }, exception -> {
+            logger.warn(exception);
+            // detector job index does not exist
+            if (exception instanceof IndexNotFoundException) {
+                DetectorProfile profile = new DetectorProfile();
+                profile.setState(DetectorState.DISABLED);
+                listener.onResponse(profile);
+            } else {
+                logger.error("Fail to get detector state for " + detectorId);
+                listener.onFailure(exception);
+            }
+        }));
+
+    }
+
+    private ActionListener<SearchResponse> onInittedEver(
+        DelegateActionListener<DetectorProfile> listener,
+        String detectorId,
+        long lastUpdateTimeMs
+    ) {
+        return ActionListener.wrap(searchResponse -> {
+            SearchHits hits = searchResponse.getHits();
+            DetectorProfile profile = new DetectorProfile();
+            if (hits.getTotalHits().value == 0L) {
+                profile.setState(DetectorState.INIT);
+            } else {
+                profile.setState(DetectorState.RUNNING);
+            }
+
+            listener.onResponse(profile);
+
+        }, exception -> {
+            if (exception instanceof IndexNotFoundException) {
+                DetectorProfile profile = new DetectorProfile();
+                // anomaly result index is not created yet
+                profile.setState(DetectorState.INIT);
+                listener.onResponse(profile);
+            } else {
+                logger.error("Fail to find latest anomaly result of id: {}", detectorId);
+                listener.onFailure(new RuntimeException("Fail to find detector state: " + detectorId, exception));
+            }
+        });
+    }
+
+    /**
+     * Error is populated if error of the latest anomaly result is not empty.
+     * @param detectorId detector id
+     * @param lastUpdateTimeMs last update time of the detector in milliseconds
+     * @param listener listener to process the returned error or exception
+     */
+    private void profileError(String detectorId, long lastUpdateTimeMs, DelegateActionListener<DetectorProfile> listener) {
+        SearchRequest searchLatestResult = createLatestAnomalyResultRequest(detectorId, lastUpdateTimeMs);
+        client.search(searchLatestResult, onGetLatestAnomalyResult(listener, detectorId));
+    }
+
+    private ActionListener<SearchResponse> onGetLatestAnomalyResult(ActionListener<DetectorProfile> listener, String detectorId) {
+        return ActionListener.wrap(searchResponse -> {
+            SearchHits hits = searchResponse.getHits();
+            if (hits.getTotalHits().value == 0L) {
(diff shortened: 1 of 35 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+    /**
+     * Create search request to check if we have at least 1 anomaly score larger than 0 after last update time
+     * @param detectorId detector id
+     * @param lastUpdateTimeEpochMs last update time in milliseconds
+     * @return the search request
+     */
+    private SearchRequest createInittedEverRequest(String detectorId, long lastUpdateTimeEpochMs) {
+        BoolQueryBuilder filterQuery = new BoolQueryBuilder();
+        filterQuery.filter(QueryBuilders.termQuery(AnomalyResult.DETECTOR_ID_FIELD, detectorId));
+        filterQuery.filter(QueryBuilders.rangeQuery(AnomalyResult.EXECUTION_END_TIME_FIELD).gte(lastUpdateTimeEpochMs));
```
ylwu-amzn, line 260: Use job `enabled_time` here, think of the case : detector last update time not changed, but we disabled and restarted job multiple times. So we may get some AD result with non-zero anomaly score which generated before latest job enabled time, but actually the latest AD job is still initializing.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+
+        SearchRequest request = new SearchRequest(AnomalyResult.ANOMALY_RESULT_INDEX);
+        request.source(source);
+        return request;
+    }
+
+    private SearchRequest createLatestAnomalyResultRequest(String detectorId, long lastUpdateTimeEpochMs) {
+        BoolQueryBuilder filterQuery = new BoolQueryBuilder();
+        filterQuery.filter(QueryBuilders.termQuery(AnomalyResult.DETECTOR_ID_FIELD, detectorId));
+        filterQuery.filter(QueryBuilders.rangeQuery(AnomalyResult.EXECUTION_END_TIME_FIELD).gte(lastUpdateTimeEpochMs));
```
ylwu-amzn, line 273: Similar as line 260, we should use AD job "enabled_time"

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+                        SearchRequest searchLatestResult = createInittedEverRequest(detectorId, lastUpdateTimeMs);
+                        client.search(searchLatestResult, onInittedEver(listener, detectorId, lastUpdateTimeMs));
+                    } else {
+                        DetectorProfile profile = new DetectorProfile();
+                        profile.setState(DetectorState.DISABLED);
+                        listener.onResponse(profile);
+                    }
+                } catch (IOException | XContentParseException e) {
+                    String error = "Fail to parse detector with id: " + detectorId;
+                    logger.error(error);
```
ylwu-amzn, line 151: log exception stack trace to make operation easier? Similar to other places

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
```diff
@@ -0,0 +1,104 @@
...
+import org.elasticsearch.action.ActionListener;
+
+import com.amazon.opendistroforelasticsearch.ad.model.Mergeable;
+
+public class DelegateActionListener<T extends Mergeable> implements ActionListener<T> {
+    private static final Logger LOG = LogManager.getLogger(DelegateActionListener.class);
+    private final ActionListener<T> delegate;
+    private final AtomicInteger collectedResponseCount;
+    private final int expectedResponseCount;
+    private final List<T> savedResponses;
```
ylwu-amzn, line 35: `saved` means the responses are from some saved result from ES indices ? Or means we cache these response ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestGetAnomalyDetectorAction.java
```diff
@@ -55,11 +65,32 @@
...
+        controller
+            .registerHandler(
+                RestRequest.Method.GET,
+                String.format(Locale.ROOT, "%s/{%s}/%s", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, DETECTOR_ID, PROFILE),
+                this
+            );
+        controller
+            .registerHandler(
+                RestRequest.Method.GET,
+                String.format(Locale.ROOT, "%s/{%s}/%s/{%s}", AnomalyDetectorPlugin.AD_BASE_DETECTORS_URI, DETECTOR_ID, PROFILE, TYPE),
```
ylwu-amzn, line 91: How about we add some comments about what TYPE means and the supported value?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestGetAnomalyDetectorAction.java
```diff
@@ -148,4 +178,25 @@ public RestResponse buildResponse(MultiGetResponse multiGetResponse) throws Exce
...
+                exception -> { channel.sendResponse(buildInternalServerErrorResponse(exception, exception.getMessage())); }
+            );
+    }
+
+    private RestResponse buildInternalServerErrorResponse(Exception e, String errorMsg) {
+        logger.error(errorMsg, e);
+        return new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, errorMsg);
+    }
+
+    private Set<String> getProfilesToCollect(String typesStr) {
```
ylwu-amzn, line 194: How about we validate type here and return `Set<ProfileName>`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Locale;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.elasticsearch.action.ActionListener;
+
+import com.amazon.opendistroforelasticsearch.ad.model.Mergeable;
+
+public class DelegateActionListener<T extends Mergeable> implements ActionListener<T> {
```
ylwu-amzn, line 30: Seems you design this general delegate listener not only for profile API. Can you add more comments? Suggest to use a more specific name like `MultiResponsesDelegateActionListener` ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
```diff
@@ -0,0 +1,104 @@
...
+
+    }
+
+    @Override
+    public void onFailure(Exception e) {
+        LOG.info(e);
+        try {
+            this.exceptions.add(e.getMessage());
+        } finally {
+            if (collectedResponseCount.incrementAndGet() == expectedResponseCount) {
```
ylwu-amzn, line 69: If `expectedResponseCount==0` , `collectedResponseCount.incrementAndGet()` will always greater than `expectedResponseCount `, please add some validation for `expectedResponseCount`, or change to `collectedResponseCount.incrementAndGet() >= expectedResponseCount`

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    private void finish() {
+        if (this.exceptions.size() == 0) {
+            if (savedResponses.size() == 0) {
+                this.delegate.onFailure(new RuntimeException(String.format("Unexpected exceptions")));
+            } else {
+                T response0 = savedResponses.get(0);
+                LOG.info(response0);
```
ylwu-amzn, line 81: Why we log `response0` here?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+    public AnomalyDetectorProfileRunner(Client client, NamedXContentRegistry xContentRegistry) {
+        this.client = client;
+        this.xContentRegistry = xContentRegistry;
+    }
+
+    public void profile(String detectorId, ActionListener<DetectorProfile> listener, Set<String> profiles) {
+        DelegateActionListener<DetectorProfile> delegateListener = new DelegateActionListener<DetectorProfile>(
+            listener,
+            profiles.size(),
+            "Fail to fetch profile for " + detectorId
```
ylwu-amzn, line 70: Here, the `finalErrorMsg` is `"Fail to fetch profile for " + detectorId`. 
From line89 of class `DelegateActionListener`: `this.delegate.onFailure(new RuntimeException(String.format(Locale.ROOT, finalErrorMsg, exceptions)));`,  `String.format(...)` will not include exceptions, is this by design?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+        );
+
+        if (profiles.isEmpty()) {
+            listener.onFailure(new RuntimeException("Unsupported profile types."));
```
ylwu-amzn, line 74: How about we change to "Must set at least one profile type" to avoid confusion between empty profile types and wrong profile types which we don't support?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+                    if (profiles.contains(ProfileName.STATE.getName())) {
+                        profileState(detectorId, lastUpdateTimeMs, listener);
+                    }
+                    if (profiles.contains(ProfileName.ERROR.getName())) {
+                        profileError(detectorId, lastUpdateTimeMs, listener);
+                    }
+
+                } catch (IOException | XContentParseException | NullPointerException e) {
+                    logger.error(e);
+                    listener.failImmediately(new RuntimeException(FAIL_TO_FIND_DETECTOR_MSG + detectorId, e));
```
ylwu-amzn, line 113: minor: can use this method: `listener.failImmediately(String errMsg, Exception e)`

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+                    AnomalyDetectorJob job = AnomalyDetectorJob.parse(parser);
+                    if (job.isEnabled()) {
+                        SearchRequest searchLatestResult = createInittedEverRequest(detectorId, lastUpdateTimeMs);
+                        client.search(searchLatestResult, onInittedEver(listener, detectorId, lastUpdateTimeMs));
+                    } else {
+                        DetectorProfile profile = new DetectorProfile();
+                        profile.setState(DetectorState.DISABLED);
+                        listener.onResponse(profile);
+                    }
+                } catch (IOException | XContentParseException e) {
```
ylwu-amzn, line 149: If some uncatched exception, will not execute `listener.onFailure` method, `collectedResponseCount` will not increase, so will never execute `finish`. 
Suggest to catch `Exception` here to avoid some uncatched exception caused `DelegateActionListener` never run `finish` method.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+                    logger.error(error);
+                    listener.onFailure(new RuntimeException(error, e));
+                }
+            } else {
+                DetectorProfile profile = new DetectorProfile();
+                profile.setState(DetectorState.DISABLED);
+                listener.onResponse(profile);
+            }
+        }, exception -> {
+            logger.warn(exception);
```
ylwu-amzn, line 160: Can we add custom error message here?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+            listener.onResponse(profile);
+
+        }, exception -> {
+            if (exception instanceof IndexNotFoundException) {
+                DetectorProfile profile = new DetectorProfile();
+                // anomaly result index is not created yet
+                profile.setState(DetectorState.INIT);
+                listener.onResponse(profile);
+            } else {
+                logger.error("Fail to find latest anomaly result of id: {}", detectorId);
```
ylwu-amzn, line 197: minor: make the error message more accurate, like `Fail to find latest anomaly result with anomalyScore>0 from XXX for detector XXX`

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,283 @@
...
+        SearchRequest searchLatestResult = createLatestAnomalyResultRequest(detectorId, lastUpdateTimeMs);
+        client.search(searchLatestResult, onGetLatestAnomalyResult(listener, detectorId));
+    }
+
+    private ActionListener<SearchResponse> onGetLatestAnomalyResult(ActionListener<DetectorProfile> listener, String detectorId) {
+        return ActionListener.wrap(searchResponse -> {
+            SearchHits hits = searchResponse.getHits();
+            if (hits.getTotalHits().value == 0L) {
+                logger.error("We should not get empty result: {}", detectorId);
+                listener.onFailure(new RuntimeException("Unexpected error while looking for detector state:  " + detectorId));
```
ylwu-amzn, line 219: Why throw exception if we can't find AD result ? If not AD result, that means AD job is initializing and no error.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: good point. Done.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestGetAnomalyDetectorAction.java
kaituo: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestGetAnomalyDetectorAction.java
kaituo: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
kaituo: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
kaituo: the latter. Added a comment.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
kaituo: good point.  Used the latter.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/DelegateActionListener.java
kaituo: It is used for debugging.  removed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: fixed as we discussed offline.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    public void profile(String detectorId, ActionListener<DetectorProfile> listener, Set<ProfileName> profiles) {
+        MultiResponsesDelegateActionListener<DetectorProfile> delegateListener = new MultiResponsesDelegateActionListener<DetectorProfile>(
+            listener,
+            profiles.size(),
+            "Fail to fetch profile for " + detectorId
+        );
+
+        if (profiles.isEmpty()) {
```
yizheliu-amazon, line 73: not blocker: you can move this isEmpty() check to the entry of this method, aka line 67. And then you can skip the check on line 78

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ProfileName.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static ProfileName getName(String name) {
+        switch (name) {
+            case "state":
+                return STATE;
+            case "error":
+                return ERROR;
+            default:
+                throw new IllegalArgumentException("Unsupported prof");
```
yizheliu-amazon, line 62: `Unsupported profile`

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
```diff
@@ -0,0 +1,111 @@
...
+            if (collectedResponseCount.incrementAndGet() >= expectedResponseCount) {
+                finish();
+            }
+        }
+
+    }
+
+    @Override
+    public void onFailure(Exception e) {
+        LOG.info(e);
```
yizheliu-amazon, line 70: LOG.error()

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
```diff
@@ -0,0 +1,111 @@
...
+        try {
+            this.exceptions.add(e.getMessage());
+        } finally {
+            if (collectedResponseCount.incrementAndGet() >= expectedResponseCount) {
+                finish();
+            }
+        }
```
yizheliu-amazon, line 77: It looks like to be duplicate with above line 60. Can we remove the `finally` here? And also, the only potential scenario where exception can get thrown is that `Exception e` is null, but I don't think it is possible.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public void onResponse(T response) {
+        try {
+            if (response != null) {
+                this.savedResponses.add(response);
+            }
+        } catch (Exception e) {
+            onFailure(e);
+        } finally {
+            if (collectedResponseCount.incrementAndGet() >= expectedResponseCount) {
```
yizheliu-amazon, line 61: `expected` implies that the total collected count must be more than `expectedResponseCount`, otherwise it is a failure. Based on my understanding of use of this class, I guess `maxResponseCount` might be a better name.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
yizheliu-amazon: After another look, I guess exception can be thrown if thread is interrupted.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
```diff
@@ -0,0 +1,111 @@
...
+                finish();
+            }
+        }
+    }
+
+    private void finish() {
+        if (this.exceptions.size() == 0) {
+            if (savedResponses.size() == 0) {
+                this.delegate.onFailure(new RuntimeException(String.format("Unexpected exceptions")));
```
yizheliu-amazon, line 83: String.format may not be needed if only static string is there. Also, I think in case of empty exceptions and empty savedResponses, it may be better to throw exception with message like `No response collected`, which makes more sense to me.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -0,0 +1,287 @@
...
+                } catch (IOException | XContentParseException | NullPointerException e) {
+                    logger.error(e);
+                    listener.failImmediately(FAIL_TO_GET_PROFILE_MSG, e);
+                }
+            } else {
+                GetRequest getDetectorRequest = new GetRequest(ANOMALY_DETECTORS_INDEX, detectorId);
+                client.get(getDetectorRequest, onGetDetectorResponse(listener, detectorId, profiles));
+            }
+        }, exception -> {
+            if (exception instanceof IndexNotFoundException) {
```
yizheliu-amazon, line 115: can you log exception here as well?

kaituo: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
kaituo: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
kaituo: yes, any exception can be thrown by an asynchronous request.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/MultiResponsesDelegateActionListener.java
kaituo: good catch.  Removed String.format and changed to "No response collected".

Issue #86, opened 2020-04-14
AD plugin opendistro1.6 support


vamshin:
Make AD plugin compatible for ODFE 1.6 version which uses Elasticsearch OSS 7.6.1 version

vamshin:
https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/87

PR #87, opened 2020-04-14
AD opendistro 1.6 support

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/86

*Description of changes:*
Make AD plugin compatible for ODFE 1.6 version which uses Elasticsearch OSS 7.6.1 version


build.gradle
```diff
@@ -42,13 +43,8 @@ repositories {
 }
 
 ext {
-    opendistroVersion = '1.4.0'
+    opendistroVersion = '1.6.0'
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
-    if (System.properties['os.name'].toLowerCase().contains('windows')) {
```
kaituo, line 47: After removing these lines, can we build on both windows and Linux?

build.gradle
```diff
@@ -120,80 +119,6 @@ thirdPartyAudit.enabled = false
 // See package README.md for details on using these tasks.
 def _numNodes = findProperty('numNodes') as Integer ?: 1
 
-def getSeedHosts = { int num ->
```
kaituo, line 123: Why do we remove the following, which is used to start and stop a multi-node cluster?

vamshin: I think this was added to to provide hack to install job scheduler plugin on the test clusters.   With ES 7.5 onwards Elasticsearch provided a way to install dependent plugins on test cluster. 
Now we do not need it. To run multi node cluster, we need to use
```./gradlew run -PnumNodes=<numberOfNodesYouWant>```

build.gradle
vamshin: Previously we were installing using ```setupCommand``` where we used absolute path for file.  Now we just need to pass ```filetree``` 
Example:-
plugin(fileTree("src/test/resources/job-scheduler").getSingleFile())

kaituo: great

build.gradle
kaituo: Good to know.  Could you keep runSingleNode and stopMultiNode (line 165~196) so that I can start node one by one and kill all nodes?  Need them for fault tolerance tests.

vamshin: Created issue to add this back https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/90

PR #93, opened 2020-04-28
add async stopModel

wnbts:
Current stopModel is a synchronous operation for stopping hosting models. As part of migration to asynchronous implementation, this change adds an asynchronous equivalent using the same business logic with one minor change/fix - setting lastCheckpointTime on a stopped and saved model object going into garbage collection is unnecessary and is therefore removed for cleanness.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
@@ -537,10 +537,50 @@ public class ModelManager {
         Optional
             .ofNullable(models.remove(modelId))
             .filter(model -> model.getLastCheckpointTime().plus(checkpointInterval).isBefore(now))
-            .ifPresent(model -> {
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -537,10 +537,50 @@ public void stopModel(String detectorId, String modelId) {
         Optional
             .ofNullable(models.remove(modelId))
             .filter(model -> model.getLastCheckpointTime().plus(checkpointInterval).isBefore(now))
-            .ifPresent(model -> {
-                checkpointDao.putModelCheckpoint(modelId, toCheckpoint.apply(model.getModel()));
-                model.setLastCheckpointTime(now);
```
kaituo, line 542: Why is setting last checkpoint time not necessary any more?

wnbts: because the model has been removed from the managed models at line 538 and has become garbage at the end of the method.

PR #105, opened 2020-05-04
Add settings to disable/enable AD dynamically

kaituo:
*Issue #, if available:*

*Description of changes:*

This PR adds settings to disable/enable AD dynamically.  AD plugin rejects RESTful requests and stops AD jobs when it is disabled.

Testing done:
- added tests to verify if AD plugin rejects RESTful requests when it is disabled
- manually verified AD jobs are stopped when AD plugin is disabled

Usage:

```
curl -XPUT "localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d'
{
  "persistent": {
    "opendistro.anomaly_detection.enabled": false
  }
}'
```

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
@@ -1,0 +1,111 @@
+/*
+ *   Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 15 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
```diff
@@ -0,0 +1,111 @@
...
+     * Settings name
+     */
+    public static final String AD_PLUGIN_ENABLED = "opendistro.anomaly_detection.enabled";
+
+    private final Map<String, Setting<?>> settings = unmodifiableMap(new HashMap<String, Setting<?>>() {
+        {
+            /**
+             * AD plugin enable/disable setting
+             */
+            put(AD_PLUGIN_ENABLED, Setting.boolSetting(AD_PLUGIN_ENABLED, true, NodeScope, Dynamic));
```
ylwu-amzn, line 53: We will set AD plugin as enabled by default. After user disabling it and restart domain, are we able to read the latest configuration and set AD as disabled by defulat?

kaituo: It depends on whether they do it in a transient fashion or persistent fashion.  If transient, after reboot,  it would be enabled by default.  If persistent, after reboot, it would disabled by default.

Issue #107, opened 2020-05-05
Stats API Call timeout exception


jmazanec15:
`GET /_opendistro/_anomaly_detection/stats` occasionally results in the following error message:

```
{
  "error": {
    "root_cause": [
      {
        "type": "timeout_exception",
        "reason": "Cannot get response within time limit: SearchRequest{searchType=QUERY_THEN_FETCH, indices=[.opendistro-anomaly-detectors], indicesOptions=IndicesOptions[ignore_unavailable=false, allow_no_indices=true, expand_wildcards_open=true, expand_wildcards_closed=false, allow_aliases_to_multiple_indices=true, forbid_closed_indices=true, ignore_aliases=false, ignore_throttled=true], types=[], routing='null', preference='null', requestCache=null, scroll=null, maxConcurrentShardRequests=0, batchedReduceSize=512, preFilterShardSize=128, allowPartialSearchResults=true, localClusterAlias=null, getOrCreateAbsoluteStartMillis=-1, ccsMinimizeRoundtrips=true, source={\"size\":0}}"
      }
    ],
    "type": "timeout_exception",
    "reason": "Cannot get response within time limit: SearchRequest{searchType=QUERY_THEN_FETCH, indices=[.opendistro-anomaly-detectors], indicesOptions=IndicesOptions[ignore_unavailable=false, allow_no_indices=true, expand_wildcards_open=true, expand_wildcards_closed=false, allow_aliases_to_multiple_indices=true, forbid_closed_indices=true, ignore_aliases=false, ignore_throttled=true], types=[], routing='null', preference='null', requestCache=null, scroll=null, maxConcurrentShardRequests=0, batchedReduceSize=512, preFilterShardSize=128, allowPartialSearchResults=true, localClusterAlias=null, getOrCreateAbsoluteStartMillis=-1, ccsMinimizeRoundtrips=true, source={\"size\":0}}"
  },
  "status": 500
}
```

This is a result of an expensive timed blocking call [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/util/IndexUtils.java#L110) to count up the documents in the index via Search. (1) SearchRequest should be replaced with an IndicesStatsRequest and (2) the call should be converted to an asynchronous call outside of the transport layer.

PR #108, opened 2020-05-05
Stats API: moved detector count call outside transport layer and make asynchronous

jmazanec15:
*Issue #, if available:*
#107 

*Description of changes:*
Previously, getting the detector count for the stats API was leading to timeouts. The detector count was retrieved by making a timed blocking `SearchRequest` in the transport layer and tallying up the number of results.

This PR changes the way in which the detector count is retrieved. It moves the logic to get the detector count to a nonblocking `IndicesStatsRequest` in the Rest layer. Now, the rest layer makes two asynchronous requests:
1. IndicesStatsRequest to get detector count (only if needed)
2. ADStatsRequest to get the node level stats

On response from those requests, it combines the node stats and the cluster stats together and sends the response back to the user.

In order to test, I modified existing tests and added a new one. I also manually tested on a local cluster using the command `./gradlew run -PnumNodes=3`


src/main/java/com/amazon/opendistroforelasticsearch/ad/stats/ADStatsResponse.java
```diff
@@ -0,0 +1,139 @@
...
+        for (Map.Entry<String, Object> clusterStat : clusterStats.entrySet()) {
+            builder.field(clusterStat.getKey(), clusterStat.getValue());
+        }
+        adStatsNodesResponse.toXContent(xContentBuilder, ToXContent.EMPTY_PARAMS);
+        return xContentBuilder.endObject();
+    }
+
+    @Override
+    public void merge(Mergeable other) {
+        if (this == other || other == null || getClass() != other.getClass()) {
```
wnbts, line 96: Minor. other instanceof ADStatsResponse seems better.

src/main/java/com/amazon/opendistroforelasticsearch/ad/stats/ADStatsResponse.java
```diff
@@ -0,0 +1,139 @@
...
+            return;
+        }
+
+        ADStatsResponse otherResponse = (ADStatsResponse) other;
+
+        if (otherResponse.adStatsNodesResponse != null) {
```
wnbts, line 102: Minor. Some edge cases might be handled more clearly, such as when both have node responses/cluster stats, throw an exception if unexpected.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -115,4 +131,111 @@ private ADStatsRequest getRequest(RestRequest request) {
...
+     *
+     * @param adStatsRequest Request containing stats to be retrieved
+     * @return Map containing Cluster Stats
+     */
+    private Map<String, Object> getClusterStats(ADStatsRequest adStatsRequest) {
+        Map<String, Object> clusterStats = new HashMap<>();
+        Set<String> statsToBeRetrieved = adStatsRequest.getStatsToBeRetrieved();
+        for (String statName : adStats.getClusterStats().keySet()) {
+            if (statsToBeRetrieved.contains(statName)) {
+                clusterStats.put(statName, adStats.getStats().get(statName).getValue());
```
wnbts, line 220: Minor. Seems more consistent to use adStats.getClusterStats() as in the for loop.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStatsNodesAction.java
```diff
@@ -18,18 +18,18 @@
...
 /**
- * ADStatsAction class
+ * ADStatsNodesAction class
  */
-public class ADStatsAction extends ActionType<ADStatsResponse> {
+public class ADStatsNodesAction extends ActionType<ADStatsNodesResponse> {
 
-    public static final ADStatsAction INSTANCE = new ADStatsAction();
+    public static final ADStatsNodesAction INSTANCE = new ADStatsNodesAction();
     public static final String NAME = "cluster:admin/ad_stats_action";
```
wnbts, line 26: minor. should ad_stats_action be renamed to ad_stats_node_action ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -115,4 +131,111 @@ private ADStatsRequest getRequest(RestRequest request) {
...
     }
+
+    /**
+     * Make the 2 requests to get the node and cluster statistics
+     *
+     * @param client Client
+     * @param channel Channel to send response
+     * @param adStatsRequest Request containing stats to be retrieved
+     */
+    public void getStats(Client client, RestChannel channel, ADStatsRequest adStatsRequest) {
```
wnbts, line 142: minor. The get methods should be private because they are implementation not contract. Tests are done based on contract not implementation. Because implementation is not client-facing, the full documentation is optional.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
wnbts: a different way might be iterating the entries using map::entrySet

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -115,4 +131,111 @@ private ADStatsRequest getRequest(RestRequest request) {
...
+                    adStatsResponse.setClusterStats(getClusterStats(adStatsRequest));
+                    listener.onResponse(adStatsResponse);
+                }, e -> listener.onFailure(new RuntimeException("Failed to get AD cluster stats: " + e))));
+            } else {
+                adStats.getStat(StatNames.DETECTOR_COUNT.getName()).setValue(0L);
+                adStatsResponse.setClusterStats(getClusterStats(adStatsRequest));
+                listener.onResponse(adStatsResponse);
+            }
+        } else {
+            adStatsResponse.setClusterStats(getClusterStats(adStatsRequest));
```
kaituo, line 184: Since you didn't set a value here, would it be possible that we pick some value set previously?

src/main/java/com/amazon/opendistroforelasticsearch/ad/stats/ADStatsResponse.java
jmazanec15: Good point will fix

src/main/java/com/amazon/opendistroforelasticsearch/ad/stats/ADStatsResponse.java
jmazanec15: For this object, the expected behavior if `other` has both node response and cluster stats is to merge them both into the object.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
jmazanec15: Good suggestion

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStatsNodesAction.java
jmazanec15: Yes it should good catch.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
jmazanec15: Good point. I will update them.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -115,4 +131,111 @@ private ADStatsRequest getRequest(RestRequest request) {
...
+     */
+    public void getStats(Client client, RestChannel channel, ADStatsRequest adStatsRequest) {
+        // Use MultiResponsesDelegateActionListener to execute 2 async requests and create the response once they finish
+        MultiResponsesDelegateActionListener<ADStatsResponse> delegateListener = new MultiResponsesDelegateActionListener<>(
+            getRestStatsListener(channel),
+            2,
+            "Unable to return AD Stats"
+        );
+
+        onGetClusterStats(client, delegateListener, adStatsRequest);
```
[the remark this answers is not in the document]
jmazanec15: Good point. Will update both.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -115,4 +131,111 @@ private ADStatsRequest getRequest(RestRequest request) {
...
+     */
+    public void onGetClusterStats(
+        Client client,
+        MultiResponsesDelegateActionListener<ADStatsResponse> listener,
+        ADStatsRequest adStatsRequest
+    ) {
+        ADStatsResponse adStatsResponse = new ADStatsResponse();
+        if (adStatsRequest.getStatsToBeRetrieved().contains(StatNames.DETECTOR_COUNT.getName())) {
+            if (clusterService.state().getRoutingTable().hasIndex(AnomalyDetector.ANOMALY_DETECTORS_INDEX)) {
+                IndicesStatsRequest indicesStatsRequest = new IndicesStatsRequest();
```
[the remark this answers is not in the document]
jmazanec15: Right that is correct. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -115,4 +131,111 @@ private ADStatsRequest getRequest(RestRequest request) {
...
+                client.execute(IndicesStatsAction.INSTANCE, indicesStatsRequest, ActionListener.wrap(indicesStatsResponse -> {
+                    adStats
+                        .getStat(StatNames.DETECTOR_COUNT.getName())
+                        .setValue(indicesStatsResponse.getIndex(AnomalyDetector.ANOMALY_DETECTORS_INDEX).getPrimaries().docs.getCount());
+                    adStatsResponse.setClusterStats(getClusterStats(adStatsRequest));
+                    listener.onResponse(adStatsResponse);
+                }, e -> listener.onFailure(new RuntimeException("Failed to get AD cluster stats: " + e))));
```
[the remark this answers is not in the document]
jmazanec15: Got it. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
jmazanec15: This `else` block is only entered if the user does not want to get the detector count stat. So, it will not be retrieved.

src/main/java/com/amazon/opendistroforelasticsearch/ad/stats/ADStatsResponse.java
kaituo: We don't want to use instanceof since this makes a subclass relationship returns true.  We want to have exact same class.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
kaituo: Got it.  Is SettableSupplier's value a singleton shared by all of the requests?  Would we have concurrent issue like one thread is writing to it while another is reading it?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
```diff
@@ -115,4 +131,111 @@ private ADStatsRequest getRequest(RestRequest request) {
...
+     *
+     * @param client Client
+     * @param channel Channel to send response
+     * @param adStatsRequest Request containing stats to be retrieved
+     */
+    public void getStats(Client client, RestChannel channel, ADStatsRequest adStatsRequest) {
+        // Use MultiResponsesDelegateActionListener to execute 2 async requests and create the response once they finish
+        MultiResponsesDelegateActionListener<ADStatsResponse> delegateListener = new MultiResponsesDelegateActionListener<>(
+            getRestStatsListener(channel),
+            2,
```
kaituo, line 146: Is it always 2?  If customers don't want to get detector count or only want to get detector count, this value should be 1.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStatsNodesAction.java
kaituo: all of our transport actions are using prefix "cluster:admin/ad" like:

"cluster:admin/ad/result"
"cluster:admin/ad/cron"
"cluster:admin/ad/model/delete"

It would be good to have the same style.

jmazanec15: Good call out. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/ad/stats/ADStatsResponse.java
jmazanec15: That makes sense, will change back.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
jmazanec15: Yes SettableSupplier's value is a singleton. SettableSupplier uses `AtomicLong` for thread safety.

I believe this is the scenario of concern:
1. async request(req_1) to get detector count is made (detector count = 3)
2. detector is created (detector count = 4)
3. async request(req_2) to get detector count is made (detector count = 4)
4. req_2 returns and sets detector count to 4
5. req_1 returns and sets detector count to 3
6. stats are returned for req_1 with detector count = 3
7. stats are returned for req_2 with detector count = 3

I think this is a minor issue for a couple of reasons. (1) On each stats request, the detector count is updated so it will not put the system in a permanent state of inconsistency (2) The scenario above is low probability because detector count is not expected to be that volatile and also stats api calls are not expected to be in that tight of a window.

A workaround could be to pass the detector count to `getClusterStatsMap(ADStatsRequest adStatsRequest)` and have a special case for DetectorCount stat. However, I am not sure this is best practice. Do you have any other suggestions?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
jmazanec15: Yes it is always 2. We still call the listeners onResponse method even if the calls are not made. For example, [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/108/files#diff-f01f65f739290c039e9696995a9acd36R181)

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestStatsAnomalyDetectorAction.java
kaituo: Agree it is a minor issue since document count is only used by humans, right?

Can you create a github issue for this?

jmazanec15: Yes only humans will read.

#112

PR #114, opened 2020-05-06
Stats API: moved detector count call outside transport layer and make…

jmazanec15:
… asynchronous

*Issue #, if available:*
#107 

*Description of changes:*
Backporting changes made in #108 

Confirmed `./gradlew build` passes and manual tests work for single node and multinode test clusters.


src/test/java/com/amazon/opendistroforelasticsearch/ad/stats/ADStatsResponseTests.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
yizheliu-amazon, line 2: maybe change to 2020. same comment for below.

jmazanec15: Good call out. Will update.

Issue #111, opened 2020-05-06
Profile API misses error info of failed detector after result index rolls over


yizheliu-amazon:
### Issue:
Given detector fails initialization, Profile API returns error info like "all features are disabled". However, after a while, calling Profile API again won't return error info, only "Stopped" state.

### Impact:
UX shows wrong state/message because of this. UX relies on the error info to show appropriate error message, and failure state. Detector failed initialization has "Stopped" state + error info, while normal "stopped" detector has "Stopped" state + empty error info. UX uses error info to show appropriate error state message. 

### Root cause:
This is because of that Profile API of AD plugin gets error info from `.opendistro-anomaly-results` index, which rolls over every 12 hours.

PR #117, opened 2020-05-08
Improve profile API's error fetching efficiency

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/111

*Description of changes:*

Previously, profile API scans all anomaly result indices to get a detector's most recent error, which can cause performance bottleneck with large anomaly result indices. This PR improves this aspect via various efforts.

First, when a detector is running, we only need to scan the current index, not all of the rolled over ones since we are interested in the latest error.
Second, when a detector is disabled, we only need to scan the latest anomaly result indices created before the detector's disabled time.
Third, setting track total hits false makes ES terminate search early. ES will not try to count the number of documents and will be able to end the query as soon as N document have been collected per segment.

Testing done:
1. patched a cluster with 1,000 detectors and 2GB anomaly result indices. Without the PR, scanning anomaly result indices 1000 times would timeout after 30 seconds. After the PR, we would not see the timeout.
2. A detector's error message can be on a rotated index. Adds a test case to makes sure we get error info from .opendistro-anomaly-results index that has been rolled over.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
@@ -20,7 +20,16 @@ import static com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetectorJob.
 import static org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
 import java.io.IOException;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.TimeZone;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -29,8 +38,11 @@ import org.elasticsearch.action.get.GetRequest;
 import org.elasticsearch.action.get.GetResponse;
 import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
+import org.elasticsearch.action.support.IndicesOptions;
 import org.elasticsearch.client.Client;
+import org.elasticsearch.cluster.metadata.IndexNameExpressionResolver;
 import org.elasticsearch.cluster.node.DiscoveryNode;
+import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
 import org.elasticsearch.common.xcontent.NamedXContentRegistry;
 import org.elasticsearch.common.xcontent.XContentParseException;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -234,20 +260,81 @@ private void profileState(
...
+            Map<Long, List<String>> candidateIndices = new HashMap<>();
+            for (String indexName : concreteIndices) {
+                Matcher m = Pattern.compile("\\.opendistro-anomaly-results-history-(\\d{4})\\.(\\d{2})\\.(\\d{2})-\\d+").matcher(indexName);
+                if (m.matches()) {
+                    int year = Integer.parseInt(m.group(1));
+                    int month = Integer.parseInt(m.group(2));
+                    int date = Integer.parseInt(m.group(3));
+                    // month starts with 0
+                    calendar.clear();
+                    calendar.set(year, month - 1, date);
```
yizheliu-amazon, line 308: will Jan. cause any issue? I guess in case of Jan., month is 1, not sure if this can cause any issue

yizheliu-amazon: I would suggest you use current year/month/date to initialize calendar, and do `calendar.add(Calendar.MONTH, -1)` instead.

yizheliu-amazon: > will Jan. cause any issue? I guess in case of Jan., month is 1, not sure if this can cause any issue

just checked java doc. month for Jan is 0 here.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -234,20 +260,81 @@ private void profileState(
...
+                    // 2020.05.08 is translated to 1588896000000
+                    long timestamp = calendar.getTimeInMillis();
+
+                    // a candidate index can be created before or after enabled time, but the index is definitely created before disabled
+                    // time
+                    if (timestamp <= disabledTimeMillis && maxTimestamp <= timestamp) {
+                        maxTimestamp = timestamp;
+                        // we can have two rotations on the same day and we don't know which one has our data, so we keep all
```
ylwu-amzn, line 316: One edge case: suppose detector interval is 1 minute
1.Detector still running at `2020-05-07, 11:59:50 PM`, then AD result indices rolled over as `.opendistro-anomaly-results-history-2020.05.07-001`
2.Detector next run will be `2020-05-08, 00:00:50 AM`. If user stop the detector at  `2020-05-08 00:00:10 AM`, detector will not have AD result on `2020-05-08`. 
So if we will check latest AD result index on `2020-05-08` as  `2020-05-08 <= 2020-05-08 00:00:10 AM`. But we can't find any AD result for this detector. How about we check last two day's AD result indices to make sure we can always get AD result? Very similarly to set monitor interval as `2*detector_interval`

kaituo: Good point.  Changed.

Issue #153, opened 2020-06-15
[BUG] Exception when retrying saving anomaly result 


kaituo:
**Describe the bug**
[2020-06-12T14:13:02,339][ERROR][c.a.o.a.t.h.AnomalyResultHandler] [ab9f752a8eee0455180aa6feb9b2a65a] Fail to save anomaly index: ID j25fqnIBggidxd5MoOFB from 2020-06-12T21:12:00.795255Z to 2020-06-12T21:13:00.795255Z
java.lang.IllegalArgumentException: autoGeneratedTimestamp should not be set externally
        at org.elasticsearch.action.bulk.TransportBulkAction.doExecute(TransportBulkAction.java:179) ~[elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.bulk.TransportBulkAction.doExecute(TransportBulkAction.java:97) ~[elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:154) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.support.TransportAction.execute(TransportAction.java:130) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.bulk.TransportSingleItemBulkWriteAction.doExecute(TransportSingleItemBulkWriteAction.java:52) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.bulk.TransportSingleItemBulkWriteAction.doExecute(TransportSingleItemBulkWriteAction.java:36) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.support.TransportAction$RequestFilterChain.proceed(TransportAction.java:154) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.support.TransportAction.execute(TransportAction.java:130) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.action.support.TransportAction.execute(TransportAction.java:65) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.client.node.NodeClient.executeLocally(NodeClient.java:83) [elasticsearch-7.7.0.jar:7.7.0]
        at org.elasticsearch.client.node.NodeClient.doExecute(NodeClient.java:72) [elasticsearch-7.7.0.jar:7.7.0]
[...]

PR #154, opened 2020-06-15
Fix retrying saving anomaly results

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/153

*Description of changes:*
Previously, AD throws an exception when retrying saving an anomaly result, indicating autoGeneratedTimestamp should not be set. We can see the exception because we reuse the same IndexRequest, where autoGeneratedTimestamp is already set. This PR creates a new IndexRequest by copying the original request without autoGeneratedTimestamp.

This PR also adds detector id to AnomalyDetectionException’s toString method so that we can see the id in the log. This PR also adds more logging to SearchFeatureDao.

Testing done:
* Manually test to make sure the exception is gone after my fix.


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/AnomalyResultHandler.java
```diff
@@ -192,7 +192,12 @@ void saveDetectorResult(IndexRequest indexRequest, String context, Iterator<Time
             } else {
                 TimeValue nextDelay = backoff.next();
                 LOG.warn(RETRY_SAVING_ERR_MSG + context, cause);
-                threadPool.schedule(() -> saveDetectorResult(indexRequest, context, backoff), nextDelay, ThreadPool.Names.SAME);
+                // copy original request's source without other information like autoGeneratedTimestamp
+                // otherwise, an exception will be thrown indicating autoGeneratedTimestamp should not be set
+                // while request id is already set (id is set because we have already sent the request before).
+                IndexRequest newReuqest = new IndexRequest(AnomalyResult.ANOMALY_RESULT_INDEX);
```
wnbts, line 198: Minor. Should the index name be from the original index request?

kaituo: This is AnomalyResultHandler, so its index should be AnomalyResult.ANOMALY_RESULT_INDEX.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -367,6 +380,7 @@ public void getFeaturesForSampledPeriods(
         ActionListener<Optional<Entry<double[][], Integer>>> listener
     ) {
         Map<Long, double[]> cache = new HashMap<>();
+        logger.info(String.format("Getting features for detector %s starting %d", detector.getDetectorId(), endTime));
```
yizheliu-amazon, line 383: `starting %d` -> `ending at %d`?

kaituo: Good catch.  we don't know the start time at this point.  Changed to "logger.info(String.format("Getting features for detector %s ending at %d", detector.getDetectorId(), endTime));"

Issue #148, opened 2020-06-10
Add Initialization progress in Profile API


kaituo:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #164, opened 2020-06-19
Adds initialization progress to profile API

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/148

*Description of changes:*
This PR adds init_progress to profile API. init_progress helps users track initialization percentage, needed shingles, and estimated time to go if the future data stream is continuous (no missing data).

Initialization percentage measures how far away we can observe RCF models emitting scores.  The implementation fetches the RCF model's total updates while the AD job is running and materializes the value to the newly added index .opendistro-anomaly-info. Total updates record the number of times this RCF model has been updated. 

Initialization percent is computed as x/128: 
* if total updates > 128, x = 128. Otherwise, x is the total updates 
* 128 is our output after the number in RCF. After observing 128 samples, RCF starts emitting scores.

Needed shingles are computed as 128 -x.  Estimated minutes to go is computed as needed shingles * detector interval

This PR also materializes the error message in the most recent run to speed up profile API's error fetching.

During each AD execution, we also check if a checkpoint is there (the result is saved and maintained as other AD states), if yes, we cold start immediately.

Testing done:
1. adds unit tests
2. run e2e tests to verify init_progress number makes sense.

Sample output:

```
kaituo@dev-dsk-kaituo-2a-74e3f324.us-west-2: ~/code/github/anomaly-detection
% curl -X GET "localhost:9200/_opendistro/_anomaly_detection/detectors/eT6_yXIBegfEFVlvLSF3/_profile?_all=true&pretty"
{
  "state" : "INIT",
  "models" : [
    {
      "model_id" : "eT6_yXIBegfEFVlvLSF3_model_rcf_2",
      "model_size_in_bytes" : 2228224,
      "node_id" : "4l_Pb26iRLWrFydpEhVBsA"
    },
    {
      "model_id" : "eT6_yXIBegfEFVlvLSF3_model_rcf_1",
      "model_size_in_bytes" : 2228224,
      "node_id" : "09xMkHzuSqOEIyn__syHUQ"
    },
    {
      "model_id" : "eT6_yXIBegfEFVlvLSF3_model_rcf_0",
      "model_size_in_bytes" : 2228224,
      "node_id" : "09xMkHzuSqOEIyn__syHUQ"
    },
    {
      "model_id" : "eT6_yXIBegfEFVlvLSF3_model_threshold",
      "node_id" : "09xMkHzuSqOEIyn__syHUQ"
    }
  ],
  "shingle_size" : 7,
  "coordinating_node" : "4l_Pb26iRLWrFydpEhVBsA",
  "total_size_in_bytes" : 6684672,
  "init_progress" : {
    "percentage" : "70%",
    "estimated_minutes_left" : 77,
    "needed_shingles" : 77
  }
}
[2020-06-18 16:31:32 PDT] [2020-06-18 23:31:32 UTC]
kaituo@dev-dsk-kaituo-2a-74e3f324.us-west-2: ~/code/github/anomaly-detection
% curl -X GET "localhost:9200/_opendistro/_anomaly_detection/detectors/eT6_yXIBegfEFVlvLSF3/_profile/init_progress,state,error"
{"state":"INIT","init_progress":{"percentage":"70%","estimated_minutes_left":77,"needed_shingles":77}}%
```


kaituo:
> Two general comments.
> 
>     1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
The idea is to put all state related to an detector to the info index.  As you said, additional detector info field will be added to the info index if required. 
 
>     2. After initialization is completed, the status is changed and collecting data for computing initializatio progress should stop. Collecting data and computing can be done on request instead.
I prefer to keep collecting after the status is changed:
First, we need extra logic to keep track of detector state.  For example, a detector is stopped and its init_progress should be recalculated (we don't know its models are deleted or not, its detector definition is updated or not).  Tracking the state is not trivial, and we should not put the burden on ModelManager or AnomalyResultTransportAction.
Second, tracking can help us debugging when we suspect whether the model is not working or not.

wnbts:
> > Two general comments.
> > ```
> > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > ```
> 
> The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.

I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.

wnbts:
> > 2. After initialization is completed, the status is changed and collecting data for computing initializatio progress should stop. Collecting data and computing can be done on request instead.
> > ```
> 
> I prefer to keep collecting after the status is changed:
> First, we need extra logic to keep track of detector state. For example, a detector is stopped and its init_progress should be recalculated (we don't know its models are deleted or not, its detector definition is updated or not). Tracking the state is not trivial, and we should not put the burden on ModelManager or AnomalyResultTransportAction.
> Second, tracking can help us debugging when we suspect whether the model is not working or not.

Tracking the progress can be done on request when the get profile api is called, correct?

kaituo:
> > > Two general comments.
> > > ```
> > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > ```
> > 
> > 
> > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> 
> I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.

We should use an index based on its usage.  Detector index is for detector definition, not intermediate state.  That's why we have checkpoint, job, result index.  Currently, info index is created to store intermediate state, not just init_progress related.  For example, we can implement custom routing and store the routing table in the info index.  Custom routing allows more flexible routing like allocating a model on a node with the most memory available.

kaituo:
> > > 1. After initialization is completed, the status is changed and collecting data for computing initializatio progress should stop. Collecting data and computing can be done on request instead.
> > 
> > 
> > I prefer to keep collecting after the status is changed:
> > First, we need extra logic to keep track of detector state. For example, a detector is stopped and its init_progress should be recalculated (we don't know its models are deleted or not, its detector definition is updated or not). Tracking the state is not trivial, and we should not put the burden on ModelManager or AnomalyResultTransportAction.
> > Second, tracking can help us debugging when we suspect whether the model is not working or not.
> 
> Tracking the progress can be done on request when the get profile api is called, correct?

That requires a broadcast call to all nodes since we don't know which node has which model.  Kibana would call this api every few minutes, which is costly.

wnbts:
> > > > Two general comments.
> > > > ```
> > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > ```
> > > 
> > > 
> > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > 
> > 
> > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> 
> We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.

is the new index certainly needed? if detector index is to keep unchanged, what about the job index so all the execution related details/job info is managed in one index rather than scattered across multiple indices?

wnbts:
> > > > 1. After initialization is completed, the status is changed and collecting data for computing initializatio progress should stop. Collecting data and computing can be done on request instead.
> > > 
> > > 
> > > I prefer to keep collecting after the status is changed:
> > > First, we need extra logic to keep track of detector state. For example, a detector is stopped and its init_progress should be recalculated (we don't know its models are deleted or not, its detector definition is updated or not). Tracking the state is not trivial, and we should not put the burden on ModelManager or AnomalyResultTransportAction.
> > > Second, tracking can help us debugging when we suspect whether the model is not working or not.
> > 
> > 
> > Tracking the progress can be done on request when the get profile api is called, correct?
> 
> That requires a broadcast call to all nodes since we don't know which node has which model. Kibana would call this api every few minutes, which is costly.

it doesn't need to be broadcast. 
Most of the time, the models are initialized, the result should be stored in the index so there is no need to recompute.

If there is no stored initialization results, the model can be located using hash or the handler just directly pulls the first partition to get the status and save it for future queries.

The idea is to only do this work when it's needed and applicable.

yizheliu-amazon:
> > > > Two general comments.
> > > > ```
> > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > ```
> > > 
> > > 
> > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > 
> > 
> > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> 
> We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.

In this case I think we can name the index like `.opendistro-anomaly-state`, `info` sounds very general.

kaituo:
> > > > > Two general comments.
> > > > > ```
> > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > ```
> > > > 
> > > > 
> > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > 
> > > 
> > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > 
> > 
> > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> 
> is the new index certainly needed? if detector index is to keep unchanged, what about the job index so all the execution related details/job info is managed in one index rather than scattered across multiple indices?

I feel so.  We cannot modify job index as job scheduler listens to any changes in the index.  They would schedule a new job if any doc changes there.

kaituo:
> > > > > 1. After initialization is completed, the status is changed and collecting data for computing initializatio progress should stop. Collecting data and computing can be done on request instead.
> > > > 
> > > > 
> > > > I prefer to keep collecting after the status is changed:
> > > > First, we need extra logic to keep track of detector state. For example, a detector is stopped and its init_progress should be recalculated (we don't know its models are deleted or not, its detector definition is updated or not). Tracking the state is not trivial, and we should not put the burden on ModelManager or AnomalyResultTransportAction.
> > > > Second, tracking can help us debugging when we suspect whether the model is not working or not.
> > > 
> > > 
> > > Tracking the progress can be done on request when the get profile api is called, correct?
> > 
> > 
> > That requires a broadcast call to all nodes since we don't know which node has which model. Kibana would call this api every few minutes, which is costly.
> 
> it doesn't need to be broadcast.
> Most of the time, the models are initialized, the result should be stored in the index so there is no need to recompute.
> 
> If there is no stored initialization results, the model can be located using hash or the handler just directly pulls the first partition to get the status and save it for future queries.
> 
> The idea is to only do this work when it's needed and applicable.

Good idea.  I will do this then.  I still need the info index to store error information though.

kaituo:
> > > > > Two general comments.
> > > > > ```
> > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > ```
> > > > 
> > > > 
> > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > 
> > > 
> > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > 
> > 
> > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> 
> In this case I think we can name the index like `.opendistro-anomaly-state`, `info` sounds very general.

We can do that.  Yaliang may add sth in the index for his workbench.  Will collect his idea on the name as well.

wnbts:
> > > > > > Two general comments.
> > > > > > ```
> > > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > > ```
> > > > > 
> > > > > 
> > > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > > 
> > > > 
> > > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > > 
> > > 
> > > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> > 
> > 
> > is the new index certainly needed? if detector index is to keep unchanged, what about the job index so all the execution related details/job info is managed in one index rather than scattered across multiple indices?
> 
> I feel so. We cannot modify job index as job scheduler listens to any changes in the index. They would schedule a new job if any doc changes there.

ok. a new index it is then.

one minor thing regarding naming, i in general avoid names like info/data for obvious reasons in software. state is good, or i wouldn't mind a name like internal state to be clear and explicit.

kaituo:
> > > > > > > Two general comments.
> > > > > > > ```
> > > > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > > > ```
> > > > > > 
> > > > > > 
> > > > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > > > 
> > > > > 
> > > > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > > > 
> > > > 
> > > > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> > > 
> > > 
> > > is the new index certainly needed? if detector index is to keep unchanged, what about the job index so all the execution related details/job info is managed in one index rather than scattered across multiple indices?
> > 
> > 
> > I feel so. We cannot modify job index as job scheduler listens to any changes in the index. They would schedule a new job if any doc changes there.
> 
> ok. a new index it is then.
> 
> one minor thing regarding naming, i in general avoid names like info/data for obvious reasons in software. state is good, or i wouldn't mind a name like internal state to be clear and explicit.

What're the "obvious reasons in software"?

wnbts:
> > > > > > > > Two general comments.
> > > > > > > > ```
> > > > > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > > > > ```
> > > > > > > 
> > > > > > > 
> > > > > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > > > > 
> > > > > > 
> > > > > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > > > > 
> > > > > 
> > > > > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> > > > 
> > > > 
> > > > is the new index certainly needed? if detector index is to keep unchanged, what about the job index so all the execution related details/job info is managed in one index rather than scattered across multiple indices?
> > > 
> > > 
> > > I feel so. We cannot modify job index as job scheduler listens to any changes in the index. They would schedule a new job if any doc changes there.
> > 
> > 
> > ok. a new index it is then.
> > one minor thing regarding naming, i in general avoid names like info/data for obvious reasons in software. state is good, or i wouldn't mind a name like internal state to be clear and explicit.
> 
> What're the "obvious reasons in software"?

in software, code directs data. one should be surprised if a domain term is not data/info. Consider a book object vs a book data object. there is no information gain in stating the obvious.

kaituo:
> > > > > > Two general comments.
> > > > > > ```
> > > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > > ```
> > > > > 
> > > > > 
> > > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > > 
> > > > 
> > > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > > 
> > > 
> > > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> > 
> > 
> > In this case I think we can name the index like `.opendistro-anomaly-state`, `info` sounds very general.
> 
> We can do that. Yaliang may add sth in the index for his workbench. Will collect his idea on the name as well.

Changed to .opendistro-anomaly-state.

ylwu-amzn:
> > > > > > > Two general comments.
> > > > > > > ```
> > > > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > > > ```
> > > > > > 
> > > > > > 
> > > > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > > > 
> > > > > 
> > > > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > > > 
> > > > 
> > > > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> > > 
> > > 
> > > In this case I think we can name the index like `.opendistro-anomaly-state`, `info` sounds very general.
> > 
> > 
> > We can do that. Yaliang may add sth in the index for his workbench. Will collect his idea on the name as well.
> 
> Changed to .opendistro-anomaly-state.

Sudipto has some comments on AD workbench, will discuss with him today. How about we split this big code change into several smaller ones? I think we need more discussion about the new index. How about we make changes to support progress bar based on current AD result index to mitigate long initialization problem first?

kaituo:
> > > > > > > > Two general comments.
> > > > > > > > ```
> > > > > > > > 1. Does additional detector info require an additional data index? An additional internal field in detector is a better option for it requires less resources (no new index or redundant data fields) and easier to manage (less consistency issues from multiple indices).
> > > > > > > > ```
> > > > > > > 
> > > > > > > 
> > > > > > > The idea is to put all state related to an detector to the info index. As you said, additional detector info field will be added to the info index if required.
> > > > > > 
> > > > > > 
> > > > > > I think we should weigh the two options carefully. 1. creating a new index and 2. adding a new field to existing index. From what I have seen, option #2 is the more commonly used pattern in es and it's more efficient and easier to manage.
> > > > > 
> > > > > 
> > > > > We should use an index based on its usage. Detector index is for detector definition, not intermediate state. That's why we have checkpoint, job, result index. Currently, info index is created to store intermediate state, not just init_progress related. For example, we can implement custom routing and store the routing table in the info index. Custom routing allows more flexible routing like allocating a model on a node with the most memory available.
> > > > 
> > > > 
> > > > In this case I think we can name the index like `.opendistro-anomaly-state`, `info` sounds very general.
> > > 
> > > 
> > > We can do that. Yaliang may add sth in the index for his workbench. Will collect his idea on the name as well.
> > 
> > 
> > Changed to .opendistro-anomaly-state.
> 
> Sudipto has some comments on AD workbench, will discuss with him today. How about we split this big code change into several smaller ones? I think we need more discussion about the new index. How about we make changes to support progress bar based on current AD result index to mitigate long initialization problem first?

We cannot support progress bar based on current AD result index.  What's the undecided besides index name?  We can easily change name for that.

What's an easy way to split this?  Half of the PR are tests.  I have mainly three parts: one is the change to store to new index , one is new transport action to support fetching rcf updates, another part is to changes in the consumer (profile api).  If I split in this way, one PR has roughly 1000 lines, another PR has 600 lines, another PR has 1400 lines.  Are you good with that?

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -229,6 +251,18 @@ public void initAnomalyDetectorJobIndex(ActionListener<CreateIndexResponse> acti
         adminClient.indices().create(request, actionListener);
     }
 
+    /**
+     * Create an index.
+     *
+     * @param actionListener action called after create index
+     * @throws IOException IOException from {@link AnomalyDetectionIndices#getAnomalyDetectorJobMappings}
```
wnbts, line 258: remark. the issue of checked exception is clear here. they bubble up from low levels and raise unhelpful concerns/threats to higher-level in the chain, which, in this case, choose to expose unnecessary implementation.

when there is an unlikely checked exception due to lower-level implementation and the caller is unlikely to act or recover, it's likely a good place to wrap the checked exceptions early on.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -1039,4 +1039,18 @@ private double computeRcfConfidence(RandomCutForest forest) {
...
         return res;
     }
+
+    /**
+      * Get all RCF partition's size corresponding to a detector.  Thresholding
+      *  models' size is a constant since they are small in size (KB).
+      * @param modelId model id
+      */
+    public long getTotalUpdates(String modelId) {
+        ModelState<RandomCutForest> model = forests.get(modelId);
```
wnbts, line 1049: issue. what happens to a completed but not hosted model (maybe after detector is stopped)?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/InitProgressProfile.java
```diff
@@ -0,0 +1,147 @@
...
+import org.elasticsearch.common.xcontent.ToXContent;
+import org.elasticsearch.common.xcontent.XContentBuilder;
+
+/**
+ * Profile output for detector initialization progress.  When the new detector is created, it is possible that
+ * there hasn’t been enough continuous data in the index.  We need to use live data to initialize.
+ *   During initialization, we need to tell users progress (using a percentage), how many more
+ *    shingles to go, and approximately how many minutes before the detector becomes operational
+ *     if they keep their data stream continuous.
+ * @author kaituo
```
wnbts, line 50: minor. unneeded annotation.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/InitProgressProfile.java
```diff
@@ -0,0 +1,147 @@
...
+import org.elasticsearch.common.io.stream.Writeable;
+import org.elasticsearch.common.xcontent.ToXContent;
+import org.elasticsearch.common.xcontent.XContentBuilder;
+
+/**
+ * Profile output for detector initialization progress.  When the new detector is created, it is possible that
+ * there hasn’t been enough continuous data in the index.  We need to use live data to initialize.
+ *   During initialization, we need to tell users progress (using a percentage), how many more
+ *    shingles to go, and approximately how many minutes before the detector becomes operational
+ *     if they keep their data stream continuous.
```
wnbts, line 49: minor. indentation is off.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: I did it. See ThrowingConsumerWrapper.throwingConsumerWrapper.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -1039,4 +1039,18 @@ private double computeRcfConfidence(RandomCutForest forest) {
             .forEach(entry -> { res.put(entry.getKey(), 0L); });
         return res;
     }
+
+    /**
+      * Get all RCF partition's size corresponding to a detector.  Thresholding
+      *  models' size is a constant since they are small in size (KB).
```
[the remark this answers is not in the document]
kaituo: good catch. Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: We return 0 in this case.  Any other better way?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/DetectorInternalState.java
```diff
@@ -0,0 +1,186 @@
...
+        return Objects.hashCode(rcfUpdates, lastUpdateTime, error);
+    }
+
+    @Override
+    public Object clone() {
+        DetectorInternalState state = null;
+        try {
+            state = (DetectorInternalState) super.clone();
+        } catch (CloneNotSupportedException e) {
+            state = new DetectorInternalState.Builder().rcfUpdates(rcfUpdates).lastUpdateTime(lastUpdateTime).error(error).build();
```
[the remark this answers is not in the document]
kaituo: fixed

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/InitProgressProfile.java
kaituo: fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/InitProgressProfile.java
kaituo: removed

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -52,6 +53,7 @@
     private static final Logger LOG = LogManager.getLogger(ADStateManager.class);
     private ConcurrentHashMap<String, Entry<AnomalyDetector, Instant>> currentDetectors;
     private ConcurrentHashMap<String, Entry<Integer, Instant>> partitionNumber;
+    private ConcurrentHashMap<String, Instant> currentCheckpoints;
```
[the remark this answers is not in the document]
kaituo: Could you explain your question?

This is for starting cold start whenever there is missing checkpoint.  The key is detector id, while the value is the timestamp when we find a checkpoint.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -140,6 +143,30 @@ public void getAnomalyDetector(String adID, ActionListener<Optional<AnomalyDetec
         }, listener::onFailure);
     }
 
+    public void getDetectorCheckpoint(String adID, ActionListener<Boolean> listener) {
```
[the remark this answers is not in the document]
kaituo: doc is deprecated.  We should remove doc in checkpointdao.  Do you want me to do that?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -176,6 +205,27 @@ public void maintenance() {
...
     }
 
+    /**
+     * Clean states if it is older than our stateTtl. The input has to be a
+     * ConcurrentHashMap otherwise we will have
+     * java.util.ConcurrentModificationException.
+     *
+     * @param flags flags to be maintained
+     */
+    void maintenanceFlag(ConcurrentHashMap<String, Instant> flags) {
```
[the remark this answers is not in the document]
kaituo: I wrote unit tests for it.  That's why I am not using private.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -287,6 +291,20 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
             }
 
             if (!featureOptional.getProcessedFeatures().isPresent()) {
```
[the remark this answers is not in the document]
kaituo: other reasons could be: circuit breaker is open, cluster is under read/write block, or AD is disabled.  We should not trigger cold start in those cases.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectorStateHandler.java
```diff
@@ -0,0 +1,175 @@
...
+            createIndex,
+            indexExists,
+            true,
+            clientUtil,
+            indexUtils,
+            clusterService
+        );
+    }
+
+    public void saveRcfUpdates(long totalRcfUpdates, String detectorId) {
```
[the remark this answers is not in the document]
kaituo: make sense.  I'll modify when you answered my previous question on whether to continue to use on-demand messaging.  Also, we should use 128, right?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ThrowingConsumerWrapper.java
```diff
@@ -0,0 +1,41 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.util;
+
+import java.util.function.Consumer;
+
+public class ThrowingConsumerWrapper {
+    /**
+     * Utility method to use a method throwing checked exception inside a function
```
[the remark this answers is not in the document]
kaituo: Could you explain your remark?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
wnbts: the code should load the checkpoint and return its count.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
wnbts: there are three maps, if their keys are detector ids, they can be merged into one map sharing the same keys.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
wnbts: i mean the doc for this getDetectorCheckpoint method is missing. a client/reader cannot know what this method does without reading through code.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
wnbts: try to avoid a. leaking implementation in class interface. and b. build brittle implementation-based tests.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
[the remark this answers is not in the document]
wnbts: ok. it might be good to explicitly put in code the conditions where cold start is suppressed so that no one needs to guess why initialization is not done

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ThrowingConsumerWrapper.java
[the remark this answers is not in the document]
wnbts: if underlying low-level code doesn't threaten with checked exceptions, this wouldn't be necessary, right?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectorStateHandler.java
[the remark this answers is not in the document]
wnbts: 128 is the current number

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -436,7 +442,8 @@ private void indexAnomalyResult(
                 Instant.now(),
                 response.getError()
             );
-            anomalyResultHandler.indexAnomalyResult(anomalyResult);
+            anomalyResultHandler.index(anomalyResult, detectorId);
+            detectorStateHandler.saveError(response.getError(), detectorId);
```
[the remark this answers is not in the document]
wnbts: same idea applies to error and count, a write only happens if there is an update to the read. if the error is null/the same, it is not needed to update.

regarding the last question, i suggest writes (due to the count field) only happen when the count is below 128 and new count is different (larger) than the old count. Messaging is one way to obtain the current count but i think loading the checkpoint locally to get the persisted count also suffices.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: make sense

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
kaituo: How do I merge it?  Their values are different.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
kaituo: added doc

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
kaituo: duly noted.  Having package private for tests is not uncommon.  People invented @VisibleForTesting annotation for that.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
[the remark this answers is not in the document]
kaituo: added doc

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ThrowingConsumerWrapper.java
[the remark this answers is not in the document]
kaituo: yes

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java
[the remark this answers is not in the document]
kaituo: Error needs to be up-to-date.  When a profile request comes in, it won't know this error is up-to-date or not unless the writer side guarantees it.  For example, if there is no error, a profile API's request comes in to ask for error, we would return an error instead of null.

About the in-progress number, how about the we always depending on on-demand messaging.  If there is rcf model in memory, we ask for it.  Otherwise, we ask for checkpoint.  We can remove index write for rcf updates.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -170,28 +175,34 @@ public AnomalyDetectorPlugin() {}
         IndexNameExpressionResolver indexNameExpressionResolver,
         Supplier<DiscoveryNodes> nodesInCluster
     ) {
-        AnomalyResultHandler anomalyResultHandler = new AnomalyResultHandler(
+
+        AnomalyIndexHandler<AnomalyResult> anomalyResultHandler;
```
ylwu-amzn, line 179: minor, can we rename `anomalyResultHandler ` as `anomalyIndexHandler`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/DetectorInternalState.java
```diff
@@ -0,0 +1,186 @@
...
+public class DetectorInternalState implements ToXContentObject, Cloneable {
+
+    public static final String PARSE_FIELD_NAME = "DetectorInternalState";
+    public static final NamedXContentRegistry.Entry XCONTENT_REGISTRY = new NamedXContentRegistry.Entry(
+        DetectorInternalState.class,
+        new ParseField(PARSE_FIELD_NAME),
+        it -> parse(it)
+    );
+
+    public static final String DETECTOR_STATE_INDEX = ".opendistro-anomaly-state";
```
ylwu-amzn, line 45: How about change to `.opendistro-anomaly-detection-state`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/DetectorInternalState.java
```diff
@@ -0,0 +1,186 @@
...
+
+    public static final String RCF_UPDATES_FIELD = "rcf_updates";
+    public static final String LAST_UPDATE_TIME_FIELD = "last_update_time";
+    public static final String ERROR_FIELD = "error";
+
+    private long rcfUpdates = -1L;
```
ylwu-amzn, line 51: Any special consideration for using `-1` as default value?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
kaituo: anomalyResultHandler seems more appropriate than anomalyIndexHandler since this handler is for anomaly result.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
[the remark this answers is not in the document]
ylwu-amzn: > good catch. Fixed.

Seems new revision not published yet?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/DetectorInternalState.java
kaituo: changed

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
ylwu-amzn: I think we only show progress bar when detector started. Once detector start, will restore model from checkpoint. Then we can read updates from hosted model. Should we pull the checkpoint again?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/DetectorInternalState.java
kaituo: want to make it a illegal value. 0 is legal value.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
wnbts: a manager should manage a set of assigned detectors so for an assigned detector id, there should be one key-vale pair of detector id, run state. the various parts can be put together into one structure rather than requiring separate key-value pairs for each field.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -148,11 +175,13 @@ public void getAnomalyDetector(String adID, ActionListener<Optional<AnomalyDetec
     public void clear(String adID) {
         currentDetectors.remove(adID);
         partitionNumber.remove(adID);
+        currentCheckpoints.remove(adID);
     }
 
     public void maintenance() {
         maintenance(currentDetectors);
         maintenance(partitionNumber);
+        maintenanceFlag(currentCheckpoints);
```
ylwu-amzn, line 184: Why need to periodically maintain the checkpoint flag?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
wnbts: that's better left for legacy code. for new code, it's better not to start with it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -140,6 +143,30 @@ public void getAnomalyDetector(String adID, ActionListener<Optional<AnomalyDetec
...
+
+        GetRequest request = new GetRequest(CommonName.CHECKPOINT_INDEX_NAME, modelManager.getRcfModelId(adID, 0));
+
+        clientUtil.<GetRequest, GetResponse>asyncRequest(request, client::get, onGetCheckpointResponse(adID, listener));
+    }
+
+    private ActionListener<GetResponse> onGetCheckpointResponse(String adID, ActionListener<Boolean> listener) {
+        return ActionListener.wrap(response -> {
+            if (response == null || !response.isExists()) {
+                listener.onResponse(Boolean.FALSE);
```
ylwu-amzn, line 162: If not find checkpoint, we should remove it from `currentCheckpoints `.

Another case is if user delete checkpoint manually, we should remove it from `currentCheckpoints`.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java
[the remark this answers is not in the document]
wnbts: regarding error, the main idea is to update only when it's changed. for example, error to null/non-error, null to error, or error 1 to error 2. if it's null to null, error 1 to error 1, no update is needed.

for in-progress number, i would like to see the info is obtained on demand to make sure it's current. two ramifications. first, if the state (count >= 128) is final, a query to index should be better than a transport + a checkpoint pull from index. second, if progress doesn't depend on index, does error alone justify index?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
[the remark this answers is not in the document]
kaituo: will create a class ADState then.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -287,6 +291,20 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
...
+                stateManager.getDetectorCheckpoint(adID, ActionListener.wrap(checkpointExists -> {
```
ylwu-amzn, line 294: For happy case(detector and feature present and circuit breaker not open, no read/write block, AD enabled), will every job run call `getDetectorCheckpoint ` method?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: otherwise, who would maintain the flag?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: We would come to this code path if currentCheckpoints has the flag set in the first place.

Currently, we don't delete checkpoint.  I am waiting for you to fix it :)

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: yes

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
ylwu-amzn: Sorry, not explain clearly. My question is why need "periodically" maintain. Can we just maintain it when update/delete happen?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java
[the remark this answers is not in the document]
kaituo: Updating error when it is changed works.  Will change.

about your ramifications: 
1) if we update when it is less than 128, why do we still need a transport + a checkpoint pull?
2) yes, Yaliang and folks will add things there as needed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: make sense.  We both need to remember to do this when we fix checkpoint deletion.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: AD will take snapshot of model and save as checkpoint every 1 hour? If one detector has 1 minute interval and no checkpoint when start, is it possible that first 59 detector runs can't get checkpoint and do cold start?

kaituo: we also take checkpoint during cold start.  So in your example, we have one checkpoint after the 1st run and the next one is taken on the 61st run.

ylwu-amzn: Cool, that makes sense.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: we pull checkpoint when we cannot get updates from hosted models.  We may not do this if use the state index.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/RCFPollingTransportAction.java
```diff
@@ -0,0 +1,130 @@
...
+import org.elasticsearch.transport.TransportRequestOptions;
+import org.elasticsearch.transport.TransportResponseHandler;
+import org.elasticsearch.transport.TransportService;
+
+import com.amazon.opendistroforelasticsearch.ad.cluster.HashRing;
+import com.amazon.opendistroforelasticsearch.ad.common.exception.AnomalyDetectionException;
+import com.amazon.opendistroforelasticsearch.ad.ml.ModelManager;
+import com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings;
+
+public class RCFPollingTransportAction extends HandledTransportAction<RCFPollingRequest, RCFPollingResponse> {
```
ylwu-amzn, line 43: Add some comment doc for this class?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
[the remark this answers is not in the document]
kaituo: yeah, I am waiting to reach an agreement with Lai: whether to use index state or transport + checkpoint pull.  I am ok with both.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: I remember why I did it last year: this adds fault tolerance for state cleanup.  For example, what if customers never delete checkpoint and keep the detector stopped.  Then the checkpoint state is kept there in limbo in memory forever.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
ylwu-amzn: > We wouldn't come to this code path if currentCheckpoints has the flag set in the first place.
> 
> Currently, we don't delete checkpoint. I am waiting for you to fix it :)

Sure, can you add some todo here?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
ylwu-amzn: How about we name the class as `AnomalyResultHandler` if it's for anomaly result only?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
```diff
@@ -176,6 +205,27 @@ public void maintenance() {
...
+     *
+     * @param flags flags to be maintained
+     */
+    void maintenanceFlag(ConcurrentHashMap<String, Instant> flags) {
+        flags.entrySet().stream().forEach(entry -> {
+            String detectorId = entry.getKey();
+            try {
+                Instant time = entry.getValue();
+                if (time.plus(stateTtl).isBefore(clock.instant())) {
+                    flags.remove(detectorId);
```
ylwu-amzn, line 221: We can save some memory if remove detector from `currentCheckpoints`, but needs another index query to get checkpoint as almost every detector run will try to get checkpoint. Can we just remove it when detector stopped/deleted?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java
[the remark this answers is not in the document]
wnbts: 1. when the count in the previously saved state is in [0, 128), in order to get the current count to a. return in response and b. update state with, there is two ways: 1. transport + checkpoint or 2. just checkpoint. (either is ok to me, latter is simpler, more reliable but slower to change, hourly or so).

when the count in the saved state is no less than 128, it is returned in response and done. transport or checkpoint query is no longer needed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: Created https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/183

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
kaituo: The class can be inherited by other classes writing to a different index.  We shouldn't do that.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: state manitenance would remove it from memory.  We don't have to do it manually.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
ylwu-amzn: Should we show progress before restoring model? For example, one detector has 10 minutes interval, when restart detector, we show progress 80% directly as we pull checkpoint (or state index)? Is it possible we fail to restore model and the progress still show as 80%?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
ylwu-amzn: > state manitenance would remove it from memory. We don't have to do it manually.

Yeah, the question is should we add some TTL to checkpoint cache and remove it periodically? Can we simplify the logic to just remove one detector's checkpoint cache when detector stopped/deleted?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: We need to pull checkpoint when shingle is not ready and model is in memory yet.  Sometimes this process is long, sometimes is short.  This is a tradeoff between code simplicity (maintenance/testability) vs accuracy.  What you suggested is what I did one week ago.  Now I changed my code to more complex and more accurate side.  There is no absolute right or wrong.  Let's stick to current complex solution.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADStateManager.java
kaituo: One case this periodic cleanup is useful: job schedule's shard relocate and we don't use the coordinating node anymore.  Before the detector gets deleted/stopped, these memory would be there.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/RCFPollingTransportAction.java
kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: I thought about it during the vacation.  I plan to use messaging + checkpoint polling, and remove rcf updates on index.  Lai's suggested "so when the count is above 128, the handler just returns it without doing anything extra".  Messaging + checkpoint polling is equally efficient with getting document from an index, if not faster.  And we save the trouble of maintaining another field on the state index.  Let me know if you guys think differently.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectorStateHandler.java
```diff
@@ -0,0 +1,163 @@
...
+     * Updates a detector's state according to GetStateHandler
+     * @param detectorId detector id
+     * @param handler specify how to convert from existing state object to an object we want to save
+     */
+    private void update(String detectorId, GetStateStrategy handler, Instant jobEnabledTime) {
+        try {
+            GetRequest getRequest = new GetRequest(this.indexName).id(detectorId);
+
+            clientUtil.<GetRequest, GetResponse>asyncRequest(getRequest, client::get, ActionListener.wrap(response -> {
+                DetectorInternalState newState = null;
```
wnbts, line 121: can error be part of transport state and trigger indexing when updated? so a get can be saved for each run.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -140,35 +149,61 @@ public void getAnomalyDetector(String adID, ActionListener<Optional<AnomalyDetec
         }, listener::onFailure);
     }
 
+    /**
+     * Get a detector's checkpoint and save a flag if we find any so that next time we don't need to do it again
+     * @param adID  the detector's ID
+     * @param listener listener to handle get request
+     */
+    public void getDetectorCheckpoint(String adID, ActionListener<Boolean> listener) {
```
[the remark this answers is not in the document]
wnbts: found it

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -203,242 +190,120 @@ private void prepareProfile(
...
-                listener.onResponse(profile);
-            } else {
-                logger
-                    .error(
-                        "Fail to find any anomaly result with anomaly score larger than 0 after AD job enabled time for detector {}",
-                        detectorId
-                    );
-                listener.onFailure(new RuntimeException("Fail to find detector state: " + detectorId, exception));
+            if (profilesToCollect.contains(ProfileName.INIT_PROGRESS)) {
+                listener.onResponse(new DetectorProfile.Builder().build());
```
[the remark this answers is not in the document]
kaituo: This means init progress is empty for a stopped detector.

src/main/resources/mappings/anomaly-state.json
```diff
@@ -0,0 +1,21 @@
...
+  "_meta": {
+    "schema_version": 1
+  },
+  "properties": {
+    "schema_version": {
+      "type": "integer"
+    },
+    "rcf_updates": {
+      "type": "integer"
+    },
```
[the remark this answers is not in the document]
kaituo: good catch.  Removed

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -1045,4 +1067,24 @@ private double computeRcfConfidence(RandomCutForest forest) {
...
+     * @param detectorId detector Id
+     * @param listener listener to return the result
+     */
+    public void getTotalUpdates(String modelId, String detectorId, ActionListener<Long> listener) {
+        ModelState<RandomCutForest> model = forests.get(modelId);
+        if (model != null) {
+            listener.onResponse(model.getModel().getTotalUpdates());
+        } else {
+            checkpointDao
+                .getModelCheckpoint(
```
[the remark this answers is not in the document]
kaituo: yeah, fortunately this is one-time thing and we don't have to do it repeatedly.  Let me know if you any way to improve this.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -203,242 +190,120 @@ private void prepareProfile(
...
                     ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser::getTokenLocation);
-                    AnomalyResult result = parser.namedObject(AnomalyResult.class, AnomalyResult.PARSE_FIELD_NAME, null);
-                    DetectorProfile profile = new DetectorProfile();
-                    if (result.getError() != null) {
-                        profile.setError(result.getError());
+                    DetectorInternalState detectorState = DetectorInternalState.parse(parser);
+                    long lastUpdateTimeMs = detectorState.getLastUpdateTime().toEpochMilli();
+
+                    // if state index hasn't been updated, we should not use the error field
+                    // For example, before enabled, error is stopped due to, we should not show
```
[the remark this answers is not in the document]
kaituo: Is the following easier to read?

                    // if state index hasn't been updated, we should not use the error field
                    // For example, before enabled, error is "stopped due to blah", we should not show
                    // this when the detector is enabled.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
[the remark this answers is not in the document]
wnbts: if the status of the model is completed, i would think the profile init status should also be completed to reflect the truth, whether the detector is running or stopped. otherwise, it might cause confusion that stopping a detector deletes the model or uncertainty about the model of a stopped detector.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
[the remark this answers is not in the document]
wnbts: ok. now i understand it. a more complete sentence will be clearer.

For example, before (the detector is) enabled, if the error message contains the phrase "stopped due to blah", ....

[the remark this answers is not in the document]
kaituo: will change.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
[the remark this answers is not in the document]
wnbts: i am worried about for every get profile call, the model gets pulled, loaded, and hosted.

some optimizations
1. there is not much to save from getting the checkpoint. but the checkpoint doesn't need to be fully deserialized, i.e., the bulk of the model, the trees are not needed. a lightweight deserialization only involving the forest containing the update count can be added to avoid recomputing trees and discarding them.
2. if the model is not hosted before deserialization, it doesn't need to be hosted since it's not being used by active jobs, saves memory space and write back to index again later during maintenance.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectorStateHandler.java
kaituo: On second thought, what you said make sense.  I'll keep the last error in transport state and access it in DetectorStateHandler.

wnbts: if after each run, the detector coordinator itself, holding the anomaly result and error, manages detector error state rather than letting the job scheduler do it, does it still require messaging?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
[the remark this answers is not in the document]
kaituo: The model does not need to be pulled, loaded, and hosted for every get profile call.  Only the first profile call if the model hasn't been hosted.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectorStateHandler.java
kaituo: You meant does it still require indexing into state index?

Yes, the state can be cleared due to maintenance, or coordinating node can change.  We need a permanent place to host the data.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -91,31 +92,31 @@ public ADStateManager(
      * @throws LimitExceededException when there is no sufficient resource available
      */
     public int getPartitionNumber(String adID, AnomalyDetector detector) {
-        Entry<Integer, Instant> partitonAndTime = partitionNumber.get(adID);
-        if (partitonAndTime != null) {
+        if (transportStates.containsKey(adID) && transportStates.get(adID).getPartitonNumber() != null) {
```
wnbts, line 95: minor. computeIfPresent is more concise

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -130,7 +131,13 @@ public void getAnomalyDetector(String adID, ActionListener<Optional<AnomalyDetec
             ) {
                 ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser::getTokenLocation);
                 AnomalyDetector detector = AnomalyDetector.parse(parser, response.getId());
-                currentDetectors.put(adID, new SimpleEntry<>(detector, clock.instant()));
+                TransportState state = transportStates.get(adID);
```
wnbts, line 134: minor. computeIfAbsent is more concise.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -204,4 +237,29 @@ public void resetBackpressureCounter(String nodeId) {
...
+     * @return last error for the detector
+     */
+    public String getLastError(String adID) {
+        if (transportStates.containsKey(adID) && transportStates.get(adID).getLastError() != null) {
+            Entry<String, Instant> errorAndTime = transportStates.get(adID).getLastError();
+            errorAndTime.setValue(clock.instant());
+            return errorAndTime.getKey();
+        }
+
+        return IMPOSSIBLE_ERROR;
```
wnbts, line 253: minor. this seems more no_error.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectorStateHandler.java
```diff
@@ -0,0 +1,165 @@
...
+            indexUtils,
+            clusterService
+        );
+        this.xContentRegistry = xContentRegistry;
+        this.adStateManager = adStateManager;
+    }
+
+    public void saveError(String error, String detectorId) {
+        // trigger indexing if no error recorded (e.g., this detector got enabled just now)
+        // or the recorded error is different than this one.
```
ylwu-amzn, line 114: Should we update error to record the latest error time if the error message is the same?

kaituo: We need last update time to know if the error is recorded before or after enabled time.  Current code path can handle that.  So we don't need to record the latest error time if the error message is the same.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -141,6 +141,7 @@ private AnomalyDetectorSettings() {}
     public static final String ANOMALY_DETECTORS_INDEX_MAPPING_FILE = "mappings/anomaly-detectors.json";
     public static final String ANOMALY_DETECTOR_JOBS_INDEX_MAPPING_FILE = "mappings/anomaly-detector-jobs.json";
     public static final String ANOMALY_RESULTS_INDEX_MAPPING_FILE = "mappings/anomaly-results.json";
+    public static final String ANOMALY_DETECTOR_STATE_INDEX_MAPPING_FILE = "mappings/anomaly-state.json";
```
ylwu-amzn, line 144: Minor, how about rename  the json file as "anomaly-detection-state.json"? In AD workbench, we will introduce detection task, plan to put task execution state in this index ".opendistro-anomaly-detection-state". 
BTW, how about we rename ANOMALY_DETECTOR_STATE_INDEX_MAPPING_FILE as ANOMALY_DETECTION_STATE_INDEX_MAPPING_FILE, similar to ANOMALY_DETECTOR_STATE_INDEX.

kaituo: changed both

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java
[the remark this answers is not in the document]
wnbts: 1. if job runner (sender) and state manager (responder) are on separate nodes for better scalability in the future, why let this change add an unnecessary coupling to prevent that?

2. the state manager is already a dependency of transport action, which generates errors first hand, why not letting the transport action directly update state manager, which updates the handler, (the chain of command is job runner > transport action > state manager > state handler), rather than taking a longer route involving the high-level job runner for more lower level details, which should only be responsible for sending job requests by schedule and have as little business logic as possible?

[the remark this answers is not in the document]
kaituo: 1. I don't see it is coming in the near feature.  State manager is saving the state of job runner.  It makes no sense to save it in different node.
2. detectorStateHandler and anomalyResultHandler are both dependency of job runner.  Previously, the logic of anomalyResultHandler is in the transport action, as you suggested.  Yaliang separated it out and moved it to job runner.  It is easier to read and being consistent by keeping the way as it is now.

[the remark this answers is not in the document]
ylwu-amzn: > 1. if job runner (sender) and state manager (responder) are on separate nodes for better scalability in the future, why let this change add an unnecessary coupling to prevent that?
> 2. the state manager is already a dependency of transport action, which generates errors first hand, why not letting the transport action directly update state manager, which updates the handler, (the chain of command is job runner > transport action > state manager > state handler), rather than taking a longer route involving the high-level job runner for more lower level details, which should only be responsible for sending job requests by schedule and have as little business logic as possible?

hi, Lai, good questions, before decoupling, we have no job runner and we handle all business logic and exceptions in transport action as it stands on top of other components. After decoupling, job runner is the top component, split the exception handling to job runner to avoid the exception handling logic scattered into multiple components(transport actions, job runners, etc).  So job runner will catch all exceptions including its own exceptions, then decide the following error handling such as stop detector if no enough resource, and write error in AD result index. Now Kaituo add `detectorStateHandler`, which will track the latest error. So it's same design logic to do error handling in one place.

Of course, we may have other options to do this, like create a new error handler, pass it to every component to handle exception and persist in AD result individually. Both options have pros and cons, need to do some tradeoff. How about we keep current design for this PR? We can discuss different options and solve the error handling in another PR if we have better option.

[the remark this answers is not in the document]
wnbts: this is not a blocker, as i have approved the pr.

it's good to use interface to decouple business logic from work execution dependencies. in another word, consider if the job scheduler code (a work execution dependency) change/migrate to some other dependency for execution, how much needs to re-developed, copy and pasted, re-tested? can/should the work be further reduced? in the ideal case, the interface should serve to protect the business logic.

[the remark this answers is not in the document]
ylwu-amzn: Thanks for the suggestion. Agree to "use interface to decouple business logic from work execution dependencies".  I think we can restructure when do AD workbench as it will change a lot of how to schedule detection task/job but not overdesign. Let's discuss more when do AD workbench.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java
```diff
@@ -119,6 +121,10 @@ public void setSettings(Settings settings) {
         this.maxRetryForEndRunException = AnomalyDetectorSettings.MAX_RETRY_FOR_END_RUN_EXCEPTION.get(settings);
     }
 
+    public void setDetectorStateHandler(DetectorStateHandler detectorStateHandler) {
```
ylwu-amzn, line 124: minor, should we rename `setDetectorStateHandler ` as `setDetectionStateHandler ` ? Same question for other `...detectorState...`

kaituo: renamed

PR #165, opened 2020-06-19
Run tests on docker if docker image exists

kaituo:
*Issue #, if available:*

*Description of changes:*

Previously, CI ran into issues when the docker image was not out yet after a version upgrade. This PR fixes it by using an environment variable to track whether the docker image exists or not.

This PR also removes test-workflow.yml since CI.yml can cover its functionality.

Testing done:
1. Verified in github docker tests skip if no docker image exists.


.github/workflows/CI.yml
```diff
@@ -48,15 +52,20 @@ jobs:
             echo "RUN /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch file:/tmp/opendistro-anomaly-detection-$plugin_version.zip" >> Dockerfile
 
             docker build -t odfe-ad:test .
+            echo "::set-env name=imagePresent::true"
+          else
+            echo "::set-env name=imagePresent::false"
```
yizheliu-amazon, line 57: if it is false, will the CI workflow fail?

kaituo: no, it won't.  See the workflow ran as part of the PR.

yizheliu-amazon: hmmm, should we fail workflow if docker image doesn't exist? It is a dependency to the workflow. If any PR comes when docker image is not there, we may miss the unit test as protection here.

kaituo: unit tests still runs.  Only docker related tests won't run.

yizheliu-amazon: I see. IT won't run if docker is missing. Will it be a concern if PR which may break IT gets merged, but not caught in CI given docker is missing?

kaituo: IT runs as well. Just won't run in a docker environment.

yizheliu-amazon: okay.

Issue #166, opened 2020-06-19
Use aggregation's default value if no docs


kaituo:
**Is your feature request related to a problem? Please describe.**
For sparse data (e.g., we only have docs half hour every day), currently preview/prediciton API cannot find data and prediction does not work.

This is related to the following line (verified with real data examples), which we filter out respond that has zero docs.

diff --git a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
index 140cb53..4b81a8b 100644
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
@@ -130,7 +130,7 @@ public class SearchFeatureDao {
return Optional.ofNullable(response)
- .filter(resp -> response.getHits().getTotalHits().value > 0L)
+ //.filter(resp -> response.getHits().getTotalHits().value > 0L)

**Describe the solution you'd like**
We should revisit whether we should do that. For example, when my aggregation is about the count of tts of a certain issue type in a stack, it is perfectly fine to use 0 if there are no docs. In fact, that's what Elasticsearch do. If there are no docs, a sum/count/cardinality aggregation in ES returns 0; a min/max/avg in ES returns null. Maybe we should follow what ES is doing.

PR #167, opened 2020-06-20
Use aggregation's default value if no docs

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/166

*Description of changes:*
Currently, for sparse data (e.g., we only have docs half an hour every day), the preview/prediction API cannot find aggregated value and thus doesn't work.

This PR mitigates the sparse data issue by using an aggregation's default value if any.

Testing done:
1. Without this PR, preview/prediction cannot find any features to use given sparse data for cardinality aggregation. With this PR, preview/prediction can.
2. Verified no unexpected exceptions are thrown with this PR during prediction or preview.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
@@ -173,10 +173,7 @@ public class SearchFeatureDao {
     }
 
     private Optional<double[]> parseResponse(SearchResponse response, List<String> featureIds) {
-        return parseAggregations(
-            Optional.ofNullable(response).filter(resp -> response.getHits().getTotalHits().value > 0L).map(resp -> resp.getAggregations()),
-            featureIds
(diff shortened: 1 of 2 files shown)


ylwu-amzn:
> Testing done:
> 
> 1. Without this PR, preview/prediction cannot find any features to use given sparse data for cardinality aggregation. With this PR, preview/prediction can.
> 2. Verified no unexpected exceptions are thrown with this PR during prediction or preview.
> 

For test1, `With this PR, preview/prediction can.`, I think it works for sum/count, but not works for max/min/avg. 
For test2, `no unexpected exceptions` have you tested all aggregation types we supported currently?

kaituo:
> For test1, `With this PR, preview/prediction can.`, I think it works for sum/count, but not works for max/min/avg.

that's correct.

> For test2, `no unexpected exceptions` have you tested all aggregation types we supported currently?
Yes, verified avg/min/max.  No unexpected exceptions when we are missing data.

PR #168, opened 2020-06-20
Rollover AD result index less frequently

kaituo:
*Issue #, if available:*

*Description of changes:*
Currently, we roll over the result index every 30 days or every 300000 docs. Assuming each doc has 1 KB and our result index has five shards, each shard takes about 60 MB, which is too small. Small shards are against ES best practice. This PR increases the rollover threshold to 9000000 docs, which increases the max shard size to roughly 1.8 GB. 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
@@ -86,7 +86,7 @@ public final class AnomalyDetectorSettings {
         .longSetting(
             "opendistro.anomaly_detection.ad_result_history_max_docs",
             // Suppose generally per cluster has 200 detectors and all run with 1 minute interval.
-            // We will get 288,000 AD result docs. So set it as 300k to avoid multiple roll overs
+            // We will get 288,000 AD result docs. So set it as 9000k to avoid multiple roll overs
             // per day.
             9_000_000L,
             0L,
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -88,7 +88,7 @@ private AnomalyDetectorSettings() {}
             // Suppose generally per cluster has 200 detectors and all run with 1 minute interval.
             // We will get 288,000 AD result docs. So set it as 300k to avoid multiple roll overs
             // per day.
```
wnbts, line 90: minor. comments are outdated.

kaituo: updated

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -86,9 +86,9 @@ private AnomalyDetectorSettings() {}
         .longSetting(
             "opendistro.anomaly_detection.ad_result_history_max_docs",
             // Suppose generally per cluster has 200 detectors and all run with 1 minute interval.
-            // We will get 288,000 AD result docs. So set it as 300k to avoid multiple roll overs
+            // We will get 288,000 AD result docs. So set it as 9000k to avoid multiple roll overs
             // per day.
-            300 * 1000L,
+            9_000_000L,
```
vamshin, line 91: Is this a problem for 200 detectors? From the comment 288.000 limit should be ok right?

kaituo: it is not a problem for 200 detectors.  It is a problem of small shard.   Please see my PR descriptions.

vamshin: Right i get that. Comment is misleading here. Why not mention the message in PR description as comment? We are talking about 200 detectors and roll overs in the comment. Atleast mention the PR?

Issue #159, opened 2020-06-17
Stopped detector: limit_exceeded_exception: AD memory circuit is broken.


ulasyuksel1979:
**Describe the bug**
We are using ELK Stack 7.7.0 and installed OpenDistro AD plugin ver. 1.8 on top of ELK. 
We have defined a simple detector job running in 1 minute period. It successfully goes into "running" state  and works for a while and then fails into "Initialization failure" status after a few minutes.   In elastic log, there is and entry "[2020-06-17T00:12:48,653][ERROR][c.a.o.a.t.AnomalyResultTransportAction] [ISMLAELSLX12] Received an error from node WyJji2luRSW-MckiY6tD-w while fetching anomaly grade for jqEQpXIBVGwASRJcv1us" for failure and after that error mesage job has been disbled with "[2020-06-17T00:12:48,751][INFO ][c.a.o.a.AnomalyDetectorJobRunner] [ISMLAELSLX12] AD Job was disabled by JobRunner for jqEQpXIBVGwASRJcv1us" message.

**Other plugins installed**
opendistro-anomaly-detection
opendistro-job-scheduler
opendistro_alerting
opendistro_index_management
opendistro_performance_analyzer
opendistro_security
opendistro_sql

**To Reproduce**

Go to AD page and under Detectors section, just run the defined job.

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/59475285/84861901-ca044e80-b07a-11ea-9a38-636c6ef7ed02.png)
![image](https://user-images.githubusercontent.com/59475285/84861935-ddafb500-b07a-11ea-8927-c0817bde53d0.png)
![image](https://user-images.githubusercontent.com/59475285/84861969-f029ee80-b07a-11ea-8646-21c7ed8c4cbc.png)
![image](https://user-images.githubusercontent.com/59475285/84862077-30896c80-b07b-11ea-92ac-3d579b488c26.png)

**Desktop (please complete the following information):**

**Additional context**
We have recently upgrade our system from ELK 7.6.1 with AD 1.7 to ELK 7.7.0 with AD 1.8. this jog was running successfully in previous version. Also, I've seen a similar bug specifying limited resources (1 processor and a small amount of memory) but our system has plenty amount of resources. We have 16 Elasticsearch node on virtual machines and each has 12 core with 32 GB of memory. Cpu consumption is around %30 and JVM memory consumption is around %60.

kaituo:
[...]

Issue #143, opened 2020-05-29
Add setting for enable/disable circuit breaker


kaituo:
**Is your feature request related to a problem? Please describe.**
Similar to https://github.com/opendistro-for-elasticsearch/odfe-learning-to-rank/pull/14#pullrequestreview-406253575

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #169, opened 2020-06-20
Add setting for enabling/disabling circuit breaker

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/143
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/159

*Description of changes:*
A circuit breaker is broken when heap memory usage exceeds 85%, and the related AD job would be disabled because of that. It is possible at one point the heap memory usage exceeds 85% and gets back to less than 85% soon afterward.

This PR mitigates the issue in the two following ways:
First, only disable the AD job after the circuit breaker is broken for a consecutive number of times (3 times).
Second, add a setting for enabling/disabling circuit breaker.

Testing done:
1. After disabling circuit breaker, an open circuit breaker does not affect AD job execution.
2. Verified an open circuit breaker wouldn't cause an AD job to be stopped immediately.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/LimitExceededException.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/LimitExceededException.java
@@ -31,7 +31,7 @@ public class LimitExceededException extends EndRunException {
     }
 
     /**
-     * Constructor with an anomaly detector ID and an explanation.
+     * Constructor with an anomaly detector ID and an explanation, and a flag for stopping.
      *
      * @param anomalyDetectorId ID of the anomaly detector for which the limit is exceeded
      * @param message explanation for the limit
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/LimitExceededException.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/LimitExceededException.java
@@ -29,4 +29,15 @@ public class LimitExceededException extends EndRunException {
     public LimitExceededException(String anomalyDetectorId, String message) {
         super(anomalyDetectorId, message, true);
     }
+
+    /**
+     * Constructor with an anomaly detector ID and an explanation.
+     *
+     * @param anomalyDetectorId ID of the anomaly detector for which the limit is exceeded
+     * @param message explanation for the limit
+     * @param stopNow whether to stop detector immediately
+     */
+    public LimitExceededException(String anomalyDetectorId, String message, boolean stopNow) {
+        super(anomalyDetectorId, message, stopNow);
+    }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
@@ -133,7 +133,7 @@ public final class AnomalyDetectorSettings {
     public static final Setting<Integer> MAX_RETRY_FOR_END_RUN_EXCEPTION = Setting
         .intSetting(
             "opendistro.anomaly_detection.max_retry_for_end_run_exception",
-            3,
+            6,
             0,
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
@@ -55,7 +55,7 @@ public class EnabledSetting {
             put(AD_PLUGIN_ENABLED, Setting.boolSetting(AD_PLUGIN_ENABLED, true, NodeScope, Dynamic));
 
             /**
-             * LTR breaker enable/disable setting
+             * AD breaker enable/disable setting
              */
             put(AD_BREAKER_ENABLED, Setting.boolSetting(AD_BREAKER_ENABLED, true, NodeScope, Dynamic));
         }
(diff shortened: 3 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
```diff
@@ -45,12 +45,19 @@
...
+
     private final Map<String, Setting<?>> settings = unmodifiableMap(new HashMap<String, Setting<?>>() {
         {
             /**
              * AD plugin enable/disable setting
              */
             put(AD_PLUGIN_ENABLED, Setting.boolSetting(AD_PLUGIN_ENABLED, true, NodeScope, Dynamic));
+
+            /**
+             * LTR breaker enable/disable setting
```
wnbts, line 58: issue. LTR seems to refer to AD.

ylwu-amzn:
How about we set a dynamic setting for the memory threshold? If user just use one cluster mainly for anomaly detection, they should be able to increase the memory for AD.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
kaituo: fixed

kaituo:
> one general comment. Three data points before disabling seems too quick. The system should be resilient and be able to recover on itself when conditions improve, or human operators would be burdened with every system resource spike.

How about 6?

kaituo:
> How about we set a dynamic setting for the memory threshold? If user just use one cluster mainly for anomaly detection, they should be able to increase the memory for AD.

This is beyond the scope of this PR.  Created an issue: https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/170

wnbts:
> > one general comment. Three data points before disabling seems too quick. The system should be resilient and be able to recover on itself when conditions improve, or human operators would be burdened with every system resource spike.
> 
> How about 6?

my opinion is that this number should be reasonably large for recovering expected system spikes and on the other side, not too large that the task is needlessly retrying in a permanently/consistently inoperable system. 3 data points/3 minutes seems too short for the former. I am not sure about 6 but it seems more reasonable.

kaituo:
> > > one general comment. Three data points before disabling seems too quick. The system should be resilient and be able to recover on itself when conditions improve, or human operators would be burdened with every system resource spike.
> > 
> > 
> > How about 6?
> 
> my opinion is that this number should be reasonably large for recovering expected system spikes and on the other side, not too large that the task is needlessly retrying in a permanently/consistently inoperable system. 3 data points/3 minutes seems too short for the former. I am not sure about 6 but it seems more reasonable.

suppose detector interval is 1 hour, then 6 means 6 hour retry before terminating models.  This value should be different among detectors, considering detector interval.  What do you think?

wnbts:
> hen 6 means 6 hour retry before terminating models. This value should be different among detectors, considering detector interval. What do you think?

that would be ideal. if that's not easy to do, at worst 6 hours means 6 attempts, not too big a problem.

kaituo:
> > hen 6 means 6 hour retry before terminating models. This value should be different among detectors, considering detector interval. What do you think?
> 
> that would be ideal. if that's not easy to do, at worst 6 hours means 6 attempts, not too big a problem.

I'll set it to 6 for now and created an issue https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/171 to track it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/LimitExceededException.java
```diff
@@ -29,4 +29,15 @@
     public LimitExceededException(String anomalyDetectorId, String message) {
         super(anomalyDetectorId, message, true);
     }
+
+    /**
+     * Constructor with an anomaly detector ID and an explanation.
```
wnbts, line 34: minor. and flag for stopping.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
```diff
@@ -100,6 +107,10 @@ public static boolean isADPluginEnabled() {
         return EnabledSetting.getInstance().getSettingValue(EnabledSetting.AD_PLUGIN_ENABLED);
     }
 
+    public static boolean isADBreakerEnabled() {
```
wnbts, line 110: Minor. Java doc for public interface is missing.

src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/LimitExceededException.java
kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -133,7 +133,7 @@ private AnomalyDetectorSettings() {}
     public static final Setting<Integer> MAX_RETRY_FOR_END_RUN_EXCEPTION = Setting
         .intSetting(
             "opendistro.anomaly_detection.max_retry_for_end_run_exception",
-            3,
+            6,
```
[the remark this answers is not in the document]
kaituo: added to the tracking issue.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/EnabledSetting.java
kaituo: added doc

Issue #37, opened 2020-01-29
Don't rollover empty AD result index


ylwu-amzn:
We roll over AD result by daily or if its records exceeds limitation. But if AD result index has no document, we should not roll over it.

PR #174, opened 2020-06-24
Add result indices retention period

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/37

*Description of changes:*
Currently we never delete the result index, even though customers have deleted the detector.   An increasing amount of result indices can use significant disk space, as well as memory pressure due to the creation of rolled over indices. This PR adds retention period to anomaly results.  We delete result indices when they are older than a retention period, which is 90 days by default. We use 90 days because that's the maximum days we allow users to view results on Kibana.  Users can configure the retention period via the setting opendistro.anomaly_detection.ad_result_history_retention_period dynamically.

Also, previously we roll over empty result indices.  This PR fixes that by removing the max age condition of result indices.  So we only roll over the result index when the maximum number of documents in the index is reached.

Testing done:
* manually tested result indices would be deleted after passing retention period.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
@@ -75,7 +75,7 @@ public class AnomalyDetectionIndices implements LocalNodeMasterListener {
     public static final String ALL_AD_RESULTS_INDEX_PATTERN = ".opendistro-anomaly-results*";
 
     // Elastic mapping type
-    private static final String MAPPING_TYPE = "_doc";
+    static final String MAPPING_TYPE = "_doc";
 
     private ClusterService clusterService;
     private final AdminClient adminClient;
@@ -274,7 +274,7 @@ public class AnomalyDetectionIndices implements LocalNodeMasterListener {
         }
     }
 
-    private void rolloverAndDeleteHistoryIndex() {
+    void rolloverAndDeleteHistoryIndex() {
         if (!doesAnomalyResultIndexExist()) {
             return;
         }
@@ -300,7 +300,7 @@ public class AnomalyDetectionIndices implements LocalNodeMasterListener {
         }, exception -> { logger.error("Fail to roll over result index", exception); }));
     }
 
-    private void deleteOldHistoryIndices() {
+    void deleteOldHistoryIndices() {
         Set<String> candidates = new HashSet<String>();
 
         ClusterStateRequest clusterStateRequest = new ClusterStateRequest()
@@ -341,9 +341,12 @@ public class AnomalyDetectionIndices implements LocalNodeMasterListener {
                             );
                         deleteIndexIteration(toDelete);
                     } else {
-                        logger.error("Succeeded in deleting expired anomaly result indices: {}.", Arrays.toString(toDelete));
+                        logger.info("Succeeded in deleting expired anomaly result indices: {}.", Arrays.toString(toDelete));
                     }
-                }, exception -> { deleteIndexIteration(toDelete); }));
+                }, exception -> {
+                    logger.error("Failed to delete expired anomaly result indices: {}.", Arrays.toString(toDelete));
+                    deleteIndexIteration(toDelete);
+                }));
             }
         }, exception -> { logger.error("Fail to delete result indices", exception); }));
     }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
@@ -15,38 +15,47 @@
 
 package com.amazon.opendistroforelasticsearch.ad.indices;
 
-import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.AD_RESULT_HISTORY_INDEX_MAX_AGE;
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.AD_RESULT_HISTORY_MAX_DOCS;
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.AD_RESULT_HISTORY_RETENTION_PERIOD;
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.AD_RESULT_HISTORY_ROLLOVER_PERIOD;
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.ANOMALY_DETECTORS_INDEX_MAPPING_FILE;
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.ANOMALY_DETECTOR_JOBS_INDEX_MAPPING_FILE;
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.ANOMALY_RESULTS_INDEX_MAPPING_FILE;
-import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.REQUEST_TIMEOUT;
 
 import java.io.IOException;
 import java.net.URL;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.admin.cluster.state.ClusterStateRequest;
 import org.elasticsearch.action.admin.indices.alias.Alias;
 import org.elasticsearch.action.admin.indices.create.CreateIndexRequest;
 import org.elasticsearch.action.admin.indices.create.CreateIndexResponse;
+import org.elasticsearch.action.admin.indices.delete.DeleteIndexRequest;
 import org.elasticsearch.action.admin.indices.rollover.RolloverRequest;
-import org.elasticsearch.action.admin.indices.rollover.RolloverResponse;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -275,15 +285,86 @@ private boolean rolloverHistoryIndex() {
...
-            logger.warn("{} not rolled over. Conditions were: {}", AD_RESULT_HISTORY_WRITE_INDEX_ALIAS, response.getConditionStatus());
+        adminClient.indices().rolloverIndex(request, ActionListener.wrap(response -> {
+            if (!response.isRolledOver()) {
+                logger.warn("{} not rolled over. Conditions were: {}", AD_RESULT_HISTORY_WRITE_INDEX_ALIAS, response.getConditionStatus());
+            } else {
+                logger.info("{} rolled over. Conditions were: {}", AD_RESULT_HISTORY_WRITE_INDEX_ALIAS, response.getConditionStatus());
+                deleteOldHistoryIndices();
+            }
+        }, exception -> {
+            logger.error("Fail to roll over result index");
```
ylwu-amzn, line 300: Why put into two lines? How about we use `logger.error("Fail to roll over result index", exception);` ? So we don't need to go to two lines when check log. There are maybe other request's log between these two lines, if check log manually, we need to skip other request's log.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -275,15 +285,86 @@ private boolean rolloverHistoryIndex() {
...
+                    if (latest < creationTime) {
+                        latest = creationTime;
+                        latestToDelete = indexName;
+                    }
+                }
+            }
+
+            if (candidates.size() > 1) {
+                // delete all indices except the last one because the last one may contain docs newer than the retention period
+                candidates.remove(latestToDelete);
```
ylwu-amzn, line 334: How about we just don't add `latestToDelete` to `candidates ` in the `for` loop above.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -275,15 +285,86 @@ private boolean rolloverHistoryIndex() {
...
+                                "Could not delete one or more Anomaly result indices: {}. Retrying one by one.",
+                                Arrays.toString(toDelete)
+                            );
+                        deleteIndexIteration(toDelete);
+                    } else {
+                        logger.error("Succeeded in deleting expired anomaly result indices: {}.", Arrays.toString(toDelete));
+                    }
+                }, exception -> { deleteIndexIteration(toDelete); }));
+            }
+        }, exception -> { logger.error("Fail to get creation dates of result indices"); }));
```
ylwu-amzn, line 350: This exception is for catching exception of `clusterStateRequest`? Not quite get why log `Fail to get creation dates` here.  The exception only caused by `get creation dates` ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -275,15 +285,86 @@ private boolean rolloverHistoryIndex() {
...
+            DeleteIndexRequest singleDeleteRequest = new DeleteIndexRequest(index);
+            adminClient.indices().delete(singleDeleteRequest, ActionListener.wrap(singleDeleteResponse -> {
+                if (!singleDeleteResponse.isAcknowledged()) {
+                    logger.error("Retrying deleting {} does not succeed.", index);
+                }
+            }, exception -> {
+                if (exception instanceof IndexNotFoundException) {
+                    logger.info("{} was already deleted.", index);
+                } else {
+                    logger.error("Retrying deleting {} does not succeed.", index);
```
ylwu-amzn, line 364: Same here, why put into two lines?

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -275,15 +285,86 @@ private boolean rolloverHistoryIndex() {
...
+                String[] toDelete = candidates.toArray(Strings.EMPTY_ARRAY);
+                DeleteIndexRequest deleteIndexRequest = new DeleteIndexRequest(toDelete);
+                adminClient.indices().delete(deleteIndexRequest, ActionListener.wrap(deleteIndexResponse -> {
+                    if (!deleteIndexResponse.isAcknowledged()) {
+                        logger
+                            .error(
+                                "Could not delete one or more Anomaly result indices: {}. Retrying one by one.",
+                                Arrays.toString(toDelete)
+                            );
+                        deleteIndexIteration(toDelete);
```
ylwu-amzn, line 344: Is it necessary to use two rounds of deletion? Roll over will be triggered periodically, so next run can re-delete the failed indices.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: How do we do that?  We don't know latestToDelete before looping through all indices.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: By default, we need to wait for 12 hours before re-deleting.  Adding a retry can mitigate some disk/memory issue without waiting for too long.  This is best effort and we cannot guarantee this would help.  We don't retry endlessly.  Just once.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: good catch.  Changed to more general error message.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: good catch.  Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: Fixed

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
ylwu-amzn: Can we iterate indices reversely and ignore the first index which meet this condition `(Instant.now().toEpochMilli() - creationTime) > historyRetentionPeriod.millis()` ?  

This is a minor suggestion. Ignore it if it's hard/impossible to do so.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
ylwu-amzn: Can we get which index failed to delete? Can we delete these failed indices in one call?

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -275,15 +286,82 @@ private boolean rolloverHistoryIndex() {
...
+                    } else {
+                        logger.error("Succeeded in deleting expired anomaly result indices: {}.", Arrays.toString(toDelete));
```
yizheliu-amazon, line 344: `logger.info`

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -275,15 +286,82 @@ private boolean rolloverHistoryIndex() {
...
+                    }
+                }, exception -> { deleteIndexIteration(toDelete); }));
```
yizheliu-amazon, line 346: maybe we can log the exception before retrying

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: The indices has no order. So we cannot do that.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: I don't know how to.  Delete response is of AcknowledgedResponse type that contains only isAcknowledged field.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: good catch. Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: added

kaituo:
> 2 minor comments. we may add UT or IT in the future.

added UT.

Issue #170, opened 2020-06-22
Setting dynamic threshold for memory setting


kaituo:
**Is your feature request related to a problem? Please describe.**
We have the following two memory threshold:
* a detector's model per node cannot exceed 10% of heap memory
* total model per node cannot exceed 10% of heap memory.

If user just use one cluster mainly for anomaly detection, they should be able to increase the memory for AD. 

**Describe the solution you'd like**
We can set a dynamic setting for the memory threshold.

PR #175, opened 2020-06-24
Add dyanmic setting for memory threshold

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/170

*Description of changes:*

We have the following two memory threshold:
* a detector's model size per node cannot exceed 10% of heap memory
* the total model size per node cannot exceed 10% of heap memory.

If users use the ES cluster mainly for anomaly detection, they should be able to increase the memory for AD.  This PR makes the threshold dynamically adjustable via the setting opendistro.anomaly_detection.model_max_size_percent.

Testing done:
* Manually verified I can dynamically adjust the memory threshold setting.  

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
@@ -44,7 +44,6 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.cluster.service.ClusterService;
-import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.monitor.jvm.JvmService;
 
 import com.amazon.opendistroforelasticsearch.ad.common.exception.LimitExceededException;
@@ -91,7 +90,6 @@ public class ModelManager {
 
     // configuration
     private final double modelDesiredSizePercentage;
-    // percentage of heap for the max size of a model
     private double modelMaxSizePercentage;
     private final int rcfNumTrees;
     private final int rcfNumSamplesInTree;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -163,7 +163,8 @@ private AnomalyDetectorSettings() {}
 
     public static final double DESIRED_MODEL_SIZE_PERCENTAGE = 0.0002;
 
-    public static final double MODEL_MAX_SIZE_PERCENTAGE = 0.1;
+    public static final Setting<Double> MODEL_MAX_SIZE_PERCENTAGE = Setting
+        .doubleSetting("opendistro.anomaly_detection.model_max_size_percent", 0.1, 0, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
wnbts, line 167: minor. why not add maxvalue?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -197,6 +200,9 @@ public ModelManager(
         this.forests = new ConcurrentHashMap<>();
         this.thresholds = new ConcurrentHashMap<>();
         this.shingleSize = shingleSize;
+
+        this.modelMaxSizePercentage = MODEL_MAX_SIZE_PERCENTAGE.get(settings);
```
wnbts, line 204: why not do this at plugin configuration to reduce unnecessary dependencies only used once?

kaituo: good idea.  Changed.

PR #176, opened 2020-06-25
Queries data from the index when insufficient data in buffer to form a full shingle

LiuJoyceC:
*Issue #, if available:*

*Description of changes:*
When running an AD on an interval, data points from past consecutive intervals are required in the shingling process. Currently, the AD only checks data points from a buffer which only contains data points for intervals which were previously run. This means for the first several intervals which the AD runs on, or if there were previously missed intervals (for example, due to restarts), the AD is unable to form a shingle and will output no results, even if the data points for the missed intervals are in the data index.

With this change, instead of issuing a query which only retrieves the current data point, a single query to the index will be issued for the current data point and any missing data points from the buffer that are needed to form a full shingle. The results of the query are cached so that if the data point is missing from the index itself, it will not be re-queried later.

Testing:
A comprehensive set of unit tests have been added (8 new tests covering 25 test cases) to test the FeatureManager getCurrentFeatures method under various initial states and query responses.


src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -118,40 +119,59 @@ public FeatureManager(
...
+                    Optional<double[]> point = points.get(i);
+                    if (point.isPresent()) {
+                        featuresMap.put(missingRanges.get(i).getValue(), point.get());
+                    }
+                }
+                updateUnprocessedFeatures(detector, shingle, featuresMap, endTime, listener);
+            }, listener::onFailure));
         } else {
             getProcessedFeatures(shingle, detector, endTime, listener);
         }
```
wnbts, line 145: is this branch still correct?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -118,67 +119,114 @@ public FeatureManager(
...
+     * Returns to listener unprocessed features and processed features (such as shingle) for the current data point.
+     *
+     * @param detector anomaly detector for which the features are returned
+     * @param startTime start time of the data point in epoch milliseconds
+     * @param endTime end time of the data point in epoch milliseconds
+     * @param cacheMissingDataPoints if set to true, missing data points are remembered and not re-queried later
+     * @param listener onResponse is called with unprocessed features and processed features for the current data point
+     * @throws IOException if a user gives wrong query input when defining a detector
+     */
+    public void getCurrentFeatures(
```
wnbts, line 139: this method should be private to the class.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -118,67 +119,114 @@ public FeatureManager(
...
 
+    private List<Entry<Long, Long>> getMissingRangesInShingle(
+        AnomalyDetector detector,
+        Map<Long, Optional<double[]>> featuresMap,
+        long endTime
+    ) {
+        long intervalMilli = getDetectorIntervalInMilliseconds(detector);
+
+        return getFullShingleEndTimes(endTime, intervalMilli)
+            .filter(time -> !featuresMap.containsKey(time))
```
wnbts, line 176: note the timestamps might not have exactly the same interval. there can be some jittering around each interval. for example, the timestamps from requests can be 59 sec, 1 min 58 sec, 2 min 03 sec, etc. if the data point already exists, the range can be skipped for query.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -118,67 +119,114 @@ public FeatureManager(
...
-                );
+    /**
+     * Returns to listener unprocessed features and processed features (such as shingle) for the current data point.
+     *
+     * @param detector anomaly detector for which the features are returned
+     * @param startTime start time of the data point in epoch milliseconds
+     * @param endTime end time of the data point in epoch milliseconds
+     * @param cacheMissingDataPoints if set to true, missing data points are remembered and not re-queried later
+     * @param listener onResponse is called with unprocessed features and processed features for the current data point
+     * @throws IOException if a user gives wrong query input when defining a detector
```
wnbts, line 137: minor. the query is from both the input and the system. if that happens, there will be no recovery for caller. usually those exceptions are caught and wrapped (abstracted) in unchecked exceptions for high-level callers, like in many other programming languages.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
LiuJoyceC: Ok, I will change the implementation to also find nearby timestamps rather than requiring an exact match on the timestamp. What's an appropriate threshold for how much the time the nearest data point can be off by before we consider a data point missing? For example, if the interval is 1 minute and the nearest data point to a particular timestamp is 30 seconds away, should the data point for the desired timestamp be counted as missing and therefore need to be queried?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
LiuJoyceC: The reason this is made public is to allow the unit tests to set the initial state of the shingle without caching missing data points during the setup (since the caching would affect the desired initial state of the shingle to be tested).

I can try to find alternative ways to set the initial state of the shingle, in which case I'll just remove the `cacheMissingDataPoints` parameter (and just always cache missing data points) and remove the method above.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
LiuJoyceC: Ok, I'll add a try catch block around `searchFeatureDao.getFeatureSamplesForPeriods()` and pass the error to `listener.onFailure()` instead, since that's where other failures from `searchFeatureDao.getFeatureSamplesForPeriods()` are being passed to.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
wnbts: the nearest data point within +- half interval from the expected can be used such that no data point is thrown away. if an existing data point happens to be right in the middle and there are no other data points nearby, then it may potentially be used for two neighboring ranges, which is unlikely but ok.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -118,67 +119,114 @@ public FeatureManager(
      * @param startTime start time of the data point in epoch milliseconds
      * @param endTime end time of the data point in epoch milliseconds
      * @param listener onResponse is called with unprocessed features and processed features for the current data point
+     * @throws IOException if a user gives wrong query input when defining a detector
      */
-    public void getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime, ActionListener<SinglePointFeatures> listener) {
+    public void getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime, ActionListener<SinglePointFeatures> listener)
+        throws IOException {
+        getCurrentFeatures(detector, startTime, endTime, true, listener);
```
kaituo, line 126: Why do we have another method with the same signature?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -118,67 +119,114 @@ public FeatureManager(
...
+        long endTime,
+        boolean cacheMissingDataPoints,
+        ActionListener<SinglePointFeatures> listener
+    ) throws IOException {
+
+        Deque<Entry<Long, Optional<double[]>>> shingle = detectorIdsToTimeShingles
+            .computeIfAbsent(detector.getDetectorId(), id -> new ArrayDeque<Entry<Long, Optional<double[]>>>(shingleSize));
+
+        Map<Long, Optional<double[]>> featuresMap = shingle.stream().collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+        List<Entry<Long, Long>> missingRanges = getMissingRangesInShingle(detector, featuresMap, endTime);
```
kaituo, line 151: What's the use case of always looking for missing data in the shingle besides starting detector running?  @wnbts

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: Please make sure AnomalyResultTransportAction.onFeatureResponse can handle the exception you throw.  The IOException sounds like sth you should rethrow EndRunException in AnomalyResultTransportAction and let user know.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
LiuJoyceC: The signature is slightly different. See above comments.

However, the most recent change addresses this, and there is no longer two different signatures for this method.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: We have to use the method before modification for high cardinality case.  Should we keep both versions?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -121,81 +122,133 @@ public FeatureManager(
...
-        double[][] currentPoints = filterAndFill(shingle, endTime, detector);
-        Optional<double[]> currentPoint = Optional.ofNullable(shingle.peekLast()).map(Entry::getValue);
-        listener
-            .onResponse(
-                Optional
-                    .ofNullable(currentPoints)
-                    .map(points -> new SinglePointFeatures(currentPoint, Optional.of(batchShingle(points, shingleSize)[0])))
-                    .orElse(new SinglePointFeatures(currentPoint, Optional.empty()))
-            );
+        if (shingle.isEmpty() || shingle.getLast().getKey() < endTime || !shingle.getLast().getValue().isPresent()) {
```
wnbts, line 187: minor. the second condition `shingle.getLast().getKey() < endTime` might need some relaxation. if the existing buffer is [1min, 2min,..., 8min] and endTime is 8min1sec, the last data point would still be valid for the endTime.

also, this if branch may be merged with the else branch, starting with optional current point and if current point is present, then optional current shingle.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -53,7 +54,7 @@
     private static final Logger logger = LogManager.getLogger(FeatureManager.class);
 
     // Each anomaly detector has a queue of data points with timestamps (in epoch milliseconds).
-    private final Map<String, ArrayDeque<Entry<Long, double[]>>> detectorIdsToTimeShingles;
+    private Map<String, ArrayDeque<Entry<Long, Optional<double[]>>>> detectorIdsToTimeShingles;
```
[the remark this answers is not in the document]
LiuJoyceC: Good catch. I forgot to add back the final when I was originally experimenting with mocking private variables for unit tests.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -184,19 +188,14 @@ private void getProcessedFeatures(
...
-                .onResponse(
-                    Optional
-                        .ofNullable(currentPoints)
-                        .map(points -> new SinglePointFeatures(currentPoint, Optional.of(batchShingle(points, shingleSize)[0])))
-                        .orElse(new SinglePointFeatures(currentPoint, Optional.empty()))
-                );
-        }
+        Optional<double[]> currentPoint = shingle.peekLast().getValue();
+        listener
+            .onResponse(
```
kaituo, line 193: The original code is easier to read.  Also, you do "currentPoint.map(point -> filterAndFill(shingle, endTime, detector))" without using currentPoint in filterAndFill, which looks strange.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -121,81 +122,132 @@ public FeatureManager(
...
     ) {
-        if (point.isPresent()) {
-            if (shingle.size() == shingleSize) {
-                shingle.remove();
-            }
-            shingle.add(new SimpleImmutableEntry<>(endTime, point.get()));
+        shingle.clear();
+        getFullShingleEndTimes(endTime, getDetectorIntervalInMilliseconds(detector))
+            .filter(time -> featuresMap.containsKey(time))
+            .mapToObj(time -> featuresMap.get(time))
```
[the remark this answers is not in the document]
LiuJoyceC: It depends on how we want to handle this case: https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java#L211-L214. This is the case that could cause some of the times to be missing from `featuresMap` at this line.

Is `aggs == null` in the search response equivalent to each queried time range having no data? In other words, do we want to treat `Collections.emptyList()` equivalent to `[Optional.empty(), Optional.empty(), ..., Optional.empty()]`? In the case of `Collections.emptyList()`, do we want to cache the null values for each of the queried time ranges so they will not get re-queried.

If yes, it might be cleaner to handle this logic in the callback passed to searchFeatureDao, so that the logic for determining shingle values based on the search response is located in one function rather than spread across multiple functions.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -121,81 +122,132 @@ public FeatureManager(
...
+            .forEach(e -> shingle.add(e));
+
+        if (featuresMap.containsKey(endTime)) {
             getProcessedFeatures(shingle, detector, endTime, listener);
         } else {
             listener.onResponse(new SinglePointFeatures(Optional.empty(), Optional.empty()));
```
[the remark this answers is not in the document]
LiuJoyceC: Are you suggesting to pass `featuresMap` as a parameter into `getProcessedFeatures` and check this condition in there?

The condition `featuresMap.containsKey(endTime)` is the simplest/most concise way to ensure the conditions `!shingle.isEmpty()` and `shingle.getLast().getKey() ≈ endTime` (where `≈` means within half an interval away as discussed, which basically means that the current point is present). The reason this check is being done in `updateUnprocessedFeatures` is because this check is only needed if `updateUnprocessedFeatures` is invoked, and not needed if `getProcessedFeatures` is directly invoked by `getCurrentFeatures` (in which case the `missingRanges.size() > 0` already ensures the above conditions, so checking `featuresMap` is unnecessary).

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
LiuJoyceC: This change was made to address the comments above from wnbts to simplify the code here to not use if else to ensure `currentPoint` is not an empty Optional. The way this is written is such that if `currentPoint` is an empty Optional, it triggers the `orElse` line, so that if the `currentPoint` is empty, we don't end up with a non-empty array of points in the `SinglePointFeatures` (if that were to happen, it would be an unexpected change in behavior from before).

The reason the original code didn't need to check that was because `currentPoint` wasn't an Optional before. Now that we are caching empty responses from the search request to ensure we don't re-query missing data, the points in the shingle are wrapped in Optionals.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
wnbts: no, featuremap is not needed. by the end of line 176, the collection shingle should contain exactly shingle sized data points, present or absent, the last being the most ending around endTime. getProcessedFeatures can output the same results for both if and else branches.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
wnbts: i haven't seen or know that happens in practice. if that ever happens, caching the results is ok. what i was suggesting is just simplify the implementation to one line like stream.maptoobject(time -> map.getOrDefault(time, <time, empty()>)).

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
LiuJoyceC: With the current implementation (where results are not added to the shingle if `searchFeatureDao` returns `Collections.emptyList()`), the shingle is not guaranteed to contain an entry around `endTime`, which is why this check is currently being done.

However, per your response on the comment below, if we go ahead and cache all missing ranges when `searchFeatureDao` returns `Collections.emptyList()`, then the shingle will be guaranteed to have the entry at or near `endTime`, and then checking the `featureMap` will no longer be needed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
LiuJoyceC: Alright, I will change the implementation to cache results for the missing ranges when the response is an empty list.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
LiuJoyceC: Another unexpected change in behavior that this is avoiding is running `filterAndFill` on a shingle where the current point is not present. This would result in the shingle filling in the missing current point with older data. In the original code, there were checks in `getCurrentFeatures` and `updateUnprocessedFeature` to ensure the current point is not missing. These checks were removed to address comments above about simplifying the logic.

LiuJoyceC: The code was modified to no longer have an unused param `point` in a function, and now `currentPoint.isPresent()` is explicitly checked to make the intention clear. Does this address the concern?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -132,81 +133,129 @@ public FeatureManager(
...
+                .toArray(double[][]::new);
+
             if (result.length < shingleSize) {
                 result = null;
             }
         }
         return result;
     }
 
+    private Stream<Entry<Long, Entry<Long, Optional<double[]>>>> getNearbyPointsForShingle(
```
kaituo, line 229: Have some documentation of this method since the return value is nested and not easy to understand?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -132,81 +133,129 @@ public FeatureManager(
      */
     public void getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime, ActionListener<SinglePointFeatures> listener) {
```
kaituo, line 134: Add comments on your workflow?  Is the following understanding correct?

First, you interpolate using nearest points.
Then, query for missing points.
Finally, do another round of interpolation using nearest points.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: yes, thanks for the change.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
LiuJoyceC: The code isn't actually interpolating before querying for missing points. It is meant to account for a small amount of jitter in time (see wnbts's comment above) so that points that are not actually missing aren't re-queried simply because the millisecond timestamp doesn't exactly match up. The fact that we have to account for random time jitter does make the implementation a bit more complicated than the original code (and my original first version of this feature).

For example, let's say the interval is 60,000 milliseconds, and the detector ends up running at time 60,000, time 120,001, and time 179,999 (simply due to random time jitter). At time 120,001, the function shouldn't re-query historical data just because there doesn't exist a data point stored in the shingle buffer whose timestamp matches up to exactly 60,001. It should just recognize that the point at timestamp 60,000 is the point from 1 interval ago. Similarly, at time 179,999, it shouldn't re-query the previous 2 intervals just because there aren't points whose timestamps are exactly 119,999 and 59,999.

Also, what actually gets stored in the shingle buffer (which persists to intervals in the future) is the actual original timestamp of the data point, so no interpolation is being done. In the example above, what gets stored in the shingle is:
`[<60000, x>, <120001, y>, <179999, z>]`. So the data point at time 60,000 is not getting interpolated to time 60,001. Instead, time 60,000 is just recognized as the timestamp of the previous interval.

The actual interpolation is done after the query for missing points. Here, missing points will be interpolated with points from neighboring intervals up to the configured `maxNeighborDistance` (which is currently configured to 2 intervals away). So if the shingle is missing an interval, such as `[<60000, x>, <179999, z>]`, then the data point from time 179,999 will get interpolated to time 119,999 as if that were the actual data point from the previous interval, so the resulting double[] from `filterAndFill` becomes `[x, z, z]`.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -132,81 +133,129 @@ public FeatureManager(
...
+                searchFeatureDao.getFeatureSamplesForPeriods(detector, missingRanges, ActionListener.wrap(points -> {
+                    for (int i = 0; i < points.size(); i++) {
+                        Optional<double[]> point = points.get(i);
+                        long rangeEndTime = missingRanges.get(i).getValue();
+                        featuresMap.put(rangeEndTime, new SimpleImmutableEntry<>(rangeEndTime, point));
+                    }
+                    updateUnprocessedFeatures(detector, shingle, featuresMap, endTime, listener);
+                }, listener::onFailure));
+            } catch (IOException e) {
+                listener.onFailure(new EndRunException(detector.getDetectorId(), CommonErrorMessages.INVALID_SEARCH_QUERY_MSG, e, true));
```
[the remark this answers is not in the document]
LiuJoyceC: Sure.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: First of all, there is no jitter.  So the logic is unnecessary. (FYI @wnbts)

You called getNearbyPointsForShingle before running query that actually use nearby points for interpolation, right? getNearbyPointsForShingle returns a map whose key is 60,000, 120,000, 180,000 in your example because the key's source is what getFullShingleEndTimes returns, right?

wnbts: @kaituo  i am not following this entire thread, just comment on the statement that "there is no jiftter". It looks like the time range comes from system clock https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/129d249b1f61d19413ce5b1b9491f9381e7b5c0b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorJobRunner.java#L132 , which has jitters. Also, scheduled job runs are not exactly evenly spaced for jitters are intentionally added to scheduler.

kaituo: @wnbts  scheduled job's jitter is disabled for AD.  It's possible that a scheduled job may not happen exactly as it is scheduled (e.g., system is under heavy load).  But in normal cases, the job should run on time.

@LiuJoyceC You can keep your code.  That should give some fault tolerance.

LiuJoyceC: > You called getNearbyPointsForShingle before running query that actually use nearby points for interpolation, right? getNearbyPointsForShingle returns a map whose key is 60,000, 120,000, 180,000 in your example because the key's source is what getFullShingleEndTimes returns, right?

The key in the map is meant to allow the shingle to match up data points to intervals. The actual timestamp of the data point is stored in the value of the map along with the data point value. In `updateUnprocessedFeatures`, it is the actual timestamp of the data point that is added to the shingle and persisted to future intervals. So the shingle in the example will contain the timestamps 60,000, 120,001, 179,999. The data points are only associated with their actual timestamps and not imputed to other timestamps at that stage.

> You can keep your code. That should give some fault tolerance.

👍

Issue #142, opened 2020-05-29
Use security plugin in CI


kaituo:
**Is your feature request related to a problem? Please describe.**
We don't have a way to enable security plugin while still doing CI.  As per my understanding it needs updates in ESRestTestCase class, cause they have provision to enable https calls but it needs to be set in the code.  

**Describe the solution you'd like**
We may have a subclass of ESRestTestCase and override some of their methods.  See https://github.com/opendistro-for-elasticsearch/sql/pull/473/files and https://github.com/opendistro-for-elasticsearch/alerting/pull/213

**Describe alternatives you've considered**

**Additional context**

PR #182, opened 2020-06-30
Support security plugin enabled testing in CI

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/142

*Description of changes:*
This change is to support integration testing against remote security-enabled clustering in our CI. Currently, we only support the integration testing without a security plugin and have to remove the security plugin in CI.

The critical change is to create an https client with basic auth in ODFERestTestCase and let our rest test case inherit from ODFERestTestCase. With default security configuration, the password for the admin user is also “admin.” For ssl certification, we choose to disable it since it’s the default demo cert used here.

Testing done:
* ran the scripts on CI workflow on Mac and tested it works.
* gradle build still passes.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -46,7 +46,6 @@ jobs:
 
             ## The ESRestTest Client uses http by default.
             ## Need to disable the security plugin to call the rest api over http.
-            echo "RUN if [ -d /usr/share/elasticsearch/plugins/opendistro_security ]; then /usr/share/elasticsearch/bin/elasticsearch-plugin remove opendistro_security; fi" >> Dockerfile
             echo "RUN if [ -d /usr/share/elasticsearch/plugins/opendistro-anomaly-detection ]; then /usr/share/elasticsearch/bin/elasticsearch-plugin remove opendistro-anomaly-detection; fi" >> Dockerfile
             echo "ADD anomaly-detection/build/distributions/opendistro-anomaly-detection-$plugin_version.zip /tmp/" >> Dockerfile
             echo "RUN /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch file:/tmp/opendistro-anomaly-detection-$plugin_version.zip" >> Dockerfile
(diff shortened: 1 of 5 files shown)


.github/workflows/CI.yml
```diff
@@ -63,9 +62,16 @@ jobs:
...
           docker run -p 9200:9200 -d -p 9600:9600 -e "discovery.type=single-node" odfe-ad:test
           sleep 90
-          curl -XGET http://localhost:9200/_cat/plugins
 
       - name: Run AD Test
         if: env.imagePresent == 'true'
         run: |
-          ./gradlew :integTestRunner --tests "com.amazon.opendistroforelasticsearch.ad.rest.*IT" --tests "com.amazon.opendistroforelasticsearch.ad.e2e.*IT" -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster"
+          security=`curl -XGET https://localhost:9200/_cat/plugins?v -u admin:admin --insecure |grep opendistro_security|wc -l`
+          if [ $security -gt 0 ]
```
weicongs-amazon, line 70: we can consider having both IT with security and without security running. This is the way we are using the ODFE building workflow. The team can decide which one is better

src/test/java/com/amazon/opendistroforelasticsearch/ad/ODFERestTestCase.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        builder.setStrictDeprecationMode(true);
+        return builder.build();
+    }
+
+    @SuppressWarnings("unchecked")
+    @After
+    protected void wipeAllODFEIndices() throws IOException {
+        Response response = client().performRequest(new Request("GET", "/_cat/indices?format=json"));
```
weicongs-amazon, line 86: not sure if AD plugin is using any hidden index or not. I got this issue for alerting plugin. https://github.com/opendistro-for-elasticsearch/alerting/pull/213/files#diff-e1e03fb0ddd2ab303d2631b957c116e1R64

kaituo: added

.github/workflows/CI.yml
kaituo: As discussed offline, gradle build already runs IT without security enabled.  Will keep CI as it is for now.

src/test/java/com/amazon/opendistroforelasticsearch/ad/ODFERestTestCase.java
```diff
@@ -0,0 +1,158 @@
...
+ * ODFE integration test base class to support both security disabled and enabled ODFE cluster.
+ */
+public abstract class ODFERestTestCase extends ESRestTestCase {
+
+    protected boolean isHttps() {
+        boolean isHttps = Optional.ofNullable(System.getProperty("https")).map("true"::equalsIgnoreCase).orElse(false);
+        if (isHttps) {
+            // currently only external cluster is supported for security enabled testing
+            if (!Optional.ofNullable(System.getProperty("tests.rest.cluster")).isPresent()) {
+                throw new RuntimeException("external cluster url should be provided for security enabled testing");
```
yizheliu-amazon, line 58: `external` might be confusing as in the CI we are actually using `localhost`. Maybe we can just say "cluster url should be provided for security enabled testing"

kaituo: make sense.  changed.

Issue #180, opened 2020-06-26
Make our initialization sampling more robust


kaituo:
**Is your feature request related to a problem? Please describe.**
Currently, we sample 24 points (with 60 points apart between two neighboring sample) in the history for cold start.  If there is any data hole, we stop moving forward.  For example, if the 3rd sample is missing, we can only end up with 61 data points (with interpolation and other optimization).  This causes long initialization.

**Describe the solution you'd like**
We can still move forward if there is a data hole and only interpolate points between present samples.  Continuous shingles will be used for cold start.

wnbts:
sampling-based data collection suffers from missing data. To eliminate this issue, the robust solution is collecting all points by scanning.

kaituo:
Why do we have to go to the other extreme to collect all samples? Even with all of the samples, we may suffer missing data.  Also, what do you think of my solution?

wnbts:
The goal should be making data collection robust enough that it won't become a recurring problem in the future. The proposed solution is towards that direction but is still vulnerable to missing data points and will be a recurring problem again.

wnbts:
regarding the extreme comment, this is not an extreme use for a. the workload is limited with specified number of samples, which is far less than that used in visualization/exploration, b. the workload occurs only once per model, c. on sparse streams, the process results into no aggregation work, and d. it is also used elsewhere in es.

kaituo:
> regarding the extreme comment, this is not an extreme use for a. the workload is limited with specified number of samples, which is far less than that used in visualization/exploration, b. the workload occurs only once per model, c. on sparse streams, the process results into no aggregation work, and d. it is also used elsewhere in es.

a. how many samples are you gonna collect?  1440 samples?
b. agreed.
c. you still need to deal with sparse data when sampling all data points, like more than 25% of samples are missing.
d. agreed.

I meant we don't need to self-negate so much.  Previous solution is fast, cheap, and relatively accurate.  Chris and you have spent so much time to get current thing to work.  My proposal adds minor improvement on top of yours.

wnbts:
thanks, i believe an exhaustive search with a reasonable limited range is the right solution to this issue or any potential issues of this nature in the future. The initial assumption that data stream is dense and continuous that data points can be easily/correctly interpolated has proven to be false and the data from real world is increasingly showing the opposite. Continuing with the bias does not just slow down data collection, worsening user experience, but also brings new issues to the quality of the data and the model.

[...]

PR #184, opened 2020-07-02
change to exhausive search for training data

wnbts:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/180

*Description of changes:* This pr uses a robust exhaustive search for collecting training data samples at model creation time so the process is the best-effort against missing data in indices. For more discussion, use the issue above.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -27,6 +27,7 @@ import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -41,6 +42,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.elasticsearch.action.ActionListener;
 
+import com.amazon.opendistroforelasticsearch.ad.common.exception.EndRunException;
+import com.amazon.opendistroforelasticsearch.ad.constant.CommonErrorMessages;
 import com.amazon.opendistroforelasticsearch.ad.dataprocessor.Interpolator;
 import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
 import com.amazon.opendistroforelasticsearch.ad.model.IntervalTimeConfiguration;
@@ -61,6 +64,8 @@ public class FeatureManager {
 
     private final int maxTrainSamples;
     private final int maxSampleStride;
+    private final int trainSampleTimeRangeInHours;
+    private final int minTrainSamples;
     private final int shingleSize;
     private final int maxMissingPoints;
     private final int maxNeighborDistance;
@@ -76,6 +81,8 @@ public class FeatureManager {
      * @param clock clock for system time
      * @param maxTrainSamples max number of samples from search
      * @param maxSampleStride max stride between uninterpolated train samples
+     * @param trainSampleTimeRangeInHours time range in hours for collect train samples
+     * @param minTrainSamples min number of train samples
      * @param shingleSize size of feature shingles
      * @param maxMissingPoints max number of missing points allowed to generate a shingle
      * @param maxNeighborDistance max distance (number of intervals) between a missing point and a replacement neighbor
@@ -89,6 +96,8 @@ public class FeatureManager {
         Clock clock,
         int maxTrainSamples,
         int maxSampleStride,
+        int trainSampleTimeRangeInHours,
+        int minTrainSamples,
         int shingleSize,
         int maxMissingPoints,
         int maxNeighborDistance,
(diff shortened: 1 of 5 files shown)


kaituo:
Please add labels to your PR so that we can generate release notes automatically.  See https://github.com/opendistro-for-elasticsearch/anomaly-detection/commit/c0622630708506e40758d1073b56056d6fbd1656

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -241,30 +250,59 @@ public void getColdStartData(AnomalyDetector detector, ActionListener<Optional<d
...
+            previous = sample;
+        }
+        batchShingle(currentBlock, this.shingleSize).ifPresent(batch -> Arrays.stream(batch).forEach(shingles::add));
+        listener.onResponse(Optional.of(shingles.toArray(new double[0][0])).filter(results -> results.length > 0));
+    }
+
+    private Optional<double[][]> batchShingle(List<double[]> points, int shingleSize) {
+        return Optional
+            .ofNullable(points)
+            .filter(p -> p.size() >= shingleSize)
```
kaituo, line 291: add some interpolation if you miss less than 25% of points?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -241,30 +250,59 @@ public void getColdStartData(AnomalyDetector detector, ActionListener<Optional<d
...
-                    ActionListener.wrap(samples -> processColdStartSamples(samples, listener), listener::onFailure)
-                );
+            List<Entry<Long, Long>> sampleRanges = getColdStartSampleRanges(detector, latest.get());
+            try {
+                searchFeatureDao
+                    .getFeatureSamplesForPeriods(
+                        detector,
+                        sampleRanges,
+                        ActionListener.wrap(samples -> processColdStartSamples(samples, listener), listener::onFailure)
+                    );
```
LiuJoyceC, line 260: Is checking for `latest.isPresent()` (or even passing in `latest` as a param) still needed, since `latest` is not used by `searchFeatureDao.getFeatureSamplesForPeriods` or anywhere else in this function anymore?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: Interpolation is for robust results.  I don't agree with truth should not be manipulated assertion.  Say samples are always 6 points, you are gonna drop them all and start with 0 shingles.

wnbts: yes, the behavior in the example is expected. when data is sparse and shingle is required, the conflict will persist throughout the detector run. even training data were to be populated by introducing non-existent data, real-time prediciton will still fail for lack of data. there will need to be a different solution for solving this problem. let's start with the right solution and see the feedback.

kaituo: We had this conversation before: once in your initial cold start CR where you dropped all samples if you encounter a hole, once for missing data inside the shingle.  We agree to use limited interpolation and not dropping samples.  What's the difference this time?

wnbts: Before this change, due to the few number of samples (a few dozens), interpolation has to be used or there is no training data at all.

With exhaustive search, there are more samples for training and knowingly inserting data that is known not to exist becomes unnecessary given its downsides.
1. interpolated data affects model accuracy.
2. it's highly controversial and idiosyncratic when to/not to fake data, how, and even why and therefore there is unlikely a solution satisfactory to all. Think person A thinks 20% is ok for interpolation. person B thinks 60% is better. person C thinks a different interpolation is the best. and yet person D says even 10% is bad. Using data as-it-is is the most agreeable approach.

This pr is adding improvements from standard data processing over the existing one, serving as the least controversial foundation yet not a one-way door. In case someone has a strong personal preference, they can always raise a new pr based on this one for everyone to see and take/reject.

kaituo: About your downside points:

1. no doubt
2. Interpolation is basic in machine learning.  I am not sure about the high controversial part.  We don't want to put users on such delicate choice of the percentage of fake data.  We want to guide users with reasonable choice.

I feel the PR can be better by considering data imputation.  I don't want a half-done solution.  We have already gone this far by running much more queries than we planned.  It's better to make it worthwhile and not trivially throw away points.

wnbts: the controversies are as follows.
1. when to interpolate/not to interpolate. with 1% data, can anyone "interpolate" the rest 99%? 10% to 90%? what's the boundary? who makes the rules on what authority? even with interpolation, how many more data points are expected to be produced? no one can guarantee those arbitrary rules will produce more data points. if there is still no enough samples, then what? keep beating this dead horse?
2. how to interpolate/impute. there are many options. for example, why not just filter out gaps and shingle the remaining? again, any proposal will be unfair to some and attract challenges and who is the judge.

it should be clear now that on this topic, the matters are not what everyone, or even we, can agree on, and they may continue forever due to their flexible/ill-defined targets and lack of a legal means/authority to resolve. What in one's opinion is completely-done can be considered further wrong in another's.

PR #203, opened 2020-08-06
Allows window size to be set per detector.

LiuJoyceC:
*Description of changes:*

- This change adds a new `windowSize` instance variable to the AnomalyDetector class along with a public getter for this variable, which the model uses to determine the shingle size. 
- The Create Detector and Update Detector APIs allow `window_size` to be set per detector. If no `window_size` param is provided, the default is set to 8.
- The Get Detector, Preview Detector, and Search Detector APIs will also reflect the `window_size` value in the API response.
- The FeatureManager and ModelManager classes have `shingleSize` removed as instance variables, and instead gets the shingle size from the anomaly detector that is passed to their methods.
- The `maxMissingPoints` (that determines how many data points can be missing and still form a valid shingle) is set to be at most 25% of the shingle size. The floor is still set to the same as before: `max(2, shingleSize)`. The `maxNeighborDistance` (max number of intervals away that a data point value can be imputed to a missing data point) is the same as before: `max(2, shingleSize)`.
- 6 new unit tests

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -67,6 +67,8 @@ public class FeatureManager {
     private final int maxSampleStride;
     private final int trainSampleTimeRangeInHours;
     private final int minTrainSamples;
+    private final double maxMissingPointsRate;
+    private final int maxNeighborDistance;
     private final double previewSampleRate;
     private final int maxPreviewSamples;
     private final Duration featureBufferTtl;
@@ -81,6 +83,8 @@ public class FeatureManager {
      * @param maxSampleStride max stride between uninterpolated train samples
      * @param trainSampleTimeRangeInHours time range in hours for collect train samples
      * @param minTrainSamples min number of train samples
+     * @param maxMissingPointsRate max proportion of shingle with missing points allowed to generate a shingle
+     * @param maxNeighborDistance max distance (number of intervals) between a missing point and a replacement neighbor
      * @param previewSampleRate number of samples to number of all the data points in the preview time range
      * @param maxPreviewSamples max number of samples from search for preview features
      * @param featureBufferTtl time to live for stale feature buffers
@@ -93,6 +97,8 @@ public class FeatureManager {
         int maxSampleStride,
         int trainSampleTimeRangeInHours,
         int minTrainSamples,
+        double maxMissingPointsRate,
+        int maxNeighborDistance,
         double previewSampleRate,
         int maxPreviewSamples,
         Duration featureBufferTtl
@@ -104,6 +110,8 @@ public class FeatureManager {
         this.maxSampleStride = maxSampleStride;
         this.trainSampleTimeRangeInHours = trainSampleTimeRangeInHours;
         this.minTrainSamples = minTrainSamples;
+        this.maxMissingPointsRate = maxMissingPointsRate;
+        this.maxNeighborDistance = maxNeighborDistance;
         this.previewSampleRate = previewSampleRate;
         this.maxPreviewSamples = maxPreviewSamples;
         this.featureBufferTtl = featureBufferTtl;
@@ -128,7 +136,7 @@ public class FeatureManager {
      */
     public void getCurrentFeatures(AnomalyDetector detector, long startTime, long endTime, ActionListener<SinglePointFeatures> listener) {
 
-        int shingleSize = detector.getWindowSize();
+        int shingleSize = detector.getShingleSize();
         Deque<Entry<Long, Optional<double[]>>> shingle = detectorIdsToTimeShingles
             .computeIfAbsent(detector.getDetectorId(), id -> new ArrayDeque<>(shingleSize));
 
@@ -163,7 +171,7 @@ public class FeatureManager {
         long endTime
     ) {
         long intervalMilli = getDetectorIntervalInMilliseconds(detector);
-        int shingleSize = detector.getWindowSize();
+        int shingleSize = detector.getShingleSize();
         return getFullShingleEndTimes(endTime, intervalMilli, shingleSize)
             .filter(time -> !featuresMap.containsKey(time))
             .mapToObj(time -> new SimpleImmutableEntry<>(time - intervalMilli, time))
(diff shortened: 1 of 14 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -285,6 +295,9 @@ public static AnomalyDetector parse(
                 case WINDOW_DELAY_FIELD:
                     windowDelay = TimeConfiguration.parse(parser);
                     break;
+                case WINDOW_SIZE_FIELD:
+                    windowSize = parser.intValue();
```
wnbts, line 299: minor. if the value is from input, a validation can be used such as non-positive values.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -597,6 +594,20 @@ private long truncateToMinute(long epochMillis) {
         return Instant.ofEpochMilli(epochMillis).truncatedTo(ChronoUnit.MINUTES).toEpochMilli();
     }
 
+    /**
+     * @return max number of missing points allowed to generate a shingle
+     */
+    private int getMaxMissingPoints(int shingleSize) {
```
wnbts, line 600: issue. Limit the imputed to up to 25%. The current value of 2 is based on shingle size 8. When shingle size is 1 ~ 3,  0 is the max. 4 ~ 7, 1.

LiuJoyceC: Sure, I will simplify the function to just `Math.floor(shingleSize * 0.25)`.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -597,6 +594,20 @@ private long truncateToMinute(long epochMillis) {
...
+        return Math.max(Math.min(2, shingleSize), (int) Math.floor(shingleSize * 0.25));
+    }
+
+    /**
+     * @return max distance (number of intervals) between a missing point and a replacement neighbor
+     */
+    private int getMaxNeighborDistance(int shingleSize) {
+        return Math.min(2, shingleSize);
```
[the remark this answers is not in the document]
LiuJoyceC: Ok, I'll add it to the AnomalyDetectorSettings.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -199,7 +206,7 @@ public static AnomalyDetector parse(XContentParser parser, String detectorId) th
      * @throws IOException IOException if content can't be parsed correctly
      */
     public static AnomalyDetector parse(XContentParser parser, String detectorId, Long version) throws IOException {
-        return parse(parser, detectorId, version, null, null);
+        return parse(parser, detectorId, version, null, null, AnomalyDetectorSettings.DEFAULT_SHINGLE_SIZE);
```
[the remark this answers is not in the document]
LiuJoyceC: A unit test was failing when I passed `null` for the default shingle size here, but I can find another way to fix the test. I agree it makes sense to follow the convention of the other fields and pass a `null`.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
LiuJoyceC: Sure, I'll add a change to validate and throw an IllegalArgumentException (unless is there another more appropriate exception class for this?)

LiuJoyceC:
@wnbts 
> question. is it tested whether this change will break web ui?

I've tested via script calling the Create/Update/Get/Search/Preview/Run/Start/Stop Detector APIs and the Get Detector Profile API. These are all of the APIs that are affected by the new parameter (in either the request, response, or behavior). The new parameter is optional (and defaults to the current prod behavior of using shingle size 8) and no other parameters in the request nor response are impacted. So if there haven't been any changes in how the web UI interacts with the APIs, the behavior should be the same as before.

wnbts:
> @wnbts
> 
> > question. is it tested whether this change will break web ui?
> 
> I've tested via script calling the Create/Update/Get/Search/Preview/Run/Start/Stop Detector APIs and the Get Detector Profile API. These are all of the APIs that are affected by the new parameter (in either the request, response, or behavior). The new parameter is optional (and defaults to the current prod behavior of using shingle size 8) and no other parameters in the request nor response are impacted. So if there haven't been any changes in how the web UI interacts with the APIs, the behavior should be the same as before.

also, this change requires testing on backwards compatibility. The system should handle without failure detectors that are created prior to this pr.

LiuJoyceC:
> also, this change requires testing on backwards compatibility. The system should handle without failure detectors that are created prior to this pr.

Good point. I will add the default shingle value to the `getShingleSize()` method so detectors without a `shingleSize` parameter will return the default shingle size value.

src/main/resources/mappings/anomaly-detectors.json
```diff
@@ -93,6 +93,9 @@
         }
       }
     },
+    "shingle_size": {
+      "type": "integer"
```
kaituo, line 97: Should the type be short as we don't expect shingle size exceed 32k?

LiuJoyceC: I don't see "short" commonly used in ES mappings (I know it's available as an ES number mapping type, but haven't seen it used in practice much). It seems the common convention is to use "integer" type for integers where "long" is not needed, and the other integers in this same file seem to reflect that. It doesn't seem we would ever expect schema version, detection interval, nor window delay to exceed 32k, but we still map them as the "integer" type.

kaituo: How do you evaluate the number is not used much in practice?  How much schema have you seen?

schema version would increment as long as the detector is updated.  We cannot limit the detector update saying you have exceeded 32k.
Detector interval and window delay can exceed 32 k if I am running detector on an old data.

short is like any programming language data type.  It is there for saving space purpose.  Otherwise, we only need double and long, nothing else.

Just a suggestion.  Won't block your commit.

LiuJoyceC: Ok thanks, since it's not a blocker I will merge it. I am aware what the short data type is, my comment was merely about consistency and common convention. Since we also haven't evaluated maximum limits to the shingle size, there isn't concrete evidence to support that the integer type is not appropriate for this field.

PR #207, opened 2020-08-07
Get detector

VijayanB:
Cat command to concatenate and print detectors
Added get method to get detector configuration

### DIFF
--- a/cli/cmd/cat.go
+++ b/cli/cmd/cat.go
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	entity "esad/internal/entity/ad"
+	"esad/internal/handler/ad"
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+const (
+	commandCat = "cat"
+)
+
+//catCmd prints detectors configuration based on id and name pattern.
+var catCmd = &cobra.Command{
+	Use:   commandCat + " [flags] [list of detectors]",
+	Short: "Concatenate and print detectors based on id or name pattern",
+	Long:  `concatenate and print detectors based on pattern, use "" to make sure the name is not matched with pwd lists'`,
+	Run: func(cmd *cobra.Command, args []string) {
+		//If no args, display usage
+		if len(args) < 1 {
+			if err := cmd.Usage(); err != nil {
+				fmt.Println(err)
+			}
+			return
+		}
+		idStatus, _ := cmd.Flags().GetBool("id")
+		commandHandler, err := getCommandHandler()
+		if err != nil {
+			fmt.Println(err)
+		}
+		// default is name
+		action := ad.GetAnomalyDetectorsByNamePattern
+		if idStatus {
+			action = getDetectorsByID
+		}
+		results, err := getDetectors(commandHandler, args, action)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+		printDetectors(results)
+	},
+}
+
+func getDetectors(
+	commandHandler *ad.Handler, args []string, get func(*ad.Handler, string) (
+		[]*entity.DetectorOutput, error)) ([]*entity.DetectorOutput, error) {
+	var results []*entity.DetectorOutput
+	for _, detector := range args {
+		output, err := get(commandHandler, detector)
+		if err != nil {
+			return nil, err
+		}
+		results = append(results, output...)
+	}
+	return results, nil
+}
+
+//getDetectorsByID gets detector output based on ID as argument
+func getDetectorsByID(commandHandler *ad.Handler, ID string) ([]*entity.DetectorOutput, error) {
+
+	output, err := ad.GetAnomalyDetectorByID(commandHandler, ID)
+	if err != nil {
+		return nil, err
+	}
+	return []*entity.DetectorOutput{output}, nil
+}
+
+//printDetectors displays the list of output. Since this is json format, use indent function to
+// pretty print before printing on console
+func printDetectors(results []*entity.DetectorOutput) {
+	if results == nil {
+		return
+	}
+	for _, d := range results {
+		formattedOutput, err := json.MarshalIndent(d, "", "  ")
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+		fmt.Println(string(formattedOutput))
+	}
+}
+
+func init() {
+	esadCmd.AddCommand(catCmd)
+	catCmd.Flags().BoolP("name", "", true, "Input is name or pattern")
+	catCmd.Flags().BoolP("id", "", false, "Input is id")
+}
--- a/cli/internal/controller/ad/ad.go
+++ b/cli/internal/controller/ad/ad.go
@@ -35,12 +35,14 @@ type Controller interface {
 	StartDetector(context.Context, string) error
 	StopDetector(context.Context, string) error
 	DeleteDetector(context.Context, string, bool, bool) error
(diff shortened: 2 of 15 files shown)


cli/internal/gateway/ad/ad.go
```diff
@@ -59,6 +61,56 @@ func (g *gateway) buildCreateURL() (*url.URL, error) {
 	return endpoint, nil
 }
 
+// CreateDetector Creates an anomaly detector job.
```
kaituo, line 64: remove?

cli/cmd/cat.go
```diff
@@ -0,0 +1,105 @@
...
+			results = executeByID(commandHandler, args)
+		} else {
+			results = executeByNames(commandHandler, args)
+		}
+		printDetectors(results)
+	},
+}
+
+//executeByNames gets detector output based on name as argument
+func executeByNames(commandHandler *ad.Handler, args []string) []*entity.DetectorOutput {
```
kaituo, line 56: The code looks same as execyteByID.  Am I missing anything?

cli/internal/entity/ad/ad.go
```diff
@@ -143,3 +143,32 @@ type Container struct {
...
+
+//DetectorResponse represents detector's setting
+type DetectorResponse struct {
+	ID              string          `json:"_id"`
+	AnomalyDetector AnomalyDetector `json:"anomaly_detector"`
+}
+
+//DetectorOutput represents detector's setting displayed to user
+type DetectorOutput struct {
+	ID            string
```
kaituo, line 163: If we add a field in the get detector API on the plugin side, would it break anything here?

cli/internal/gateway/ad/ad.go
VijayanB: Do you mean to remove this line? In go, if a method is exported, it is practice to write comment else linter will complain. At line 114, i  have a method CreateDetector, hence the comment.

cli/internal/entity/ad/ad.go
VijayanB: It won't break, it will not be deserialized.

cli/cmd/cat.go
VijayanB: yes.
execute by name calls method ad.GetAnomalyDetectorsByNamePattern which returns []*entity.DetectorOutput, while
executeby id calls method ad.GetAnomalyDetectorsByID which returns *entity.DetectorOutput. Thats way i am not able to use pass function as parameter. In go, type has to be matched.

cli/internal/gateway/ad/ad.go
kaituo: my bad.  I thought the following is a commented out code block.  Does Go has block comment?

cli/cmd/cat.go
kaituo: Does it make sense to pass a boolean or enum say "by field".  And then you call different functions based on this by field?

cli/internal/gateway/ad/ad.go
VijayanB: Yeah. I updated with block comments.

cli/cmd/cat.go
VijayanB: Sure. I wrapped the method for executeByID and used flags to differentiate action .

cli/cmd/cat.go
```diff
@@ -0,0 +1,99 @@
...
+		if idStatus {
+			action = executeByID
+		}
+		if results := getDetectors(commandHandler, args, action); results != nil {
+			printDetectors(results)
+		}
+	},
+}
+
+func getDetectors(commandHandler *ad.Handler, args []string, get func(*ad.Handler, string) ([]*entity.DetectorOutput, error)) []*entity.DetectorOutput {
```
ylwu-amzn, line 56: This line seems very long. How about breaking it into two lines?

cli/cmd/cat.go
```diff
@@ -0,0 +1,99 @@
...
+			fmt.Println(err)
+			return nil
+		}
+		results = append(results, output...)
+	}
+	return results
+}
+
+//executeByID gets detector output based on ID as argument
+func executeByID(commandHandler *ad.Handler, ID string) ([]*entity.DetectorOutput, error) {
```
ylwu-amzn, line 70: This method is not to execute a detector by ID. How about rename it as "getDetectorByID"

cli/cmd/cat.go
```diff
@@ -0,0 +1,99 @@
...
+		commandHandler, err := getCommandHandler()
+		if err != nil {
+			fmt.Println(err)
+		}
+		// default is name
+		action := ad.GetAnomalyDetectorsByNamePattern
+		if idStatus {
+			action = executeByID
+		}
+		if results := getDetectors(commandHandler, args, action); results != nil {
```
ylwu-amzn, line 50: If `results` is `nil`, what message will user see?

cli/cmd/cat.go
```diff
@@ -0,0 +1,99 @@
...
+	return []*entity.DetectorOutput{output}, nil
+}
+
+//printDetectors displays the list of output. Since this is json format, use indent function to
+// pretty print before printing on console
+func printDetectors(results []*entity.DetectorOutput) {
+	if results == nil {
+		return
+	}
+	for _, d := range results {
```
ylwu-amzn, line 85: Is it best practice to use single-letter variable name in Go? How about replace `d` with `detector` ? That will be easier to read the code.

cli/cmd/cat.go
VijayanB: Ack. Changed

cli/cmd/cat.go
VijayanB: Ack. Changed

cli/cmd/cat.go
VijayanB: So only possibility getDetectors can get nil is if there is an error from GET method. In that case it will display the error message ( why get failed ). This was handled inside getDetectors. I refactored now to return err and print outside now.

cli/cmd/cat.go
VijayanB: https://github.com/golang/go/wiki/CodeReviewComments#variable-names
The convention is you only need more descriptive if it is used far away from declaration. Especially for iterators, single variable name is preferred.

cli/cmd/cat.go
```diff
@@ -0,0 +1,101 @@
...
+			fmt.Println(err)
+			return
+		}
+		fmt.Println(string(formattedOutput))
+	}
+}
+
+func init() {
+	esadCmd.AddCommand(catCmd)
+	catCmd.Flags().BoolP("name", "", true, "Input is name or pattern")
```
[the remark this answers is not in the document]
VijayanB: As discussed offline, i will fix this accordingly.

cli/cmd/cat.go
VijayanB: There is will be either json output or error message per detector.

cli/internal/controller/ad/ad.go
```diff
@@ -467,5 +487,35 @@ func (c controller) DeleteDetectorByName(ctx context.Context, name string, force
 		}
 	}
 	return nil
+}
 
+//GetDetectorsByName get detector based on name pattern. It first calls SearchDetectorByName and then
+// gets lists of detectorId and call GetDetector to get individual detector configuration
+func (c controller) GetDetectorsByName(ctx context.Context, pattern string, display bool) ([]*entity.DetectorOutput, error) {
+	matchedDetectors, err := c.getDetectorsToProcess(ctx, "fetch", pattern)
```
ylwu-amzn, line 495: What use cases have you tested for the name pattern?  `SearchDetectorByName` method is using `match` query. Why not use `regexp` ?

```
	payload := entity.SearchRequest{
		Query: entity.SearchQuery{
			Match: entity.Match{
				Name: name,
			},
		},
	}
```

VijayanB: I followed this from documentation:
https://opendistro.github.io/for-elasticsearch-docs/docs/ad/api/#search-detector

ylwu-amzn: Can you do some test for different regex use cases? Like "test.*", "test[0-9]" etc.

cli/cmd/cat.go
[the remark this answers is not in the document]
ylwu-amzn: Sure, will approve this PR. Please fix this in another PR.

cli/internal/controller/ad/ad.go
VijayanB: I will create an issue to update search detecter using regex and will submit new PR with test cases.

VijayanB: https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/213

PR #208, opened 2020-08-11
Change to use callbacks in cold start

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/78

*Description of changes:*

This PR changes the code path of cold start in the transport layer to use callbacks.

Previously, I created AD’s ExecutorService that has one thread for cold starts in ColdStartRunner. When triggering a cold start, we can submit a task in the ExecutorService, and consult a hash map (keyed by detector Id) that cached the results of recent cold start results. Since I have to invoke the cold start thread in various callbacks, ColdStartRunner does not work anymore.  I created a cold start thread pool and put the cold start result in the transport state instead.

This PR also handles new exceptions like invalid queries introduced by recent changes on ModelManager and FeatureManager.

This PR lowers the severity of a couple of log messages in HashRing and RCFPollingTransportAction to avoid overwhelming log file readers. These log messages are common.

This PR corrects typos and updates known causes of EndRunException in comments.

Testing done:
1. Simulated cold start failures: Exceptions of cold starts can be seen by the transport layer.  EndRunException can cause AD jobs to be terminated.
2. Happy case of a cold start still works.


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -142,6 +141,7 @@
     public static final String AD_BASE_URI = "/_opendistro/_anomaly_detection";
     public static final String AD_BASE_DETECTORS_URI = AD_BASE_URI + "/detectors";
     public static final String AD_THREAD_POOL_NAME = "ad-threadpool";
+    public static final String COLD_START_THREAD_POOL_NAME = "ad-cold-start-threadpool";
```
VijayanB, line 144: Shall we add this as last constant? ( constants arranged alphabetically )

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
```diff
@@ -88,8 +99,11 @@ public boolean expired(Duration stateTtl, Instant now) {
         if (checkpoint != null) {
             ans = ans && expired(stateTtl, now, checkpoint);
         }
-        if (lastError != null) {
-            ans = ans && expired(stateTtl, now, lastError.getValue());
+        if (lastDetectionError != null) {
```
VijayanB, line 102: Question: Is it possible to have both (lastDetectionError and lastColdStartException) to be not null ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -250,12 +250,37 @@ public String getLastError(String adID) {
...
+        TransportState state = transportStates.computeIfAbsent(adID, id -> new TransportState(id));
+        state.setLastColdStartException(new SimpleEntry<>(exception, clock.instant()));
+    }
+
+    /**
+     * Get last cold start exception of a detector
+     * @param adID detector id
+     * @return last cold start exception for the detector
+     */
+    public Exception fetchColdStartException(String adID) {
```
VijayanB, line 277: Question: Where is this used?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
kaituo: changed

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
kaituo: yes, it is possible.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
kaituo: It is used in AnomalyResultTransportAction.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -770,39 +775,122 @@ private boolean shouldStart(
...
+                    .wrap(res -> { LOG.info("Succeeded in training {}", detectorId); }, exception -> {
+                        if (exception instanceof AnomalyDetectionException) {
+                            // partitioned model exceeds memory limit
+                            stateManager.setLastColdStartException(detectorId, exception);
+                        } else if (exception instanceof IllegalArgumentException) {
+                            // IllegalArgumentException due to invalid training data
+                            stateManager
+                                .setLastColdStartException(
+                                    detectorId,
+                                    new EndRunException(detector.getDetectorId(), "Invalid training data", exception, false)
```
VijayanB, line 795: since detectorId is already extracted, can we reuse it here and other places like 811, 802 ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
VijayanB: Got it. For some reason AnomalyResultTransportAction was not loaded .

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -397,32 +396,44 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
...
+     *         a. invalid search query
      *     2. LimitExceededException from one of RCF model node when the total size of the models
      *      is more than X% of heap memory.
      *     3. InternalFailure wrapping ElasticsearchTimeoutException inside caused by
      *      RCF/Threshold model node failing to get checkpoint to restore model before timeout.
      */
-    private boolean coldStartIfNoModel(AtomicReference<AnomalyDetectionException> failure, AnomalyDetector detector)
+    private Exception coldStartIfNoModel(AtomicReference<AnomalyDetectionException> failure, AnomalyDetector detector)
         throws AnomalyDetectionException {
         AnomalyDetectionException exp = failure.get();
```
VijayanB, line 417: Shall we do short circuit like below?
if ( exp == null )
    return null
and then...
if !(exp instanceof  ResourceNotFoundException)
   throw exp
String adID = detector.getDetectorID();
......

Usually it is hard to read if there is nested scope and else .
Also it increases Cyclomatic complexity .

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: Nice catch.  Changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -770,39 +776,115 @@ private boolean shouldStart(
...
+        String detectorId = detector.getDetectorId();
+
+        ActionListener<Optional<double[][]>> listener = ActionListener.wrap(trainingData -> {
+            if (trainingData.isPresent()) {
+                double[][] dataPoints = trainingData.get();
+
+                ActionListener<Void> trainModelListener = ActionListener
+                    .wrap(res -> { LOG.info("Succeeded in training {}", detectorId); }, exception -> {
+                        if (exception instanceof AnomalyDetectionException) {
+                            // partitioned model exceeds memory limit
```
wnbts, line 789: the two cases are different

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -400,14 +399,21 @@ private static Void initGson() {
...
+        return Arrays
+            .asList(
                 new FixedExecutorBuilder(
                     settings,
                     AD_THREAD_POOL_NAME,
                     Math.max(1, EsExecutors.allocatedProcessors(settings) / 4),
                     AnomalyDetectorSettings.AD_THEAD_POOL_QUEUE_SIZE,
                     "opendistro.ad." + AD_THREAD_POOL_NAME
+                ),
+                new FixedExecutorBuilder(
```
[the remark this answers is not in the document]
kaituo: removed

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -330,22 +336,38 @@ private LongStream getFullShingleEndTimes(long endTime, long intervalMilli) {
      *                 onFailure is called with EndRunException on feature query creation errors
      */
     public void getColdStartData(AnomalyDetector detector, ActionListener<Optional<double[][]>> listener) {
+        ActionListener<Optional<Long>> latestTimeListener = ActionListener
+            .wrap(latest -> getColdStartSamples(latest, detector, listener), listener::onFailure);
         searchFeatureDao
             .getLatestDataTime(
                 detector,
-                ActionListener.wrap(latest -> getColdStartSamples(latest, detector, listener), listener::onFailure)
+                new ThreadedActionListener<>(
```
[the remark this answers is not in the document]
kaituo: Ideally, all AD code should run inside AD threadpool.  This is long overdue.  I am starting with cold start code.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -250,12 +250,37 @@ public String getLastError(String adID) {
...
+    public void setLastDetectionError(String adID, String error) {
         TransportState state = transportStates.computeIfAbsent(adID, id -> new TransportState(id));
-        state.setLastError(new SimpleEntry<>(error, clock.instant()));
+        state.setLastDetectionError(new SimpleEntry<>(error, clock.instant()));
+    }
+
+    /**
+     * Set last cold start error of a detector
+     * @param adID detector id
+     * @param exception exception, can be null
```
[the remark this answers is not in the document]
kaituo: null means there is no error, so no exception.  I know you are passionate about optional.  Here it is unnecessary.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -250,12 +250,37 @@ public String getLastError(String adID) {
...
+
+    /**
+     * Get last cold start exception of a detector
+     * @param adID detector id
+     * @return last cold start exception for the detector
+     */
+    public Exception fetchColdStartException(String adID) {
+        if (transportStates.containsKey(adID) && transportStates.get(adID).getLastColdStartException() != null) {
+            Entry<Exception, Instant> errorAndTime = transportStates.get(adID).getLastColdStartException();
+            errorAndTime.setValue(clock.instant());
```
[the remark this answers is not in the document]
kaituo: It is the same as other transport state.  We are gonna clear these states if they are not accessed within one hour.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: LimitExceededException is a subtype of AnomalyDetectionException.  I am commenting on one example here.  Updated comment to make it more clear.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
wnbts: does it mean every callback needs to run this way? can this be simplified at a higher level?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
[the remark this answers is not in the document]
wnbts: if it's no error, would it suffice to leave the entire record to null? <exception, time> is null seems simpler.

optional question. why optional is **unnecessary** ? clearly this variable can be present or absent and the need seems to exist.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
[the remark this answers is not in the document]
wnbts: Would one last access time for one entire statue be sufficient? instead of keeping one timestamp for each field and still using the latest one

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
kaituo: yes. I don't know how to simplify this.  Any suggestion?

kaituo:
> code coverage seems to decrease

Yes.  Maybe because I deleted a class.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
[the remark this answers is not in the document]
kaituo: changed to optional

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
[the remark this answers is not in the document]
kaituo: changed

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -770,39 +776,109 @@ private boolean shouldStart(
...
-                    detector.getDetectorId(),
-                    "Time out while indexing cold start checkpoint or get training data",
-                    timeoutEx
-                );
-            } catch (EndRunException endRunEx) {
-                throw endRunEx;
-            } catch (Exception ex) {
-                throw new EndRunException(detector.getDetectorId(), "Error while cold start", ex, false);
+        if (previousException != null) {
+            LOG.error("Previous exception of {}: {}", () -> detectorId, () -> previousException);
```
ylwu-amzn, line 861: Why use `() -> detectorId`?  `detectorId` should work.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -770,39 +776,109 @@ private boolean shouldStart(
...
+            if (previousException instanceof EndRunException && ((EndRunException) previousException).isEndNow()) {
```
ylwu-amzn, line 862: If it's `EndRunException ` and  `isEndNow()` is true, JobRunner will stop detector job. Will the `previousException` be cleaned when stop detector job? Why need to check previous exception ? If user wait for some time and restart detector, the `previousException` may be gone.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
wnbts: TransportNodesAction seems to take thread pool name.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -770,39 +776,109 @@ private boolean shouldStart(
...
+                LOG.info("Trigger cold start for {}", detectorId);
+                coldStart(detector);
+            }
+        }, exception -> {
+            Throwable cause = ExceptionsHelper.unwrapCause(exception);
+            if (cause instanceof IndexNotFoundException) {
+                LOG.info("Trigger cold start for {}", detectorId);
+                coldStart(detector);
+            } else {
+                LOG.error(String.format("Fail to get checkpoint state for %s", detectorId), exception);
```
ylwu-amzn, line 878: Should we throw exception ? Why not record this exception as `previousException `?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
```diff
@@ -29,16 +29,19 @@
     private Entry<Integer, Instant> partitonNumber;
     // checkpoint fetch time
     private Instant checkpoint;
-    // last error. Used by DetectorStateHandler to check if the error for a
+    // last detection error. Used by DetectorStateHandler to check if the error for a
     // detector has changed or not. If changed, trigger indexing.
-    private Entry<String, Instant> lastError;
+    private Entry<String, Instant> lastDetectionError;
+    // last training error. Used to save cold error by a concurrent cold start thread.
```
ylwu-amzn, line 35: `cold error`  means `cold start error`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -397,32 +396,45 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
...
-                globalRunner.compute(new ColdStartJob(detector));
-                return true;
-            } else {
-                throw exp;
+        if (exp == null) {
+            return null;
+        }
+
+        if (!(exp instanceof ResourceNotFoundException)) {
+            throw exp;
```
ylwu-amzn, line 423: Why not return this exception ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
kaituo: You meant the threadpool name in TransportResponseHandler.  Our case is not just transport thread.  Also, the threadpool name is good for one callback.  It does not cover nested call backs.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: See https://www.baeldung.com/log4j-2-lazy-logging

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: Yes, it will be cleared.  We need to check previous exception because cold start is asynchronous.  We don't know cold start error instantly.

If transport state gets cleared up, next few runs would trigger the same error.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: The returned exception is for cold start related failure only.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: This is in callback.  We cannot throw exception directly as it is asynchronous and won't be caught by AD job.  What do we expect the job runner to do here?  For example, we can reach here if cold start does not happen before and the system is under extreme heavy load and get request gets rejected.  We can record the error and retry later.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
kaituo: Good catch.  Fixed

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: From the document you shared , 

> The lambda expression is only evaluated if the corresponding log level is enabled. This is referred to as lazy logging.

I think `error` log level is always enabled, so the lambda `() -> detectorId` will always be executed?

kaituo: you are right.  Changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -331,10 +338,12 @@ private LongStream getFullShingleEndTimes(long endTime, long intervalMilli, int
      *                 onFailure is called with EndRunException on feature query creation errors
      */
     public void getColdStartData(AnomalyDetector detector, ActionListener<Optional<double[][]>> listener) {
+        ActionListener<Optional<Long>> latestTimeListener = ActionListener
+            .wrap(latest -> getColdStartSamples(latest, detector, listener), listener::onFailure);
         searchFeatureDao
             .getLatestDataTime(
                 detector,
-                ActionListener.wrap(latest -> getColdStartSamples(latest, detector, listener), listener::onFailure)
+                new ThreadedActionListener<>(logger, threadPool, AnomalyDetectorPlugin.AD_THREAD_POOL_NAME, latestTimeListener, false)
```
wnbts, line 346: minor. why not using configured executor name instead of hardcoding it in all places?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
```diff
@@ -15,86 +15,101 @@
...
 
+import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Map.Entry;
+import java.util.Optional;
 
 import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
 
 public class TransportState {
```
wnbts, line 25: minor. documentation is missing for the interface of this class.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
```diff
@@ -15,86 +15,101 @@
...
+        refreshLastUpdateTime();
     }
 
-    private boolean expired(Duration stateTtl, Instant now, Instant toCheck) {
-        return toCheck.plus(stateTtl).isBefore(now);
+    private void refreshLastUpdateTime() {
+        lastAccessTime = clock.instant();
+    }
+
+    public boolean expired(Duration stateTtl, Instant now) {
```
wnbts, line 112: minor. the input can be simplified to one parameter as the expiration time.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -397,32 +396,46 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
      *
      * @param failure  object that may contain exceptions thrown
      * @param detector detector object
-     * @return whether cold start runs
+     * @return exception if we get resource not found exception
```
wnbts, line 399: minor.

when resource is not found, what exception does this method return? why the general exception instead of the more specific one? if other exceptions occur, what does this method return?

also who is **we**? does it refer to the reader of the doc? the client of the interface? the developer of the implementation? or the software implementation? please avoid incorrect/ambiguous use of we when the referee is software.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -770,39 +777,110 @@ private boolean shouldStart(
...
+        if (previousException.isPresent()) {
+            Exception exception = previousException.get();
+            LOG.error("Previous exception of {}: {}", detectorId, exception);
+            if (exception instanceof EndRunException && ((EndRunException) exception).isEndNow()) {
+                return previousException;
             }
         }
 
+        stateManager.getDetectorCheckpoint(detectorId, ActionListener.wrap(checkpointExists -> {
+            if (!checkpointExists) {
```
wnbts, line 870: question. what should be returned when the checkpoint exists?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -770,39 +777,110 @@ private boolean shouldStart(
...
-            try {
-                Optional<double[][]> traingData = featureManager.getColdStartData(detector);
-                if (traingData.isPresent()) {
-                    double[][] trainingPoints = traingData.get();
-                    modelManager.trainModel(detector, trainingPoints);
-                    return true;
-                } else {
-                    throw new EndRunException(detectorId, "Cannot get training data", false);
-                }
+        Optional<Exception> previousException = stateManager.fetchColdStartException(detectorId);
```
wnbts, line 859: question. why should exception be cleared only when data point is missing?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: I don't understand.  Could you explain more?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
kaituo: I don't understand.  Could you explain?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: Changed to new AnomalyDetectionException.

Is there any place saying we cannot use we?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: This is a callback.  We don't need to return anything.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: What do you mean by "data point is missing"?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
wnbts: AnomalyDetectorPlugin.AD_THREAD_POOL_NAME is the name.

why not using a configured name like threadpool in configuration? instead of hardcoding it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
wnbts: why using two input parameters for this method when one parameter such as expiration time is sufficient?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
wnbts: look at when the method coldStartIfNoCheckPoint gets called

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: What do you mean by configuration?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
kaituo: changed

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: There are another method calling fetchColdStartException: coldStartIfNoModel

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -397,32 +396,46 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
...
+        if (previousException.isPresent()) {
+            Exception exception = previousException.get();
+            LOG.error("Previous exception of {}: {}", () -> adID, () -> exception);
+            if (exception instanceof EndRunException && ((EndRunException) exception).isEndNow()) {
+                return exception;
             }
         }
-        return false;
+        LOG.info("Trigger cold start for {}", detector.getDetectorId());
+        coldStart(detector);
```
ylwu-amzn, line 437: Is it possible that last cold start still not finished when execute `coldStart` ? Should we check the state of last cold start ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
wnbts: the value is configurable using constructor.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
wnbts: it was the question about who **we** refers to. https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Job runner will end detector job if gets`EndRunException` and `isEndNow` is true.  If `isEndNow` is false, will still stop detector job if the count of `EndRunException` exceeds `MAX_RETRY_FOR_END_RUN_EXCEPTION`(default 6 currently). For other exceptions except for `EndRunException`, job runner will record it in AD result.

How about we return this exception to JobRunner so it can be logged into AD result ? 

In this change, cold start is put into another thread pool and run asynchronously; will check `previousException` if no checkpoint or no model.  How about we check `EndRunException` which `isEndNow` is false and return it to JobRunner, so JobRunner can stop detector job if find too many `EndRunExceptions` ? That can avoid endlessly run cold start.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: The benefit to make it configurable is to make testing easy.

My concern is we may not have just one threadpool to use in this class.  For example, I want to add a dedicated cold start threadpool for high cardinality.  So we should add threadpool_ad, threadpool_coldStart for this?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: It is possible.  Let me add a check.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: Which part of the doc do you want me to read?  Any particular part related?

Anyway, I changed to "AD job execution".

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
wnbts: if there are additional dependency configuration (threadpool and executor names), they can also be passed via constructor and the implementation just needs to pick the corresponding configuration (threadpool and executor name combination)

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
wnbts: See Use 3rd person (descriptive) not 2nd person (prescriptive).

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: changed as we discussed offline

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
kaituo: good to know

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
```diff
@@ -187,7 +188,7 @@ public void maintenance() {
             String detectorId = entry.getKey();
             try {
                 TransportState state = entry.getValue();
-                if (state.expired(stateTtl, clock.instant())) {
+                if (state.expired(stateTtl)) {
```
wnbts, line 191: minor. why not do expired(expirationdate=now-ttl)? ttl is an odd parameter.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportState.java
wnbts: also all the public methods (interface)

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/TransportStateManager.java
kaituo: expirationdate is a date, while ttl is a duration.  ttl means time to leave.

wnbts: the method can just take one input parameter as the final expiration time, which can be calculated by the client however they want. in this case, it is now-tll. the method is easier to understand and use and more reusable.

kaituo: I don't understand.  Expiration date = last access time + ttl.  Why is it now-ttl?

wnbts: so this method takes one parameter that is the expiration time. if the last access time is before it, the method returns true, otherwise it returns false.

PR #209, opened 2020-08-11
Improve ux

VijayanB:
1) Improve profile display if no profiles are found
Before:
Screen Shot 2020-08-11 at 12 30 41 PM

After:

Screen Shot 2020-08-11 at 12 31 26 PM

2) Refactor for readability
3) Delete binary file.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -29,15 +29,15 @@ import (
 )
 
 const (
-	defaultFileType       = "yaml"
-	defaultConfigFileName = "config"
 	cliName               = "esad"
+	defaultConfigFileName = "config"
+	defaultFileType       = "yaml"
 	esadConfigFile        = "ESAD_CONFIG_FILE"
 	FlagConfig            = "config"
-	FlagUser              = "user"
-	FlagPassword          = "password"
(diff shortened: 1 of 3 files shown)


cli/cmd/root.go
```diff
@@ -29,15 +29,15 @@ import (
...
-	defaultConfigFileName = "config"
 	cliName               = "esad"
+	defaultConfigFileName = "config"
+	defaultFileType       = "yaml"
 	esadConfigFile        = "ESAD_CONFIG_FILE"
 	FlagConfig            = "config"
-	FlagUser              = "user"
-	FlagPassword          = "password"
 	FlagEndpoint          = "endpoint"
+	FlagPassword          = "password"
```
kaituo, line 38: What is a flag?

VijayanB: generally in cli, 
the action is called command eg: esad / aws-cli /  ls / git 
options are called flags or parameters like "git --config" / "git checkout -b branch name ( config and b are flags/parameters)

kaituo: I see.  I thought there might be some special meaning.

Issue #198, opened 2020-08-03
Change profile API to not return estimated minutes remaining until cold start is finished


ohltyler:
Currently, the progress bar on AD Kibana will show the estimated time remaining to initialize a detector. This can be confusing if this message is displayed before cold start is finished, where the actual initialization time may be much shorter if sufficient historical data exists. Change the profile api to not return any estimated time left until the cold start is finished to prevent this.

PR #210, opened 2020-08-12
not return estimated minutes remaining until cold start is finished

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/198

*Description of changes:*
Currently, the progress bar on AD Kibana will show the estimated time remaining to initialize a detector. This can be confusing if this message is displayed before cold start is finished, where the actual initialization time may be much shorter if sufficient historical data exists. This PR changes the profile api to not return any estimated time left until the cold start is finished to prevent this.

Testing done:
1. Manually verified the problem is fixed.
2. added unit test for the issue.

Before the PR:
kaituo@dev-dsk-kaituo-2a-74e3f324.us-west-2: ~
% curl -XGET "http://localhost:9200/_opendistro/_anomaly_detection/detectors/F5nQ4HMBALosA6jhE7Ni/_profile?_all=true&pretty"
{
  "state" : "INIT",
  "shingle_size" : 0,
  "coordinating_node" : "",
  "total_size_in_bytes" : 0,
  "init_progress" : {
    "percentage" : "0%",
    "estimated_minutes_left" : 128,
    "needed_shingles" : 128
  }
}

After the PR:
kaituo@dev-dsk-kaituo-2a-74e3f324.us-west-2: ~
% curl -XGET "http://localhost:9200/_opendistro/_anomaly_detection/detectors/FO7a4HMBTu_4lKyQCdgs/_profile?_all=true&pretty"
{
  "state" : "INIT",
  "shingle_size" : 0,
  "coordinating_node" : "",
  "total_size_in_bytes" : 0,
  "init_progress" : {
    "percentage" : "0%",
    "needed_shingles" : 128
  }
}

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
@@ -76,6 +76,9 @@ public class AnomalyDetectorProfileRunner {
         this.client = client;
         this.xContentRegistry = xContentRegistry;
         this.nodeFilter = nodeFilter;
+        if (requiredSamples <= 0) {
+            throw new IllegalArgumentException("required samples should be a positive number, but was " + requiredSamples);
+        }
         this.requiredSamples = requiredSamples;
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
@@ -287,14 +287,7 @@ public class AnomalyDetectorProfileRunner {
                     ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser::getTokenLocation);
                     AnomalyDetector detector = AnomalyDetector.parse(parser, detectorId);
                     long intervalMins = ((IntervalTimeConfiguration) detector.getDetectionInterval()).toDuration().toMinutes();
-                    float percent = (100.0f * totalUpdates) / requiredSamples;
-                    int neededPoints = (int) (requiredSamples - totalUpdates);
-                    InitProgressProfile initProgress = new InitProgressProfile(
-                        // rounding: 93.456 => 93%, 93.556 => 94%
-                        String.format("%.0f%%", percent),
-                        intervalMins * neededPoints,
-                        neededPoints
-                    );
+                    InitProgressProfile initProgress = computeInitProgressProfile(totalUpdates, intervalMins);
 
                     listener.onResponse(new DetectorProfile.Builder().initProgress(initProgress).build());
                 } catch (Exception t) {
@@ -306,6 +299,17 @@ public class AnomalyDetectorProfileRunner {
                 listener.failImmediately(FAIL_TO_FIND_DETECTOR_MSG + detectorId);
             }
         }, exception -> { listener.failImmediately(FAIL_TO_FIND_DETECTOR_MSG + detectorId, exception); });
+    }
+
+    private InitProgressProfile computeInitProgressProfile(long totalUpdates, long intervalMins) {
+        float percent = (100.0f * totalUpdates) / requiredSamples;
+        int neededPoints = (int) (requiredSamples - totalUpdates);
+        return new InitProgressProfile(
+            // rounding: 93.456 => 93%, 93.556 => 94%
+            String.format("%.0f%%", percent),
+            intervalMins * neededPoints,
+            neededPoints
+        );
     }
 
     private void profileModels(
@@ -357,7 +361,7 @@ public class AnomalyDetectorProfileRunner {
         return ActionListener.wrap(rcfPollResponse -> {
             long totalUpdates = rcfPollResponse.getTotalUpdates();
             if (totalUpdates < requiredSamples) {
-                processInitResponse(detectorId, profilesToCollect, listener, totalUpdates);
+                processInitResponse(detectorId, profilesToCollect, listener, totalUpdates, false);
             } else {
                 if (profilesToCollect.contains(ProfileName.STATE)) {
                     listener.onResponse(new DetectorProfile.Builder().state(DetectorState.RUNNING).build());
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -391,19 +399,26 @@ private void processInitResponse(
         String detectorId,
         Set<ProfileName> profilesToCollect,
         MultiResponsesDelegateActionListener<DetectorProfile> listener,
-        long totalUpdates
+        long totalUpdates,
+        boolean hideMinutesLeft
```
wnbts, line 403: this should be hide estimate, which affects minutes left and data points needed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -308,6 +301,17 @@ private void profileStateRelated(
         }, exception -> { listener.failImmediately(FAIL_TO_FIND_DETECTOR_MSG + detectorId, exception); });
     }
 
+    private InitProgressProfile computeInitProgressProfile(long totalUpdates, long intervalMins) {
+        float percent = (100.0f * totalUpdates) / requiredSamples;
```
ohltyler, line 305: any chance the passed in `requiredSamples` is null/0?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: you meant the variable name should be hideEstimate?  It only affect minutes left, not data points needed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: It is possible.  I added code to throw exception.

ohltyler: I see the call on line 290 is caught, what about on line 411?

ohltyler: > I see the call on line 290 is caught, what about on line 411?

nevermind, I see that's caught in the outer method. LGTM

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
wnbts: I see the data needed and time needed together. Before model training is done, they should be hidden together by a message such as "Model training is starting soon". Once that is done, the detailed information on needed data and time is now shown, such as "need x data and y time".

It'd be confusing to see "need x data and no time". What does the ux look like in this case?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -76,6 +76,9 @@ public AnomalyDetectorProfileRunner(
         this.client = client;
         this.xContentRegistry = xContentRegistry;
         this.nodeFilter = nodeFilter;
+        if (requiredSamples <= 0) {
+            throw new IllegalArgumentException("required samples should be a positive number, but was " + requiredSamples);
```
wnbts, line 80: testing. the testing of this condition is missing.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
ohltyler: As far as ux goes, we will probably add an intermediate callout saying something like "attempting to initialize with existing/historical data", and show that callout until we can retrieve `minutesLeft` from the backend profile call. Then we show the regular progress bar callout, which includes the "need x data and y time" info. See [Kibana issue](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/271) here

wnbts: The initial temporal call out is informative enough. Before a model is trained, both data and time that the (non-existent) model will need to wait for should be equally unknown.

ohltyler: If we have a separate callout, both of these would be ignored as far as frontend is concerned. Wouldn't we want a different callout in this case? Seems that we wouldn't want to show 0% progress either, when that may immediately go to 100% and disappear if there is sufficient historical data.

wnbts: The call out for the case before model training may just be "attempting to initialize with existing/historical data"/"Model training is starting soon", no data/time/progress needed to cause confusion or panic (yet).

kaituo: First, UX does not show needed shingles.  They only look at progress and estimated time.  Second, as an API user, I don't feel it causes confusion when we show needed shingles.  For example, one minute ago, we need 128 shingles.  After 1 minute, we need 0.  It means we find enough training data.  At that particular moment, the needed shingles are accurate.  If cold start if stuck or retrying, the number also gives clarification on the state.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
wnbts: why not apply a more consistent logic to make code more maintainable and understandable? that is before a model is created, the time and data it needs to wait is either both unknown or both guessed. Or, what's the benefit of having mixed results from different logic for each field?

this is a question not a blocker

kaituo: The data it needs to wait is not guessed.  The time is.  Without a checkpoint, it is accurate to say we need 128 shingles.

The benefit is provide transparency instead of guessing.

Thanks for not blocking.

PR #212, opened 2020-08-19
fix job index mapping

ylwu-amzn:
## Description of changes:

Current job index mapping can't match the [model class](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorJob.java). AD job can run successfully. This PR fixes the job index mapping to match the model class.


ylwu-amzn:
> we need this fix now? What issue it is causing?

It doesn't cause any failure. Just make the index mapping consistent with the java model class.

PR #214, opened 2020-08-20
Use goimports instead of gofmt

VijayanB:
goimports fixes imports order and also formats like gofmt.


VijayanB:
No functional change in this commit. I just formatted using goimports tool instead of gofmt tool.

PR #215, opened 2020-08-24
Upgrade from 1.9.0 to 1.10.0

VijayanB:
1. Upgraded ES-OSS from 7.8 to 7.9
2. Fixed compilation errors due to changes in 7.9
3. Added dependencies explicitly since after upgrading to 7.9, we saw error in powemockito.
4. Exclude asm transitive dependency, since those jar is already added by es in 7.9 explicitly
5. Remove jacoco.core, jacoco.report since it is part of ant.


build.gradle
```diff
@@ -301,14 +301,22 @@ dependencies {
     compile 'software.amazon.randomcutforest:randomcutforest-serialization-json:1.0-alpha'
 
     compile "org.jacoco:org.jacoco.agent:0.8.5"
-    compile "org.jacoco:org.jacoco.ant:0.8.5"
-    compile "org.jacoco:org.jacoco.core:0.8.5"
-    compile "org.jacoco:org.jacoco.report:0.8.5"
+    compile ("org.jacoco:org.jacoco.ant:0.8.5") {
+        exclude group: 'org.ow2.asm', module: 'asm-commons'
```
ylwu-amzn, line 305: Just asking, why excludes these modules?

VijayanB: In ES 7.9, they added those dependencies ( https://github.com/elastic/elasticsearch/blob/master/modules/lang-painless/build.gradle#L36 ). We are also including this dependency through transitive from code-coverage . This caused jar Hell error ( es checks whether any multiple jars of same or different version is being included ). Hence, i excluded this jar to be added into our plugin while downloading org.jacoco:org.jacoco.ant for code coverage.

VijayanB:
Launched  ESS-OSS + Job-scheduler +AD 
From CLI following API is tested
Create, start, stop, get, delete, update

PR #221, opened 2020-09-02
Validation API Backend for AD everywhere

amitgalitz:
Description of changes:
Added a new API endpoint to AD in order to validate anomaly detector configurations and return back if any failures or suggestions occurred. The validation API takes an input of detector configurations and then validates it against both general checks and against the data source to know if the configurations will likely lead to successful anomaly detector creation.
Endpoint: Post _opendistro/_anomaly_detection/detectors/_validate
Input : Anomaly Detector configurations
image

Overall the validation checks if any fields are missing, if any data can be found with the given filter query and feature query. I also check whether the feature aggregation type is valid for the field type. The last two steps of the validation, checks if the detector interval used will produce data that is dense enough, recommending a different interval if this isn't the case in milliseconds. The API also checks for the last seen data point and returns it as the window delay recommendation.

Example input for detector interval that is too short :
{
"name": "test-detector-02",
"description": "Test detector",
"time_field": "timestamp",
"indices": ["test-index-sparse-3-demo"],
"feature_attributes": [
{
"feature_name": "total_order",
"feature_enabled": true,
"aggregation_query": {
"total_order": {
"max": {
"field": "field-1"
}
}
}
}
],
"detection_interval": {
"period": {
"interval": 10,
"unit": "Minutes"
}
},
"window_delay": {
"period": {
"interval": 150,
"unit": "Minutes"
}
}
}

Output:

{
"failures": {},
"suggestedChanges": {
"detection_interval": [
"4320000" *90 minutes recommendation
]
}
}

Testing Done:
Integration testing added
manually tested that API works with both dense and sparse sample data and some of the tms-issue data.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -90,6 +90,7 @@ import com.amazon.opendistroforelasticsearch.ad.rest.RestIndexAnomalyDetectorAct
 import com.amazon.opendistroforelasticsearch.ad.rest.RestSearchAnomalyDetectorAction;
 import com.amazon.opendistroforelasticsearch.ad.rest.RestSearchAnomalyResultAction;
 import com.amazon.opendistroforelasticsearch.ad.rest.RestStatsAnomalyDetectorAction;
+import com.amazon.opendistroforelasticsearch.ad.rest.RestValidateAnomalyDetectorAction;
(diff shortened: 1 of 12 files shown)


kaituo:
Unit test coverage drops and cause workflow to fail.  Is it possible to write unit tests to cover it? Or can you let the workflow count your IT tests to the new coverage?

kaituo:
> Unit test coverage drops and cause workflow to fail. Is it possible to write unit tests to cover it? Or can you let the workflow count your IT tests to the new coverage?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/ValidateAnomalyDetectorActionHandler.java may require more unit tests.

PR #227, opened 2020-09-09
Fix build failure

VijayanB:
Fix build failure due to merge commit

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -97,12 +97,10 @@ configurations.all {
 }
 
 configurations {
-    all*.exclude group: 'org.ow2.asm', module: 'asm-commons'
     testCompile {
         exclude group: 'org.elasticsearch', module: 'securemock'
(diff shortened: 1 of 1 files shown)


VijayanB:
Merge commit pulled some changes from last version. Reverting it to make it consistent with master.

build.gradle
```diff
@@ -97,12 +97,10 @@ configurations.all {
 }
 
 configurations {
-    all*.exclude group: 'org.ow2.asm', module: 'asm-commons'
```
ylwu-amzn, line 100: Doesn't see this line in master branch, https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/build.gradle#L99

Seems this line was caused by a unreviewed [commit](https://github.com/opendistro-for-elasticsearch/anomaly-detection/commit/88d85650126ce9816694a7c5a32eb643a22ce702). We should review back-porting PR.

kaituo (approved):

PR #229, opened 2020-09-12
CLI: Download Detector Configuration as File

VijayanB:
Download configuration as files helps users to save it in source code control and maintains it.
This will also let users to share detector configuration across users.

### DIFF
--- a/cli/cmd/cat.go
+++ b/cli/cmd/cat.go
@@ -17,6 +17,8 @@ import (
 	entity "esad/internal/entity/ad"
 	"esad/internal/handler/ad"
 	"fmt"
+	"io"
+	"os"
 
 	"github.com/spf13/cobra"
 )
@@ -38,23 +40,28 @@ var catCmd = &cobra.Command{
 			}
 			return
 		}
-		idStatus, _ := cmd.Flags().GetBool("id")
-		commandHandler, err := getCommandHandler()
-		if err != nil {
-			fmt.Println(err)
-		}
-		// default is name
-		action := ad.GetAnomalyDetectorsByNamePattern
-		if idStatus {
-			action = getDetectorsByID
-		}
-		results, err := getDetectors(commandHandler, args, action)
-		if err != nil {
-			fmt.Println(err)
-			return
-		}
-		printDetectors(results)
+		printDetectors(Println, cmd, args)
 	},
+}
+
+//printDetectors print detectors
+func printDetectors(f func(*entity.DetectorOutput), cmd *cobra.Command, detectors []string) {
+	idStatus, _ := cmd.Flags().GetBool("id")
+	commandHandler, err := getCommandHandler()
+	if err != nil {
+		fmt.Println(err)
+	}
+	// default is name
+	action := ad.GetAnomalyDetectorsByNamePattern
+	if idStatus {
+		action = getDetectorsByID
+	}
+	results, err := getDetectors(commandHandler, detectors, action)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	print(f, results)
 }
 
 func getDetectors(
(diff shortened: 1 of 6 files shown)


cli/cmd/cat.go
```diff
@@ -38,25 +40,30 @@ var catCmd = &cobra.Command{
...
 	},
 }
 
+//printDetectors print detectors
+func printDetectors(f func(*entity.DetectorOutput), cmd *cobra.Command, detectors []string) {
+	idStatus, _ := cmd.Flags().GetBool("id")
+	commandHandler, err := getCommandHandler()
+	if err != nil {
+		fmt.Println(err)
+	}
```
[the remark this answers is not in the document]
VijayanB: Ack

wnbts (approved):

cli/cmd/download.go
```diff
@@ -0,0 +1,67 @@
...
+
+//WriteInFile writes detector's configuration on file
+//file will be created inside current working directory,
+//with detector name as file name
+func WriteInFile(d *entity.DetectorOutput) error {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	filePath := filepath.Join(cwd, d.Name)
```
ylwu-amzn, line 52: Add file extension?

cli/cmd/download.go
```diff
@@ -0,0 +1,67 @@
...
+	f, err := os.Create(filePath)
```
ylwu-amzn, line 53: If the `filePath` exists, will `os.Create` override the existing file?

ylwu-amzn: From this doc, https://golang.org/pkg/os/#Create
> If the file already exists, it is truncated
How about check file exists or not before create ? If file exists, we should not truncate it directly in case user execute command by mistake.

cli/cmd/download.go
```diff
@@ -0,0 +1,67 @@
...
+		err := printDetectors(WriteInFile, cmd, args)
+		displayError(err, commandDownload)
+	},
+}
+
+//WriteInFile writes detector's configuration on file
+//file will be created inside current working directory,
+//with detector name as file name
+func WriteInFile(d *entity.DetectorOutput) error {
+	cwd, err := os.Getwd()
```
ylwu-amzn, line 48: Is it possible to output detector into a different directory other than current directory?  Can user specify the absolute file path? For example, user want to download detector and output it into "/tmp/my_detectors.json".

cli/cmd/download.go
VijayanB: Added .json

cli/cmd/download.go
VijayanB: I introduced a flag -i similar to  cp in linux commands, which prompt to stdout whether to overwrite the file or not.

cli/cmd/download.go
VijayanB: Yes definitely possible. I am planning to include this feature as next PR.

VijayanB: Added allow user to provide output folder as new commit.

ylwu-amzn: Cool, thanks for the change

cli/cmd/download.go
ylwu-amzn: The flag `-i` is optional? What if user don't add `-i`?

VijayanB: It is interactive. By default download will overwrite but if user passes -i, it will prompt for user confirmation. Do you think prompt should be default?

ylwu-amzn: I think prompt should be default for this case, not safe to override user's existing file without any prompt.

ylwu-amzn (approved):
LGTM. Thanks for the change.

PR #232, opened 2020-09-14
add anomaly feature attribution to model output

wnbts:
*Description of changes:* This pr adds normalized anomaly score attribution to anomaly detection model output. Anomaly score is attributed to each feature dimension of a current single data point and normalized to 1 for easier consumption. Further changes to external API and persistence is not included in this pr.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
@@ -226,7 +226,7 @@ public class ModelManager {
                 double confidence = rcfResults.stream().mapToDouble(r -> r.getConfidence() * r.getForestSize()).sum() / Math
                     .max(rcfNumTrees, totalForestSize);
                 double[] attribution = combineAttribution(rcfResults, numFeatures, totalForestSize);
-                combinedResult = new CombinedRcfResult(score, confidence, combineAttribution(rcfResults, numFeatures, totalForestSize));
+                combinedResult = new CombinedRcfResult(score, confidence, attribution);
             }
         }
         return combinedResult;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
@@ -53,6 +53,7 @@ import com.amazon.opendistroforelasticsearch.ad.ml.rcf.CombinedRcfResult;
 import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
 import com.amazon.opendistroforelasticsearch.ad.util.DiscoveryNodeFilterer;
 import com.amazon.randomcutforest.RandomCutForest;
+import com.amazon.randomcutforest.returntypes.DiVector;
 import com.amazon.randomcutforest.serialize.RandomCutForestSerDe;
 import com.google.gson.Gson;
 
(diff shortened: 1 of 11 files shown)


kaituo (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -349,13 +370,25 @@ public void getRcfResult(String detectorId, String modelId, double[] point, Acti
...
-        listener.onResponse(new RcfResult(score, confidence, forestSize));
+        listener.onResponse(new RcfResult(score, confidence, forestSize, attribution));
+    }
+
+    private double[] getAnomalyAttribution(RandomCutForest rcf, double[] point) {
+        DiVector vec = rcf.getAnomalyAttribution(point);
+        vec.renormalize(1d);
+        double[] attribution = new double[vec.getDimensions()];
+        for (int i = 0; i < attribution.length; i++) {
+            attribution[i] = vec.getHighLowSum(i);
```
kaituo, line 389: Questions: 
First, how to interpret both high and low are non-zero?  Is it really high or low?  Does it mean RCF trees think the value can be both higher or lower than the recently observed data trends for that column?  Do we need a majority win rule to say it is actually high or low?
Second, when doing a high low sum, we lose direction.  Is there any way to preserve the direction?
Third, when users see two features' attribution like x: 1% and y 99%, it tells users y is the place anomaly happens.  It might as well not to show x's 1%.  I feel an attribution score less than 1/d (d is the number of features) is not useful to users.  Any comments on this?

wnbts: 1. for a given sum, the larger value indicates the direction/relative position. Note even data within normal range has non-zero values for both.
2. since the current ux design only shows feature contribution, only contribution is computed here. The direction is not lost. It can be added when it's needed.
3. 1/99 or 0/100 probably won't make a difference for users. In general, additional rules should be avoided for simplicity and correctness as they might introduce their own problems. As an extreme example, the contribution from two features is 49/51, if 49 is omitted, after normalization the result could be 0/100.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -206,28 +207,48 @@ public ModelManager(
...
             if (totalForestSize == 0) {
-                combinedResult = new CombinedRcfResult(0, 0);
+                combinedResult = new CombinedRcfResult(0, 0, new double[0]);
             } else {
                 double score = rcfResults.stream().mapToDouble(r -> r.getScore() * r.getForestSize()).sum() / totalForestSize;
                 double confidence = rcfResults.stream().mapToDouble(r -> r.getConfidence() * r.getForestSize()).sum() / Math
                     .max(rcfNumTrees, totalForestSize);
-                combinedResult = new CombinedRcfResult(score, confidence);
+                double[] attribution = combineAttribution(rcfResults, numFeatures, totalForestSize);
+                combinedResult = new CombinedRcfResult(score, confidence, combineAttribution(rcfResults, numFeatures, totalForestSize));
```
LiuJoyceC, line 229: [minor] It looks like the variable `attribution` was meant to be used in line 229, but didn't get used and instead `combineAttribution` was invoked again with identical args.

LiuJoyceC (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
wnbts: good catch. I fixed it in the new commit.

PR #233, opened 2020-09-14
CLI: Update Detector Configurations

VijayanB:
Update Detector Configurations

### DIFF
--- a/cli/cmd/update.go
+++ b/cli/cmd/update.go
@@ -22,7 +22,7 @@ import (
 const (
 	commandUpdate = "update"
 	flagForce     = "force"
-	flagRestart   = "restart"
+	flagStart     = "start"
 )
 
 //updateCmd updates detectors based on file configuration
@@ -39,7 +39,7 @@ var updateCmd = &cobra.Command{
 			return
 		}
 		force, _ := cmd.Flags().GetBool(flagForce)
-		restart, _ := cmd.Flags().GetBool(flagRestart)
+		restart, _ := cmd.Flags().GetBool(flagStart)
 		err := updateDetectors(args, force, restart)
 		if err != nil {
 			fmt.Println(commandUpdate, "command failed")
@@ -51,7 +51,7 @@ var updateCmd = &cobra.Command{
 func init() {
 	esadCmd.AddCommand(updateCmd)
 	updateCmd.Flags().BoolP(flagForce, "f", false, "Will stop detector and update it")
-	updateCmd.Flags().BoolP(flagRestart, "r", false, "Will start detector if update is successful")
+	updateCmd.Flags().BoolP(flagStart, "r", false, "Will start detector if update is successful")
 }
 
 func updateDetectors(fileNames []string, force bool, restart bool) error {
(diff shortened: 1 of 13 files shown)


VijayanB:
Update Feature Workflow:
1) User will download detector configuration using download command. This is already merged.
2)User will make changes to the downloaded file
3) User will use update command to pass file name
4) if detector is updated after user downloaded file, cmd will fail with message please download latest version and merge
5)if detector is running, it will fail saying that detector is running.
6) User can either pass --force , cmd will automatically stop the detector and override the existing setting.
7) if user pass --restart, the detector will also start after successful update.

kaituo (approved):

cli/cmd/update.go
```diff
@@ -0,0 +1,69 @@
...
+	handler "esad/internal/handler/ad"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+const (
+	commandUpdate = "update"
+	flagForce     = "force"
+	flagRestart   = "restart"
```
ylwu-amzn, line 25: User can only "restart" a running detector? If current detector hasn't been started, is it possible user only set "restart" as true to start it after update ? 
From the description "Will start detector if update is successful", it's better to name the flag as "start".

cli/internal/controller/ad/ad.go
```diff
@@ -520,3 +521,48 @@ func (c controller) GetDetectorsByName(ctx context.Context, pattern string, disp
 	}
 	return output, nil
 }
+
+//UpdateDetector updates detector based on DetectorID, if force is enabled, it overrides without checking whether
+// user downloaded latest version before updating it, if restart is true, detector will be started after update
```
ylwu-amzn, line 526: Check detector is running or not if "force" is false?

cli/internal/controller/ad/ad.go
```diff
@@ -520,3 +521,48 @@ func (c controller) GetDetectorsByName(ctx context.Context, pattern string, disp
...
+			fmt.Sprintf(
+				"esad will update detector: %s . Do you want to proceed? please type (y)es or (n)o and then press enter:",
+				input.ID,
+			),
+		),
+	)
+	if !proceed {
+		return nil
+	}
+	err := c.StopDetector(ctx, input.ID)
```
ylwu-amzn, line 552: If "force" is false, no need to stop detector.

cli/internal/controller/ad/ad.go
```diff
@@ -520,3 +521,48 @@ func (c controller) GetDetectorsByName(ctx context.Context, pattern string, disp
...
+	}
+	if !force {
+		latestDetector, err := c.GetDetector(ctx, input.ID)
+		if err != nil {
+			return err
+		}
+		if latestDetector.LastUpdatedAt > input.LastUpdatedAt {
+			return fmt.Errorf(
+				"new version for detector is available. Please fetch latest version and then merge your changes")
+		}
```
ylwu-amzn, line 539: Check if detector is running? If detector is running and `force` is false, we should fail immediately.

cli/internal/controller/ad/ad.go
VijayanB: Ack

cli/internal/controller/ad/ad.go
VijayanB: Ack.
Screen Shot 2020-10-06 at 3 36 47 PM

cli/cmd/update.go
VijayanB: Ack

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #240, opened 2020-10-07
Verifying multi-entity detectors

kaituo:
*Issue #, if available:*

*Description of changes:*

This PR adds categorical fields' number and type check. We only support one categorical field, and the categorical field can only be of type keyword and ip.  We also limit the max multi-entity detectors to 10.

Testing done:
1. added unit tests
2. did manual testing.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #241, opened 2020-10-07
Selectively store anomaly results when index pressure is high

kaituo:
*Issue #, if available:*

*Description of changes:*
Our performance evaluation has observed a large number of EsRejectedExecutionException due to indexing too many anomaly results. This PR adds a customized bulk transport action to deal with high index memory pressure. When the index pressure exceeds 80%, we only index non-zero anomaly grade index results and a random subset of zero anomaly grade index results.

This PR adds a few supporting data structures related changes:1. add constructors in AnomalyResult/FeatureData to use them in the newly added transport action.2. add Entity class to represent an entity in anomaly results.
This PR also renames ANOMALY_RESULT_INDEX to ANOMALY_RESULT_INDEX_ALIAS and moves it to the CommonName class. The name refers to an alias and is accessed in multiple places.

Testing done:
1. we don't observe EsRejectedExecutionException after applying the change (together with a series of other related changes).
2. added unit tests.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
@@ -161,7 +161,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         this.filterQuery = filterQuery;
         this.detectionInterval = detectionInterval;
         this.windowDelay = windowDelay;
-        this.shingleSize = shingleSize;
+        this.shingleSize = getShingleSize(shingleSize, categoryField);
         this.uiMetadata = uiMetadata;
         this.schemaVersion = schemaVersion;
         this.lastUpdateTime = lastUpdateTime;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
@@ -114,7 +114,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
      * @param uiMetadata        metadata used by Kibana
      * @param schemaVersion     anomaly detector index mapping version
      * @param lastUpdateTime    detector's last update time
-     * @param categoryField     a list of partition fields
+     * @param categoryFields     a list of partition fields
      */
     public AnomalyDetector(
         String detectorId,
@@ -161,7 +161,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         this.filterQuery = filterQuery;
         this.detectionInterval = detectionInterval;
         this.windowDelay = windowDelay;
-        this.shingleSize = getShingleSize(shingleSize, categoryField);
+        this.shingleSize = getShingleSize(shingleSize, categoryFields);
         this.uiMetadata = uiMetadata;
         this.schemaVersion = schemaVersion;
         this.lastUpdateTime = lastUpdateTime;
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
@@ -131,7 +131,7 @@ public class AnomalyDetector implements Writeable, ToXContentObject {
         Map<String, Object> uiMetadata,
         Integer schemaVersion,
         Instant lastUpdateTime,
-        List<String> categoryField
+        List<String> categoryFields
     ) {
         if (Strings.isBlank(name)) {
             throw new IllegalArgumentException("Detector name should be set");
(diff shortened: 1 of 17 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -151,6 +210,7 @@ public AnomalyDetector(
         this.uiMetadata = uiMetadata;
         this.schemaVersion = schemaVersion;
         this.lastUpdateTime = lastUpdateTime;
+        this.categoryField = null;
```
yizheliu-amazon, line 213: You may refactor this constructor to 
```
public AnomalyDetector(....,lastUpdateTime) {
     new AnomalyDetector(....,lastUpdateTime, null);
}

File: src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyResult.java [hunk 2]
```diff
@@ -91,6 +95,55 @@ public AnomalyResult(
         this.executionStartTime = executionStartTime;
         this.executionEndTime = executionEndTime;
         this.error = error;
+        this.entity = null;
```
yizheliu-amazon, line 98: same comment as above to refactor constructor

yizheliu-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -105,7 +109,62 @@
      * @param uiMetadata        metadata used by Kibana
      * @param schemaVersion     anomaly detector index mapping version
      * @param lastUpdateTime    detector's last update time
+     * @param categoryField     a list of partition fields
      */
+    public AnomalyDetector(
```
ylwu-amzn, line 114: This PR has some overlap with #240 ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -428,7 +497,9 @@ public TimeConfiguration getWindowDelay() {
     }
 
     public Integer getShingleSize() {
-        return shingleSize == null ? DEFAULT_SHINGLE_SIZE : shingleSize;
+        return shingleSize == null
```
ylwu-amzn, line 500: How about we move this logic to detector constructor？Always set shingle size as default value if it's null. Then this get method can return `shingleSize` directly; and we should persist shingle size in detector document even it's default shingle size in case the default shingle size changes in future.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/Entity.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Categorical field name and its value
+ * @author kaituo
+ *
+ */
+public class Entity implements ToXContentObject, Writeable {
+    public static final String ENTITY_NAME_FIELD = "name";
+    public static final String ENTITY_VALUE_FIELD = "value";
+
+    private final String name;
```
ylwu-amzn, line 41: `name` is category field name and `value` is the entity's unique value? For example, `name` will be "IP" and `value` will be `53.230.198.42` ?
Is it necessary to store same `name` for every AD result doc ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADResultBulkTransportAction.java
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+            // if exceeding 100% of hard limit, try our luck and only index non-zero grade result
+            for (AnomalyResult result : request.getAnomalyResults()) {
+                if (result.getAnomalyGrade() > 0) {
+                    addResult(bulkRequest, result);
+                }
+            }
+        }
+
+        client
```
ylwu-amzn, line 110: Check bulk request size, if no anomaly result, no need to execute bulk index.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
kaituo: Yes, changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/Entity.java
kaituo: yes, your understanding is correct.  I am mostly doing this for the future: in case we have multiple categorical fields and the parser knows which value belongs to which field.  Open to change if we have a solution.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/ADResultBulkTransportAction.java
kaituo: good catch.  Added.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/Entity.java
ylwu-amzn: Let's keep it as is currently. Remove these duplicate fields needs some snapshot mechanism to track the old version of detector configuration. Let's discuss more to make a decision.

ylwu-amzn (approved):
LGTM. Thanks for the change

PR #247, opened 2020-10-13
Add entity model training

kaituo:
Note: since there are a lot of dependencies, I only list the main class and test code to save reviewers' time.  The build will fail due to missing dependencies.  I will use that PR just for review.  will not merge it.  Will have a big one in the end and merge once after all review PRs get approved.

*Issue #, if available:*

*Description of changes:*
This PR adds entity model training funcitons. When the models are missing in the cache and checkpoint, we need to run queries to get training data and then train models.  To collect training data, we sample historical data and linearly interpolate data points between the samples.

Specifically, we first note the maximum and minimum timestamp, and sample at most 24 points (with 60 points apart between two neighboring samples) between those minimum and maximum timestamps.  Samples can be missing.  We only interpolate points between present neighboring samples. We then transform samples and interpolate points to shingles. Finally, full shingles will be used for cold start.

Testing done:
1. add unit tests
2. end-to-end testing

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
@@ -21,6 +21,7 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -103,7 +104,7 @@ public class EntityColdStarter {
      * @param interpolator Used to generate data points between samples.
      * @param searchFeatureDao Used to issue ES queries.
      * @param shingleSize The size of a data point window that appear consecutively.
-          * @param thresholdMinPvalue min P-value for thresholding
+     * @param thresholdMinPvalue min P-value for thresholding
      * @param thresholdMaxRankError  max rank error for thresholding
      * @param thresholdMaxScore max RCF score to thresholding
      * @param thresholdNumLogNormalQuantiles num of lognormal quantiles for thresholding
@@ -316,7 +317,6 @@ public class EntityColdStarter {
         }
 
         double[] scores = new double[dataPoints.length];
-        Arrays.fill(scores, 0.);
 
         for (int j = 0; j < dataPoints.length; j++) {
             scores[j] = rcf.getAnomalyScore(dataPoints[j]);
@@ -469,11 +469,10 @@ public class EntityColdStarter {
         // adjust if numStrides is more than the max samples
         if (maxTrainSamples - 1 < numStrides) {
             numStrides = maxTrainSamples - 1;
-            startMilli = endMilli - bucketSize * stride * numStrides;
         }
         List<Entry<Long, Long>> sampleRanges = Stream
             .iterate(endMilli, i -> i - stride * bucketSize)
-            .limit(numStrides + 1)
+            .limit(numStrides)
             .map(time -> new SimpleImmutableEntry<>(time - bucketSize, time))
             .collect(Collectors.toList());
         return sampleRanges;
@@ -537,14 +536,16 @@ public class EntityColdStarter {
      */
     private void combineTrainSamples(List<double[][]> coldstartDatapoints, String modelId, ModelState<EntityModel> entityState) {
         EntityModel model = entityState.getModel();
-        if (model != null) {
-            for (double[][] consecutivePoints : coldstartDatapoints) {
-                for (int i = 0; i < consecutivePoints.length; i++) {
-                    model.addSample(consecutivePoints[i]);
-                }
-            }
-            // save to checkpoint
-            checkpointDao.write(entityState, modelId, true);
+        if (model == null) {
+            model = new EntityModel(modelId, new ArrayDeque<>(), null, null);
         }
+        for (double[][] consecutivePoints : coldstartDatapoints) {
+            for (int i = 0; i < consecutivePoints.length; i++) {
+                model.addSample(consecutivePoints[i]);
+            }
+        }
+        // save to checkpoint
+        checkpointDao.write(entityState, modelId, true);
+
     }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
@@ -23,7 +23,6 @@ import java.time.Instant;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map.Entry;
@@ -37,12 +36,12 @@ import java.util.stream.Stream;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.core.util.Throwables;
 import org.apache.logging.log4j.message.ParameterizedMessage;
 import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.support.ThreadedActionListener;
 import org.elasticsearch.common.lease.Releasable;
 import org.elasticsearch.common.settings.Settings;
-import org.elasticsearch.common.util.concurrent.EsRejectedExecutionException;
 import org.elasticsearch.threadpool.ThreadPool;
 
 import com.amazon.opendistroforelasticsearch.ad.AnomalyDetectorPlugin;
@@ -209,10 +208,11 @@ public class EntityColdStarter {
                     logger.info("Cannot get training data for {}", modelId);
                 }
             }, exception -> {
-                if (exception instanceof EsRejectedExecutionException || exception instanceof RejectedExecutionException) {
+                Throwable cause = Throwables.getRootCause(exception);
+                if (cause instanceof RejectedExecutionException) {
                     logger.error("too many requests");
                     lastThrottledColdStartTime = Instant.now();
-                } else if (exception instanceof AnomalyDetectionException) {
+                } else if (cause instanceof AnomalyDetectionException || exception instanceof AnomalyDetectionException) {
                     // e.g., cannot find anomaly detector
                     nodeStateManager.setLastColdStartException(detectorId, (AnomalyDetectionException) exception);
                 } else {
@@ -276,7 +276,9 @@ public class EntityColdStarter {
         }
 
         EntityModel model = entityState.getModel();
-        assert (model != null);
+        if (model == null) {
+            model = new EntityModel(modelId, new ArrayDeque<>(), null, null);
+        }
         model.setRcf(rcf);
         double[] joinedScores = new double[totalLength];
 
@@ -465,11 +467,8 @@ public class EntityColdStarter {
     ) {
         long bucketSize = ((IntervalTimeConfiguration) detector.getDetectionInterval()).toDuration().toMillis();
         int numBuckets = (int) Math.floor((endMilli - startMilli) / (double) bucketSize);
-        int numStrides = (int) Math.ceil(numBuckets / (double) stride);
         // adjust if numStrides is more than the max samples
-        if (maxTrainSamples - 1 < numStrides) {
-            numStrides = maxTrainSamples - 1;
-        }
+        int numStrides = Math.min((int) Math.floor(numBuckets / (double) stride), maxTrainSamples);
         List<Entry<Long, Long>> sampleRanges = Stream
             .iterate(endMilli, i -> i - stride * bucketSize)
             .limit(numStrides)
@@ -546,6 +545,5 @@ public class EntityColdStarter {
         }
         // save to checkpoint
         checkpointDao.write(entityState, modelId, true);
-
     }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
@@ -1,0 +1,550 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.ml;
+
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.COOLDOWN_MINUTES;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.Queue;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.DoubleStream;
+import java.util.stream.Stream;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.support.ThreadedActionListener;
+import org.elasticsearch.common.lease.Releasable;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.common.util.concurrent.EsRejectedExecutionException;
+import org.elasticsearch.threadpool.ThreadPool;
+
+import com.amazon.opendistroforelasticsearch.ad.AnomalyDetectorPlugin;
+import com.amazon.opendistroforelasticsearch.ad.NodeStateManager;
+import com.amazon.opendistroforelasticsearch.ad.common.exception.AnomalyDetectionException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+        int stride,
+        int maxTrainSamples
+    ) {
+        long bucketSize = ((IntervalTimeConfiguration) detector.getDetectionInterval()).toDuration().toMillis();
+        int numBuckets = (int) Math.floor((endMilli - startMilli) / (double) bucketSize);
+        int numStrides = (int) Math.ceil(numBuckets / (double) stride);
+        // adjust if numStrides is more than the max samples
+        if (maxTrainSamples - 1 < numStrides) {
+            numStrides = maxTrainSamples - 1;
+            startMilli = endMilli - bucketSize * stride * numStrides;
```
wnbts, line 472: minor. this variable is not used.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+        long endMilli,
+        int stride,
+        int maxTrainSamples
+    ) {
+        long bucketSize = ((IntervalTimeConfiguration) detector.getDetectionInterval()).toDuration().toMillis();
+        int numBuckets = (int) Math.floor((endMilli - startMilli) / (double) bucketSize);
+        int numStrides = (int) Math.ceil(numBuckets / (double) stride);
+        // adjust if numStrides is more than the max samples
+        if (maxTrainSamples - 1 < numStrides) {
+            numStrides = maxTrainSamples - 1;
```
wnbts, line 471: why is stride not min(numStrides, maxTrainSamples)?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+            startMilli = endMilli - bucketSize * stride * numStrides;
+        }
+        List<Entry<Long, Long>> sampleRanges = Stream
+            .iterate(endMilli, i -> i - stride * bucketSize)
+            .limit(numStrides + 1)
```
wnbts, line 476: the results should be numStrides

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+                        }
+                        if (coldStartData.isEmpty()) {
+                            listener.onResponse(Optional.empty());
+                        } else {
+                            listener.onResponse(Optional.of(coldStartData));
+                        }
+                    }, listener::onFailure);
+
+                    searchFeatureDao
+                        .getColdStartSamplesForPeriods(
```
wnbts, line 414: are missing values from count/sum/etc aggregations filtered out?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+                        startTimeMs,
+                        endTimeMs,
+                        maxSampleStride,
+                        maxTrainSamples
+                    );
+
+                    ActionListener<List<Optional<double[]>>> getFeaturelistener = ActionListener.wrap(featureSamples -> {
+                        ArrayList<double[]> continuousSampledFeatures = new ArrayList<>(maxTrainSamples);
+
+                        // featuresSamples are in ascending order of time.
```
wnbts, line 375: the time ranges in feature query are in descending order. when does the reversing happen?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+        EntityModel model = entityState.getModel();
+        if (model != null) {
+            totalSize += model.getSamples().size();
+        }
+
+        return totalSize >= this.numMinSamples;
+    }
+
+    /**
+     * TODO: make it work for shingle
```
wnbts, line 527: isn't given data already shingle data?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+     * without reorder based on timestamp is fine.  RCF on one-dimensional datapoints
+     * without shingling is similar to just using CDF sketch on the values.  We
+     * are just finding extreme values.
+     *
+     * @param coldstartDatapoints training data generated from cold start
+     * @param entityState entity State
+     */
+    private void combineTrainSamples(List<double[][]> coldstartDatapoints, String modelId, ModelState<EntityModel> entityState) {
+        EntityModel model = entityState.getModel();
+        if (model != null) {
```
wnbts, line 540: if this is null, why not set the model and keep the data?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+            // we cannot get last RCF score since cold start happens asynchronously
+            coldStart(modelId, entityName, detectorId, modelState);
+        } else {
+            double[][] trainData = featureManager.batchShingle(samples.toArray(new double[0][0]), this.shingleSize);
+            trainModelFromDataSegments(Collections.singletonList(trainData), modelId, modelState);
+        }
+    }
+
+    /**
+     * TODO: make it work for shingle.
```
wnbts, line 507: which part doesn't work for shingle?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+     * @param numberOfTrees The number of trees in this forest.
+     * @param rcfTimeDecay rcf samples time decay constant
+     * @param numMinSamples The number of points required by stream samplers before
+     *  results are returned.
+     * @param maxSampleStride Sample distances measured in detector intervals.
+     * @param maxTrainSamples Max train samples to collect.
+     * @param interpolator Used to generate data points between samples.
+     * @param searchFeatureDao Used to issue ES queries.
+     * @param shingleSize The size of a data point window that appear consecutively.
+          * @param thresholdMinPvalue min P-value for thresholding
```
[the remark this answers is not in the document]
kaituo: fixed

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+        this.nodeStateManager = nodeStateManager;
+        this.rcfSampleSize = rcfSampleSize;
+        this.numberOfTrees = numberOfTrees;
+        this.rcfTimeDecay = rcfTimeDecay;
+        this.numMinSamples = numMinSamples;
+        this.maxSampleStride = maxSampleStride;
+        this.maxTrainSamples = maxTrainSamples;
+        this.interpolator = interpolator;
+        this.searchFeatureDao = searchFeatureDao;
+        this.shingleSize = shingleSize;
```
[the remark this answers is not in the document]
kaituo: yeah, currently I don't implement select shingle size based on detector config.  Will do that after release.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+            .lambda(rcfTimeDecay)
+            .outputAfter(numMinSamples)
+            .parallelExecutionEnabled(false)
+            .build();
+        List<double[]> allScores = new ArrayList<>();
+        int totalLength = 0;
+        // get continuous data points and send for training
+        for (double[][] continuousDataPoints : dataPoints) {
+            double[] scores = trainRCFModel(continuousDataPoints, modelId, rcf);
+            allScores.add(scores);
```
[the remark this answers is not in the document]
kaituo: I tried.  I  end up with a List<Double> and have to convert to Double[], then double[].  Not sure the amount of code is smaller and boxing/unboxing back and forth is not efficient.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+            totalLength += scores.length;
+        }
+
+        EntityModel model = entityState.getModel();
+        assert (model != null);
```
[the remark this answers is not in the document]
kaituo: The model might have already contained samples passed from upstream callers.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+            thresholdDownsamples,
+            thresholdMaxSamples
+        );
+        threshold.train(joinedScores);
+        model.setThreshold(threshold);
+
+        entityState.setLastUsedTime(clock.instant());
+
+        // save to checkpoint
+        checkpointDao.write(entityState, modelId, true);
```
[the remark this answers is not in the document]
kaituo: I did.  The write just write to a buffer.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,550 @@
...
+     * @param rcf RCF model to be trained
+     * @return scores returned by RCF models
+     */
+    private double[] trainRCFModel(double[][] dataPoints, String modelId, RandomCutForest rcf) {
+        if (dataPoints.length == 0 || dataPoints[0].length == 0) {
+            throw new IllegalArgumentException("Data points must not be empty.");
+        }
+
+        double[] scores = new double[dataPoints.length];
+        Arrays.fill(scores, 0.);
```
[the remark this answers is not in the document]
kaituo: removed

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: Please read https://github.com/kaituo/anomaly-detection/blob/test_hc3/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java#L670-L689

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: I am reusing SearchFeatureDao.parseAggregations (https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java#L561-L571) to parse each bucket.  Missing values results in empty bucket , which should give Optional.empty(), right?  If yes, then those will be filtered out.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: If maxTrainSamples is 5, the stride would be 4, right?  I start with the first point, and need 4 strides to reach the last.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: good catch.  Removed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: right, fixed

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: Have to consider timestamp of each data point for shingling.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: Current implementation only gives unshingled data.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: good point.  Changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
```diff
@@ -0,0 +1,551 @@
...
+        if (model != null) {
+            totalSize += model.getSamples().size();
+        }
+
+        return totalSize >= this.numMinSamples;
+    }
+
+    /**
+     * TODO: make it work for shingle
+     * Precondition: we don't have enough training data.
```
wnbts, line 527: question. it there an indicator that the historical data for an entity has been retrieved or attempted to retrieve? if an entitys get only 10 samples, will it get the same 10 samples at a later time? Or, there is no data for an entity, will the system try repeatedly?

kaituo: We won't retry within one hour for an entity.  Do you think this is long enough?

wnbts: there will be some issues. In the case where samples are scarce, say just 1 data point, due to repeated process, the training data only contains the same data points. If there is no data at all, this process will not end.

kaituo: Yes.  If the value is too large, we have long initialization issue.  If it is too small, we can waste resources.  Given our cold start is rate limited, the latter's impact is not that bad.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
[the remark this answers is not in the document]
wnbts: i mean if model is null, why not set a new model?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
wnbts: number of strides should be the number of samples, no greater than the max, which is then used in line 475. it can be calculated as floor(bucket/stride).

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
wnbts: for count/sum/etc on missing data, the value of the bucket should be 0. A model trained with 0s from missing data will treat new non-zero data points as anomalies.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
wnbts: isn't line 390/440 coldStartData.add(featureManager.batchShingle(points, entityShingleSize)) already producing shingled data?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
wnbts: the main issue is incorrect training data. When there is 2 data points for an entity in the index, they will be added the training data on the first run. On second run later, the same data points will be added again. And again and again until a number of samples has reached but all samples are just repeats.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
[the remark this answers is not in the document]
kaituo: good idea.  Changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: right.  I ended up with "int numStrides = Math.min((int) Math.floor(numBuckets / (double) stride), maxTrainSamples);"

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: yeah, we should have some aggregation to differentiate the two cases: count with default 0 and count with missing value remove.d

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: yes, the entity state's samples are not shignled data.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/EntityColdStarter.java
kaituo: I have a bloom filter to filter such cases.  If a point only appears once or twice, it won't trigger cold start.

wnbts (approved):
there are three main issues.
1. feature query in cold start should not return 0 for count/sum/additive aggregation on missing date. The changes will be in a separate pr.
2. cold start process might result into overlapping/duplicate samples. The issue is noted.
3. shingling support is not implemented. The issue is noted.

kaituo:
> there are three main issues.
> 
>     1. feature query in cold start should not return 0 for count/sum/additive aggregation on missing date. The changes will be in a separate pr.
> 
>     2. cold start process might result into overlapping/duplicate samples. The issue is noted.
> 
>     3. shingling support is not implemented. The issue is noted.

Thanks Lai.  Will put those things in to-do list.

yizheliu-amazon (approved):

PR #248, opened 2020-10-13
Change CD workflow to use new staging bucket for artifacts

camerski:
PLEASE DO NOT MERGE THIS REQUEST UNTIL ASKED. We need to coordinate the merge with updating Github secrets.

*Description of changes:*
The infrastructure team is separating the production and staging locations into different AWS accounts. Plugins need to modify their workflows to publish to the new locations.

This PR changes the CD workflow to add a build number and write the zip, deb, and rpm plugin artifacts to staging.artifacts.opendistroforelasticsearch.amazon.com. The write to S3 currently fails because the secrets have not been updated; the secrets will be updated at the same time this PR is merged.

The workflow currently fails due to an unrelated JaCoCo issue: https://github.com/camerski/anomaly-detection/actions/runs/305267309

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -32,11 +32,26 @@ jobs:
       - name: Run build
         run: |
           ./gradlew buildPackages --console=plain -Dbuild.snapshot=false
-          artifact=`ls build/distributions/*.zip`
-          rpm_artifact=`ls build/distributions/*.rpm`
-          deb_artifact=`ls build/distributions/*.deb`
 
-          aws s3 cp $artifact s3://artifacts.opendistroforelasticsearch.amazon.com/downloads/elasticsearch-plugins/opendistro-anomaly-detection/
-          aws s3 cp $rpm_artifact s3://artifacts.opendistroforelasticsearch.amazon.com/downloads/rpms/opendistro-anomaly-detection/
-          aws s3 cp $deb_artifact s3://artifacts.opendistroforelasticsearch.amazon.com/downloads/debs/opendistro-anomaly-detection/
-          aws cloudfront create-invalidation --distribution-id E1VG5HMIWI4SA2 --paths "/downloads/*"
+      - name: Upload to S3
+        shell: bash
+        run: |
+          zip=`ls build/distributions/*.zip`
+          rpm=`ls build/distributions/*.rpm`
+          deb=`ls build/distributions/*.deb`
+
+          # Inject the build number before the suffix
+          zip_outfile=`basename ${zip%.zip}-build-${GITHUB_RUN_NUMBER}.zip`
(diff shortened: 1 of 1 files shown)


kaituo (approved):

saratvemulapalli (approved):

.github/workflows/CD.yml
```diff
@@ -32,11 +32,26 @@ jobs:
...
+          zip=`ls build/distributions/*.zip`
+          rpm=`ls build/distributions/*.rpm`
+          deb=`ls build/distributions/*.deb`
+
+          # Inject the build number before the suffix
+          zip_outfile=`basename ${zip%.zip}-build-${GITHUB_RUN_NUMBER}.zip`
+          rpm_outfile=`basename ${rpm%.rpm}-build-${GITHUB_RUN_NUMBER}.rpm`
+          deb_outfile=`basename ${deb%.deb}-build-${GITHUB_RUN_NUMBER}.deb`
+
+          s3_prefix="s3://staging.artifacts.opendistroforelasticsearch.amazon.com/snapshot/elasticsearch-plugins/anomaly-detection/"
```
ylwu-amzn, line 48: So all plugins are uploading artifact to staging only in CD workflow ? Don't see production location in this file. Where will the production location be used ?

.github/workflows/CD.yml
```diff
@@ -32,11 +32,26 @@ jobs:
...
+
+      - name: Upload to S3
+        shell: bash
+        run: |
+          zip=`ls build/distributions/*.zip`
+          rpm=`ls build/distributions/*.rpm`
+          deb=`ls build/distributions/*.deb`
+
+          # Inject the build number before the suffix
+          zip_outfile=`basename ${zip%.zip}-build-${GITHUB_RUN_NUMBER}.zip`
```
ylwu-amzn, line 44: GITHUB_RUN_NUMBER is unique number for each run, so the artifact name will be unique for each run, while the original artifact name keep the same for same version. Is this change compatible with current CD workflow?

.github/workflows/CD.yml
camerski: The infrastructure team will promote artifacts to production as part of the ODFE release process. For out-of-cycle releases, you can promote the artifacts to production using your own workflows, or the infrastructure team can help.

.github/workflows/CD.yml
camerski: We are making corresponding changes to how we generate ODFE distributions to handle the change to the naming convention. If you need to do an out-of-cycle release then there may be manual work required; the infrastructure team can help with that.

gaiksaya:
More changes were required but unable to access the specified branch. Hence closing this PR and raising a new one https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/358

PR #249, opened 2020-10-14
Add transport action for model inference

kaituo:

*Issue #, if available:*

*Description of changes:*
The coordinating node computes feature vectors for multiple entities within a time interval through a query, finds model host nodes using consistent hashing, and asks model hosting nodes to infer anomaly grade and confidence of the feature vector.

This PR adds transport actions for model hosting nodes to infer the feature vector's anomaly grade and confidence. Model hosting nodes proceed entity features in a loop.  We first check if the entity’s models are in memory. If yes, we use the model to infer anomaly grade and save them.  If no, we skip the entity.

Testing done:
* add unit tests to cover the newly added transport actions
* end-to-end testing

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultAction.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.transport;
+
+import org.elasticsearch.action.ActionType;
+import org.elasticsearch.action.support.master.AcknowledgedResponse;
+
+public class EntityResultAction extends ActionType<AcknowledgedResponse> {
+    public static final EntityResultAction INSTANCE = new EntityResultAction();
+    public static final String NAME = "cluster:admin/ad/entity/result";
+
+    private EntityResultAction() {
+        super(NAME, AcknowledgedResponse::new);
+    }
+
+}
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultAction.java
@@ -20,7 +20,7 @@ import org.elasticsearch.action.support.master.AcknowledgedResponse;
 
 public class EntityResultAction extends ActionType<AcknowledgedResponse> {
     public static final EntityResultAction INSTANCE = new EntityResultAction();
-    public static final String NAME = "cluster:admin/ad/entity/result";
+    public static final String NAME = "cluster:admin/opendistro/ad/entity/result";
 
     private EntityResultAction() {
         super(NAME, AcknowledgedResponse::new);
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.transport;
+
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.COOLDOWN_MINUTES;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.Map.Entry;
+import java.util.Optional;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.support.ActionFilters;
+import org.elasticsearch.action.support.HandledTransportAction;
+import org.elasticsearch.action.support.master.AcknowledgedResponse;
+import org.elasticsearch.common.inject.Inject;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.tasks.Task;
+import org.elasticsearch.transport.TransportService;
+
+import com.amazon.opendistroforelasticsearch.ad.NodeStateManager;
+import com.amazon.opendistroforelasticsearch.ad.breaker.ADCircuitBreakerService;
+import com.amazon.opendistroforelasticsearch.ad.caching.CacheProvider;
+import com.amazon.opendistroforelasticsearch.ad.caching.EntityCache;
+import com.amazon.opendistroforelasticsearch.ad.common.exception.EndRunException;
+import com.amazon.opendistroforelasticsearch.ad.common.exception.LimitExceededException;
+import com.amazon.opendistroforelasticsearch.ad.constant.CommonErrorMessages;
+import com.amazon.opendistroforelasticsearch.ad.ml.CheckpointDao;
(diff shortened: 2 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -0,0 +1,181 @@
...
+            // index pressure is high. Only save anomalies
+            boolean onlySaveAnomalies = stateManager
+                .getLastIndexThrottledTime()
+                .plus(Duration.ofMinutes(coolDownMinutes))
+                .isAfter(clock.instant());
+
+            for (Entry<String, double[]> entity : request.getEntities().entrySet()) {
+                String entityName = entity.getKey();
+                // For ES, the limit of the document ID is 512 bytes.
+                // truncate to 256 characters if too long since we are using it as part of document id.
```
ylwu-amzn, line 135: This comment can't match the code logic below. line137 will skip entity if its name is too long.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -0,0 +1,181 @@
...
+                                null,
+                                Arrays.asList(new Entity(categoricalField, entityName))
+                            ),
+                            currentBulkRequest
+                        );
+                }
+            }
+            this.anomalyResultHandler.flush(currentBulkRequest, detectorId);
+            // bulk all accumulated checkpoint requests
+            this.checkpointDao.flush();
```
ylwu-amzn, line 173: How about add `listener` as param of `flush` function ? If flush successfully, execute `listener.onResponse(new AcknowledgedResponse(true));` , if fail, execute `listener.onFailure`. Otherwise user never know the flush succeed or not.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -0,0 +1,181 @@
...
+                ModelState<EntityModel> entityModel = cache.get(modelId, detector, datapoint, entityName);
+                if (entityModel == null) {
+                    // cache miss
+                    continue;
+                }
+                ThresholdingResult result = manager.getAnomalyResultForEntity(detectorId, datapoint, entityName, entityModel, modelId);
+                // result.getRcfScore() = 0 means the model is not initialized
+                // result.getGrade() = 0 means it is not an anomaly
+                // So many EsRejectedExecutionException if we write no matter what
+                if (result.getRcfScore() > 0 && (!onlySaveAnomalies || result.getGrade() > 0)) {
```
ylwu-amzn, line 151: If model not initialized (`rcfScore == 0`) for a long time such as not enough data/some error, how does user know what's going on if we don't write AD result ? The init progress bar is only for most active entity.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -0,0 +1,181 @@
...
+                        );
+                }
+            }
+            this.anomalyResultHandler.flush(currentBulkRequest, detectorId);
+            // bulk all accumulated checkpoint requests
+            this.checkpointDao.flush();
+
+            listener.onResponse(new AcknowledgedResponse(true));
+        }, exception -> {
+            LOG.error("fail to get entity's anomaly grade", exception);
```
ylwu-amzn, line 177: Suggest to add detector id and start/end in error message for easier log analysis.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
```diff
@@ -0,0 +1,181 @@
...
+                listener.onFailure(new EndRunException(detectorId, "AnomalyDetector is not available.", true));
+                return;
+            }
+
+            AnomalyDetector detector = detectorOptional.get();
+            // we only support 1 categorical field now
+            String categoricalField = detector.getCategoryField().get(0);
+
+            ADResultBulkRequest currentBulkRequest = new ADResultBulkRequest();
+            // index pressure is high. Only save anomalies
```
saratvemulapalli, line 126: Just curious, how did we determine the index pressure is high?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: good catch.  fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: I didn't do it because:
First, flush is asynchronous.  We don't know when it is gonna finish.  In our performance testing, we find our queue has 70 k checkpoints.  Hold the job too long may fail the following jobs.
Second, I don't know what's the action item for customers if they know their checkpoints fail.  Mostly they are agnostic to checkpoints.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: If the entity is in cache, profile can call getTotalUpdates(String detectorId, String entityId).

If not, profile API has to go to a checkpoint, load it to memory, and check its total updates.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultAction.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.transport;
+
+import org.elasticsearch.action.ActionType;
+import org.elasticsearch.action.support.master.AcknowledgedResponse;
+
+public class EntityResultAction extends ActionType<AcknowledgedResponse> {
+    public static final EntityResultAction INSTANCE = new EntityResultAction();
+    public static final String NAME = "cluster:admin/ad/entity/result";
```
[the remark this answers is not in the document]
kaituo: changed

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: We get exceptions from ES if index pressure is high.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
ylwu-amzn: How do we handle the long document id ? Will ES throw exception if document id exceeds length limitation?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
ylwu-amzn: > We don't know when it is gonna finish. In our performance testing, we find our queue has 70 k checkpoints. Hold the job too long may fail the following jobs.
Make sense. Do we catch exception in `checkpointDao ` and write in AD result ? 

> what's the action item for customers if they know their checkpoints fail
User know the system is under pressure, they can 
1.Scale up/out their cluster
2.Stop some testing/low priority detector
3.Tune detector configuration like use less features, tune detector interval etc

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
ylwu-amzn: Cache/checkpoint will be cleared periodically. Have some concern for Ops, if user want to know what happens, we can only rely on service log. We can address this in next phase, it's ok if we have enough logs for now.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: not sure.  The doc said it can only handle 512 bytes at most.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: We don't in checkpointDao.  We write to logs.  Write to state index might be an option.  We can have a field called checkpointError.  How's that?

The action item you mentioned great.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
kaituo: agreed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/EntityResultTransportAction.java
ylwu-amzn: I think that's ok to track in state index.

kaituo: will do it after release.

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #250, opened 2020-10-14
Get feature data

kaituo:

*Issue #, if available:*

*Description of changes:*
To get feature data, the coordinating node regularly aggregates log entries into multiple entity keys and their corresponding value vectors.  An entity key can be an IP address, while the value vector can contain the total bytes sent to the IP address over an interval (say, 1 minute).

In this PR, We use terms aggregation (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html) to calculate features.  There can be a significant number (e.g., millions) of entities in one query. By defaut, we return 1000 entities per query at most. Customers have the option to increase the limit.

This PR also adds functions to get cold start samples within a range.

Testing done:
1. added unit tests.
2. done end-to-end testing

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
@@ -15,13 +15,16 @@
 
 package com.amazon.opendistroforelasticsearch.ad.feature;
 
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.MAX_ENTITIES_PER_QUERY;
 import static org.apache.commons.math3.linear.MatrixUtils.createRealMatrix;
 
 import java.io.IOException;
 import java.util.AbstractMap.SimpleEntry;
+import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
```diff
@@ -569,4 +649,139 @@ private SearchRequest createPreviewSearchRequest(AnomalyDetector detector, List<
...
+        AnomalyDetector detector,
+        List<Entry<Long, Long>> ranges,
+        String entityName,
+        ActionListener<List<Optional<double[]>>> listener
+    ) throws IOException {
+        SearchRequest request = createColdStartFeatureSearchRequest(detector, ranges, entityName);
+        logger.debug(() -> "getColdStartSamplesForPeriods: " + request.toString());
+
+        client.search(request, ActionListener.wrap(response -> {
+            logger.debug(() -> "getColdStartSamplesForPeriods: " + response.toString());
```
ohltyler, line 663: small: remove these debugging logs?

ohltyler (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/SearchFeatureDao.java
kaituo: removed

yizheliu-amazon (approved):

Issue #195, opened 2020-07-17
Anomaly Detection integration with Security : FGAC Support


kaituo:
**Is your feature request related to a problem? Please describe.**
We need to integrate with security plugin to allow fine-grained access where AD job can assume detector definer role and non-admin can only see detectors they have defined.

PR #251, opened 2020-10-14
Adding User support for Detector and DetectorJob

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Adding User support to store User information for each detector created and store user context in .opendistro-anomaly-detectors index.
Adding User support for AD Job and store user context in .opendistro-anomaly-detectors-jobs.

User role will be used to filter when AD Rest API's are invoked. 
Background jobs will soon have injected role from the User context. 

User information is always sent to elasticsearch when Opendistro security plugin is enabled.


ohltyler (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -289,6 +308,9 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         if (categoryFields != null) {
             xContentBuilder.field(CATEGORY_FIELD, categoryFields.toArray());
         }
+        if (user != null) {
```
ylwu-amzn, line 311: Add `user` field in detector index mapping https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/resources/mappings/anomaly-detectors.json ?
Same to detector job.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -131,7 +135,8 @@ public AnomalyDetector(
         Map<String, Object> uiMetadata,
         Integer schemaVersion,
         Instant lastUpdateTime,
-        List<String> categoryFields
+        List<String> categoryFields,
+        User user
```
ylwu-amzn, line 139: Here we just add `User` in each detector document. Security plugin will do authorization check for each request based on `backend_roles` or `roles` ? Can you share the related code?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -260,6 +273,12 @@ public void writeTo(StreamOutput output) throws IOException {
         output.writeInt(schemaVersion);
         output.writeInstant(lastUpdateTime);
         output.writeStringCollection(categoryFields);
+        if (user != null) {
+            output.writeBoolean(true); // user exists
+            user.writeTo(output);
+        } else {
+            output.writeBoolean(false); // user does not exist
```
ylwu-amzn, line 280: For old detector, they have no user field, so they will be open to all users which has AD permission ?
For new detector, we will create default user with the creator's user role?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
saratvemulapalli: Great point sure. Will add.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
saratvemulapalli: Good question, security plugin will only provide the authorization headers.
My next PR will address this. Basically we have to query and filter by `roles` or `backend_roles`.
Here is a code pointer in alerting which does that: https://github.com/opendistro-for-elasticsearch/alerting/commit/44abca11b49a658fc22b25ab72c4ab62bd9057f3#diff-39bec5249e2b91e055cdb30bb3edcd28ee868ab3676e98652547bcce28eed895R86

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
saratvemulapalli: Exactly, for existing detectors where User doesn't exist, the filter will not work and users who have all access to rest api's will be able to see those detectors. 
This is to maintain backward compatibility.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
ylwu-amzn: Got it, that makes sense.

ylwu-amzn (approved):
LGTM. Thanks for the change!

skkosuri-amzn (commented):
Looks good. 👍

PR #253, opened 2020-10-14
Add caching

kaituo:

*Issue #, if available:*

*Description of changes:*

In a single-entity detector, the amount of memory used is proportional to the number of features and is fixed when the detector is defined. However, with the introduction of multi-entity detectors, the number of entities is not fixed, and active entities are changing. Thus, the size of the required memory may constantly change in the lifetime of a detector. Caching can accommodate such requirements without the need to pre-allocate memory for a detector in a fixed number. If the memory is enough, we create models for all entities and monitor anomalies. Otherwise, we cache hot entities in memory and make a best-effort attempt to admit other entities.

This PR adds caching for multi-entity detectors.  We have a three-level cache that stores entity states in each node. The first two levels are for active entities. The third level is for inactive entities. Each detector has its dedicated cache that stores ten (dynamically adjustable by humans) entities’ states per node. A detector’s hottest entities load their states in the dedicated cache. If more than 10 entities use the dedicated cache, the secondary cache can use the rest of the free memory available to AD. The secondary cache is a shared memory among all detectors for the long tail. The shared cache size is 10% heap minus all of the dedicated cache consumed by single-entity and multi-entity detectors. The shared cache’s size shrinks as the dedicated cache is filled up or more detectors are started.

Testing done:
1. added unit tests.
2. end-to-end testing

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
@@ -31,6 +31,7 @@ import org.apache.commons.lang.builder.ToStringBuilder;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import com.amazon.opendistroforelasticsearch.ad.ExpiringState;
 import com.amazon.opendistroforelasticsearch.ad.MaintenanceState;
 import com.amazon.opendistroforelasticsearch.ad.MemoryTracker;
 import com.amazon.opendistroforelasticsearch.ad.MemoryTracker.Origin;
@@ -39,7 +40,6 @@ import com.amazon.opendistroforelasticsearch.ad.ml.CheckpointDao;
 import com.amazon.opendistroforelasticsearch.ad.ml.EntityModel;
 import com.amazon.opendistroforelasticsearch.ad.ml.ModelState;
 import com.amazon.opendistroforelasticsearch.ad.model.InitProgressProfile;
-import com.amazon.opendistroforelasticsearch.ad.util.ExpiringState;
 
 /**
  * We use a layered cache to manage active entities’ states.  We have a two-level
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
@@ -20,6 +20,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.Comparator;
+import java.util.Iterator;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -104,11 +105,18 @@ public class CacheBuffer implements ExpiringState, MaintenanceState {
 
         @Override
         public int compare(PriorityNode priority, PriorityNode priority2) {
-            int cmp = Float.compare(priority.priority, priority2.priority);
-            if (cmp == 0) {
-                cmp = priority.key.compareTo(priority2.key);
-            }
-            return cmp;
+          int equality = priority.key.compareTo(priority2.key);
+          if (equality == 0) {
+              // this is consistent with PriorityNode's equals method
+              return 0;
+          }
+          // if not equal, first check priority
+          int cmp = Float.compare(priority.priority, priority2.priority);
+          if (cmp == 0) {
+              // if priority is equal, use lexicographical order of key
+              cmp = equality;
+          }
+          return cmp;
         }
     }
 
@@ -302,7 +310,7 @@ public class CacheBuffer implements ExpiringState, MaintenanceState {
         // The removed one loses references and soon GC will collect it.
         // We have memory tracking correction to fix incorrect memory usage record.
         // put: not a problem as it is unlikely we are removing and putting the same thing
-        PriorityNode smallest = priorityList.pollFirst();
+        PriorityNode smallest = priorityList.first();
         if (smallest != null) {
             remove(smallest.key);
         }
@@ -316,6 +324,8 @@ public class CacheBuffer implements ExpiringState, MaintenanceState {
      * is no associated ModelState for the key
      */
     public ModelState<EntityModel> remove(String keyToRemove) {
+        // remove if the key matches; priority does not matter
+        priorityList.remove(new PriorityNode(keyToRemove, 0));
         // if shared cache is empty, we are using reserved memory
         boolean reserved = sharedCacheEmpty();
 
@@ -405,7 +415,6 @@ public class CacheBuffer implements ExpiringState, MaintenanceState {
                     // We have memory tracking correction to fix incorrect memory usage record.
                     // put: not a problem as we are unlikely to maintain an entry that's not
                     // already in the cache
-                    priorityList.remove(new PriorityNode(entityId, modelState.getPriority()));
                     remove(entityId);
                 }
             } catch (Exception e) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
@@ -1,0 +1,515 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.caching;
+
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.Comparator;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListSet;
+
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.apache.commons.lang.builder.ToStringBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import com.amazon.opendistroforelasticsearch.ad.MaintenanceState;
+import com.amazon.opendistroforelasticsearch.ad.MemoryTracker;
+import com.amazon.opendistroforelasticsearch.ad.MemoryTracker.Origin;
+import com.amazon.opendistroforelasticsearch.ad.annotation.Generated;
+import com.amazon.opendistroforelasticsearch.ad.ml.CheckpointDao;
+import com.amazon.opendistroforelasticsearch.ad.ml.EntityModel;
+import com.amazon.opendistroforelasticsearch.ad.ml.ModelState;
+import com.amazon.opendistroforelasticsearch.ad.model.InitProgressProfile;
+import com.amazon.opendistroforelasticsearch.ad.util.ExpiringState;
+
+/**
+ * We use a layered cache to manage active entities’ states.  We have a two-level
+ * cache that stores active entity states in each node.  Each detector has its
+ * dedicated cache that stores ten (dynamically adjustable) entities’ states per
+ * node.  A detector’s hottest entities load their states in the dedicated cache.
+ * If less than 10 entities use the dedicated cache, the secondary cache can use
+ * the rest of the free memory available to AD.  The secondary cache is a shared
+ * memory among all detectors for the long tail.  The shared cache size is 10%
+ * heap minus all of the dedicated cache consumed by single-entity and multi-entity
+ * detectors.  The shared cache’s size shrinks as the dedicated cache is filled
+ * up or more detectors are started.
+ */
+public class CacheBuffer implements ExpiringState, MaintenanceState {
+    private static final Logger LOG = LogManager.getLogger(CacheBuffer.class);
+
+    static class PriorityNode {
+        private String key;
+        private float priority;
+
+        PriorityNode(String key, float priority) {
+            this.priority = priority;
+            this.key = key;
+        }
+
+        @Generated
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj)
+                return true;
+            if (obj == null)
+                return false;
+            if (getClass() != obj.getClass())
+                return false;
+            if (obj instanceof PriorityNode) {
+                PriorityNode other = (PriorityNode) obj;
+
+                EqualsBuilder equalsBuilder = new EqualsBuilder();
+                equalsBuilder.append(key, other.key);
+                return equalsBuilder.isEquals();
+            }
+            return false;
+        }
+
+        @Generated
+        @Override
+        public int hashCode() {
+            return new HashCodeBuilder().append(key).toHashCode();
+        }
+
+        @Generated
+        @Override
+        public String toString() {
+            ToStringBuilder builder = new ToStringBuilder(this);
+            builder.append("key", key);
+            builder.append("priority", priority);
+            return builder.toString();
+        }
+    }
+
+    static class PriorityNodeComparator implements Comparator<PriorityNode> {
+
+        @Override
+        public int compare(PriorityNode priority, PriorityNode priority2) {
+            int cmp = Float.compare(priority.priority, priority2.priority);
+            if (cmp == 0) {
+                cmp = priority.key.compareTo(priority2.key);
+            }
+            return cmp;
+        }
+    }
+
+    private final int minimumCapacity;
+    // key -> Priority node
+    private final ConcurrentHashMap<String, PriorityNode> key2Priority;
+    private final ConcurrentSkipListSet<PriorityNode> priorityList;
+    // key -> value
+    private final ConcurrentHashMap<String, ModelState<EntityModel>> items;
+    // when detector is created.  Can be reset.  Unit: seconds
+    private long landmarkSecs;
+    // length of seconds in one interval.  Used to compute elapsed periods
+    // since the detector has been enabled.
+    private long intervalSecs;
+    // memory consumption per entity
+    private final long memoryConsumptionPerEntity;
+    private final MemoryTracker memoryTracker;
+    private final Clock clock;
+    private final CheckpointDao checkpointDao;
+    private final Duration modelTtl;
+    private final String detectorId;
+    private Instant lastUsedTime;
+    private final int decayConstant;
+    private final long reservedBytes;
+
+    public CacheBuffer(
+        int minimumCapacity,
+        long intervalSecs,
+        CheckpointDao checkpointDao,
+        long memoryConsumptionPerEntity,
+        MemoryTracker memoryTracker,
+        Clock clock,
+        Duration modelTtl,
+        String detectorId
+    ) {
+        this.minimumCapacity = minimumCapacity;
+        this.key2Priority = new ConcurrentHashMap<>();
+        this.priorityList = new ConcurrentSkipListSet<>(new PriorityNodeComparator());
+        this.items = new ConcurrentHashMap<>();
+        this.landmarkSecs = clock.instant().getEpochSecond();
+        this.intervalSecs = intervalSecs;
+        this.memoryConsumptionPerEntity = memoryConsumptionPerEntity;
+        this.memoryTracker = memoryTracker;
+        this.clock = clock;
+        this.checkpointDao = checkpointDao;
+        this.modelTtl = modelTtl;
+        this.detectorId = detectorId;
+        this.lastUsedTime = clock.instant();
+        this.decayConstant = 3;
+        this.reservedBytes = memoryConsumptionPerEntity * minimumCapacity;
+    }
+
+    /**
+     * Update step at period t_k:
+     * new priority = old priority + log(1+e^{\log(g(t_k-L))-old priority}) where g(n) = e^{0.125n},
+     * and n is the period.
+     * @param entityId model Id
+     */
+    private void update(String entityId) {
+        PriorityNode node = key2Priority.computeIfAbsent(entityId, k -> new PriorityNode(entityId, 0f));
+        // reposition this node
+        this.priorityList.remove(node);
+        node.priority = getUpdatedPriority(node.priority);
+        this.priorityList.add(node);
+
+        Instant now = clock.instant();
+        items.get(entityId).setLastUsedTime(now);
+        lastUsedTime = now;
+    }
+
+    public float getUpdatedPriority(float oldPriority) {
+        long increment = computeWeightedCountIncrement();
+        // if overflowed, we take the short cut from now on
+
+        oldPriority += Math.log(1 + Math.exp(increment - oldPriority));
+        // if overflow happens, using \log(g(t_k-L)) instead.
+        if (oldPriority == Float.POSITIVE_INFINITY) {
+            oldPriority = increment;
+        }
+        return oldPriority;
+    }
+
+    /**
+     * Compute periods relative to landmark and the weighted count increment using 0.125n.
+     * Multiply by 0.125 is implemented using right shift for efficiency.
+     * @return the weighted count increment used in the priority update step.
+     */
+    private long computeWeightedCountIncrement() {
+        long periods = (clock.instant().getEpochSecond() - landmarkSecs) / intervalSecs;
+        return periods >> decayConstant;
+    }
+
+    /**
+     * Compute the weighted total count by considering landmark
+     * \log(C)=\log(\sum_{i=1}^{n} (g(t_i-L)/g(t-L)))=\log(\sum_{i=1}^{n} (g(t_i-L))-\log(g(t-L))
+     * @return the minimum priority entity's ID and priority
+     */
+    public Entry<String, Float> getMinimumPriority() {
+        PriorityNode smallest = priorityList.first();
+        long periods = (clock.instant().getEpochSecond() - landmarkSecs) / intervalSecs;
+        float detectorWeight = periods >> decayConstant;
+        return new SimpleImmutableEntry<>(smallest.key, smallest.priority - detectorWeight);
+    }
+
+    /**
+     * Insert the model state associated with a model Id to the cache
+     * @param entityId the model Id
+     * @param value the ModelState
+     */
+    public void put(String entityId, ModelState<EntityModel> value) {
+        // race conditions can happen between the put and one of the following operations:
+        // remove: not a problem as it is unlikely we are removing and putting the same thing
+        // maintenance: not a problem as we are unlikely to maintain an entry that's not
+        // already in the cache
+        // clear: not a problem as we are releasing memory in MemoryTracker.
+        // The newly added one loses references and soon GC will collect it.
+        // We have memory tracking correction to fix incorrect memory usage record.
+        // put from other threads: not a problem as the entry is associated with
+        // entityId and our put is idempotent
+        put(entityId, value, value.getPriority());
+    }
+
+    /**
+    * Insert the model state associated with a model Id to the cache.  Update priority.
+    * @param entityId the model Id
+    * @param value the ModelState
+    * @param priority the priority
+    */
+    private void put(String entityId, ModelState<EntityModel> value, float priority) {
+        if (minimumCapacity <= 0) {
+            return;
+        }
+        ModelState<EntityModel> contentNode = items.get(entityId);
+        if (contentNode == null) {
+            PriorityNode node = new PriorityNode(entityId, priority);
+            key2Priority.put(entityId, node);
+            priorityList.add(node);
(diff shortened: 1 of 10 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,515 @@
...
+/**
+ * We use a layered cache to manage active entities’ states.  We have a two-level
+ * cache that stores active entity states in each node.  Each detector has its
+ * dedicated cache that stores ten (dynamically adjustable) entities’ states per
+ * node.  A detector’s hottest entities load their states in the dedicated cache.
+ * If less than 10 entities use the dedicated cache, the secondary cache can use
+ * the rest of the free memory available to AD.  The secondary cache is a shared
+ * memory among all detectors for the long tail.  The shared cache size is 10%
+ * heap minus all of the dedicated cache consumed by single-entity and multi-entity
+ * detectors.  The shared cache’s size shrinks as the dedicated cache is filled
```
ylwu-amzn, line 53: So we have two kinds of cache: dedicated cache and shared cache. Where are dedicated cache and shared cache stored ? Both stored in `items` and `minimumCapacity` controls the boundary ? If `items` size is equals to or less than `minimumCapacity`, consider `items` as dedicated cache;  otherwise, consider top `minimumCapacity` active entities(last X entities in `priorityList`) as in dedicated cache and all others in shared cache?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheProvider.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.caching;
+
+import com.amazon.opendistroforelasticsearch.ad.ml.EntityModel;
+import com.amazon.opendistroforelasticsearch.ad.ml.ModelState;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+
+/**
+ * A wrapper to call concrete implementation of caching.  Used in transport
```
ylwu-amzn, line 23: Adding `CacheProvider` just for isolating concrete implementation of caching ? The `EntityCache` interface already works for that goal. We define `EntityCache` as class attribute or method param, then can pass any concrete implementation of `EntityCache`.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: right.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheProvider.java
kaituo: Yes.

I don't know how to make it work for transport layer.  Do yout know how to bind an interface with an implementation in AnomalyDetectorPlugin?

LiuJoyceC:
My review and approval of this PR is based on the implementation of the files:
- https://github.com/kaituo/anomaly-detection/blob/test_hc3/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java. 
- https://github.com/kaituo/anomaly-detection/blob/test_hc3/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java

I checked for the correctness of the update step computations to get the priority score and the correctness of comparisons of priority (both within the detector and across detectors) to determine which entities to eject from dedicated and shared caches. These are correct, so I approve this PR.

However, I do recommend adding more comments, because the calculations are not transparent, and it's not readily apparent that the implemented code is equivalent mathematically to the calculations in the proposal doc without doing some pen-and-paper derivations to verify that they are equivalent. This could make it difficult for contributors in the future to work on this logic, since it is not easily understandable.

LiuJoyceC (approved):
My review and approval of this PR is based on the implementation of the files:
- https://github.com/kaituo/anomaly-detection/blob/test_hc3/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java. 
- https://github.com/kaituo/anomaly-detection/blob/test_hc3/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java

I checked for the correctness of the update step computations to get the priority score and the correctness of comparisons of priority (both within the detector and across detectors) to determine which entities to eject from dedicated and shared caches. These are correct, so I approve this PR.

However, I do recommend adding more comments, because the calculations are not transparent, and it's not readily apparent that the implemented code is equivalent mathematically to the calculations in the proposal doc without doing some pen-and-paper derivations to verify that they are equivalent. This could make it difficult for contributors in the future to work on this logic, since it is not easily understandable.

kaituo:
> My review and approval of this PR is based on the implementation of the files:
> 
>     * https://github.com/kaituo/anomaly-detection/blob/test_hc3/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java.
> 
>     * https://github.com/kaituo/anomaly-detection/blob/test_hc3/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
> 
> 
> I checked for the correctness of the update step computations to get the priority score and the correctness of comparisons of priority (both within the detector and across detectors) to determine which entities to eject from dedicated and shared caches. These are correct, so I approve this PR.
> 
> However, I do recommend adding more comments in the code (or adding derivations in the design doc), because the calculations are not transparent, and it's not readily apparent that the implemented code is equivalent mathematically to the calculations in the exponential forward decay proposal in the design doc without doing some pen-and-paper derivations to verify that they are equivalent. This could make it difficult for contributors in the future to work on this logic, since it is not easily understandable.

Thanks so much Joyce.  Without you, the formua is not possible.  Will improve documentation after release.  Yeah, it would be hard for a person who has no context.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheProvider.java
ylwu-amzn: Got it. How about provide interface directly in this provider ? If override `EntityCache`, we need to update this provider for `EntityCache` change like add new method, update current method. Sample code

```
import org.elasticsearch.common.inject.Provider;

public class CacheProvider implements Provider<EntityCache> {
    private EntityCache cache;
    public CacheProvider(EntityCache cache) {
        this.cache = cache;
    }

    @Override public EntityCache get() {
        return cache;
    }
}

```
So client just need to use `CacheProvider.get()` to get concrete `EntityCache` instance, no need to worry the implementation and we don't need to update `CacheProvider` for `EntityCache` changes.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+        this.landmarkSecs = clock.instant().getEpochSecond();
+        this.intervalSecs = intervalSecs;
+        this.memoryConsumptionPerEntity = memoryConsumptionPerEntity;
+        this.memoryTracker = memoryTracker;
+        this.clock = clock;
+        this.checkpointDao = checkpointDao;
+        this.modelTtl = modelTtl;
+        this.detectorId = detectorId;
+        this.lastUsedTime = clock.instant();
+        this.decayConstant = 3;
```
ylwu-amzn, line 163: How about make this constant ?
```
private final int DECAY_CONSTANT = 3;
```

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+     */
+    public ModelState<EntityModel> get(String key) {
+        // We can get an item that is to be removed soon due to race condition.
+        // This is acceptable as it won't cause any corruption and exception.
+        // And this item is used for scoring one last time.
+        ModelState<EntityModel> node = items.get(key);
+        if (node == null) {
+            return null;
+        }
+        update(key);
```
ylwu-amzn, line 279: Why update for both `put` (line259) and `get` ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+        put(entityId, value, value.getPriority());
+    }
+
+    /**
+    * Insert the model state associated with a model Id to the cache.  Update priority.
+    * @param entityId the model Id
+    * @param value the ModelState
+    * @param priority the priority
+    */
+    private void put(String entityId, ModelState<EntityModel> value, float priority) {
```
ylwu-amzn, line 242: I think the `entityId` here means `entityModelId`. We share `CacheBuffer` for all detectors, will conflict if two detectors have same entity id. 
How about we rename it as `entityModelId` if my understanding correct ? I thought every detector has its own `CacheBuffer` when I take first look at this method.

ylwu-amzn: Sorry for misunderstanding. Find each detector has its own dedicated cache buffer when I read priority cache.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheProvider.java
kaituo: yeah, changed

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: It is a constant.  "private final int DECAY_CONSTANT;"  I changed the name.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: Frequency changes as long as there is an access.  Both get and put are accesses.

ylwu-amzn: I think it's ok as we have no retry currently. If we add retry, seems not reasonable to update frequency for `get` in retry.

kaituo: We don't have retry currently.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+    /**
+     * Clear associated memory.  Used when we are removing an detector.
+     */
+    public void clear() {
+        // race conditions can happen between the put and remove/maintenance/put:
+        // not a problem as we are releasing memory in MemoryTracker.
+        // The newly added one loses references and soon GC will collect it.
+        // We have memory tracking correction to fix incorrect memory usage record.
+
+        memoryTracker.releaseMemory(getReservedBytes(), true, Origin.MULTI_ENTITY_DETECTOR);
```
ylwu-amzn, line 460: Memory tracker only record how much memory used totally ? Where to remove current cache ?
`items.clear()`, `key2Priority.clear()` and `priorityList.clear()`

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+        put(entityId, value, value.getPriority());
+    }
+
+    /**
+    * Insert the model state associated with a model Id to the cache.  Update priority.
+    * @param entityId the model Id
+    * @param value the ModelState
+    * @param priority the priority
+    */
+    private void put(String entityId, ModelState<EntityModel> value, float priority) {
```
ylwu-amzn, line 242: `entityId` should be `entityModelId` ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: it records all memory usage by AD.

This method is for the caller to remove the CacheBuffer and clear up the memory.  The first part is done by the caller, while the 2nd part is done by the cachebuffer itself.  After removing the CacheBuffer's reference, GC should collect its memory.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: for CacheBuffer, I use entity id throughout the code to represent some id to use as key.  We have detector id, entity name, and entity id = model id.

kaituo: for example, one of my comment:

/**
    * Insert the model state associated with a model Id to the cache.  Update priority.
    * @param entityId the model Id
    * @param value the ModelState
    * @param priority the priority
    */

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+        return Optional.of(priorityList).map(list -> list.last()).map(node -> node.key);
+    }
+
+    /**
+     *
+     * @param entityId entity Id
+     * @return Get the model of an entity
+     */
+    public Optional<EntityModel> getModel(String entityId) {
+        return Optional.of(items).map(map -> map.get(entityId)).map(state -> state.getModel());
```
ylwu-amzn, line 448: Should we track `lastUsedTime` here?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+    public ModelState<EntityModel> remove(String keyToRemove) {
+        // if shared cache is empty, we are using reserved memory
+        boolean reserved = sharedCacheEmpty();
+
+        key2Priority.remove(keyToRemove);
+        ModelState<EntityModel> valueRemoved = items.remove(keyToRemove);
+
+        if (valueRemoved != null) {
+            // if we releasing a shared cache item, release memory as well.
+            if (!reserved) {
```
ylwu-amzn, line 327: Memory tracker only track memory usage of shared cache ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
```diff
@@ -0,0 +1,534 @@
...
+            // it is fine we exceed a little. We have regular maintenance to remove
+            // extra memory usage.
+            buffer.put(modelId, state);
+        } else if (buffer.canReplace(priority)) {
+            // can replace an entity in the same CacheBuffer living in reserved
+            // or shared cache
+            // thread safe as each detector has one thread at one time and only the
+            // thread can access its buffer.
+            buffer.replace(modelId, state);
+        } else {
```
ylwu-amzn, line 253: If one detector runs for a long time, is it true that its entity will have higher priority than another detector which runs for a shorter time? Add some explanation of the priority algorithm can help to understand. Some entity runs for longer time should have higher priority than a new entity which just starts to run ?  How to make sure the priority algorithm is fair for entities of all detectors ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    /**
+     * Remove everything associated with the key and make a checkpoint.
+     *
+     * @param keyToRemove The key to remove
+     * @return the associated ModelState associated with the key, or null if there
+     * is no associated ModelState for the key
+     */
+    public ModelState<EntityModel> remove(String keyToRemove) {
```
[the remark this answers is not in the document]
kaituo: good catch.  Fixed by remove node from priorityList.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
```diff
@@ -0,0 +1,534 @@
...
+            checkpointDao
+                .restoreModelCheckpoint(
+                    modelId,
+                    ActionListener
+                        .wrap(checkpoint -> modelManager.processEntityCheckpoint(checkpoint, modelId, entityName, state), exception -> {
+                            Throwable cause = Throwables.getRootCause(exception);
+                            if (cause instanceof IndexNotFoundException) {
+                                modelManager.processEntityCheckpoint(Optional.empty(), modelId, entityName, state);
+                            } else if (cause instanceof RejectedExecutionException) {
+                                LOG.error("too many requests");
```
ylwu-amzn, line 295: Add more info to make the error message clear, such as "too many get AD model checkpoint requests"

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
```diff
@@ -0,0 +1,534 @@
...
+
+                // update state using new priority or create a new one
+                if (state != null) {
+                    state.setPriority(priority);
+                } else {
+                    EntityModel model = new EntityModel(modelId, new ArrayDeque<>(), null, null);
+                    state = new ModelState<>(model, modelId, detectorId, ModelType.ENTITY.getName(), clock, priority);
+                }
+
+                if (random.nextInt(1000) == 1) {
```
ylwu-amzn, line 186: As we discussed, this `get` method works on entity level, if one detector has 10K entities, will call `tryClearUpMemory` about 10 times per interval.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
kaituo: yes, entity will have higher priority than another detector which runs for a shorter time given other things are equal.

Yes, some entity runs for longer time should have higher priority than a new entity which just starts to run given other things are equal.

We ensure fairness by using periods instead of absolute duration.  Entity A accessed once three intervals ago should have the same priority with entity B accessed once three periods ago, irregardless of they belong to  detectors of different intervals.

added some explanation.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
kaituo: yeah, adjusted to 10k.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: I don't invented this.  LFU/LRU cache use this as well.  Both put/get are gonna impact order of elements.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    private void tryClearUpMemory() {
+        try {
+            if (maintenanceLock.tryLock()) {
+                clearMemory();
+            } else {
+                threadPool
+                    .schedule(
+                        () -> clearMemoryRunnable.run(),
```
ylwu-amzn, line 368: If not get lock, will schedule to clear memory in 90 seconds. But `clearMemoryRunnable` will not try to get lock any more, is that by design?

kaituo: not 90 seconds, but a random integer between [0, 89].

You are right.  I will use tryClearMmoery in clearMemoryRunnable instead.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
```diff
@@ -0,0 +1,534 @@
...
+    @Override
+    public void maintenance() {
+        // clean up memory if we allocate more memory than we should
+        tryClearUpMemory();
+        activeEnities.entrySet().stream().forEach(cacheBufferEntry -> {
+            String detectorId = cacheBufferEntry.getKey();
+            CacheBuffer cacheBuffer = cacheBufferEntry.getValue();
+            // remove expired cache buffer
+            if (cacheBuffer.expired(modelTtl)) {
+                activeEnities.remove(detectorId);
```
ylwu-amzn, line 451: Why only clear `activeEntities`, should we clear `inActiveEntities`, `doorKeepers` as well?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
ylwu-amzn: To make it safe, suggest to clear these cache explicitly by call their `clear()` method too, in case the `CacheBuffer` referred by somewhere we don't know.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
kaituo: inActiveEntities is a Guava's LRU cache.  The data structure itself is gonna evict items if they are inactive for 3 days or its maximum size reached (1 million entries)

doorkeepers are maintained in the following few lines.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: added these clear

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
```diff
@@ -0,0 +1,534 @@
...
+    public void clear(String detectorId) {
+        if (detectorId == null) {
+            return;
+        }
+        CacheBuffer buffer = activeEnities.remove(detectorId);
+        if (buffer != null) {
+            buffer.clear();
+        }
+        checkpointDao.deleteModelCheckpointByDetectorId(detectorId);
+        doorKeepers.remove(detectorId);
```
ylwu-amzn, line 484: Clear `inActiveEntities` ?

kaituo: inactiveEntities are not ordered by detectorId, not easy.  They will be evicted eventually.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
```diff
@@ -0,0 +1,534 @@
...
+     * Whether an entity is active or not
+     * @param detectorId The Id of the detector that an entity belongs to
+     * @param entityId Entity Id
+     * @return Whether an entity is active or not
+     */
+    @Override
+    public boolean isActive(String detectorId, String entityId) {
+        CacheBuffer cacheBuffer = activeEnities.get(detectorId);
+        if (cacheBuffer != null) {
+            return cacheBuffer.isActive(entityId);
```
ylwu-amzn, line 511: The `entityId` should be `entityModelId` ?
I remember the entityModelId contains detector id and entityValue
```
String entityModelId = detectorId + "_entity_" + entityValue;
```

kaituo: changed all over the places.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
ylwu-amzn: LFU/LRU are general cache solutions, I think this `CacheBuffer` is specially for multi-entity detectors. I'm ok to keep it as is. As I said, "It's ok for now as this get method only used when we run detection model". I'm not 100% understand why need to update frequency when `get`, respect your decision as I said "Add some TODO if you think it's **necessary**"

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
ylwu-amzn: Got it, I'm fine if you prefer to keep it as is now. Just in case someone else have difficulty to read this code as we mentioned entity model id somewhere else like this function: 

```
public String getEntityModelId(String detectorId, String entityValue) {
        return detectorId + "_entity_" + entityValue;
}
```

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
ylwu-amzn: Got it. make sense.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: without updating frequency when get, we cannot keep entity in cache "hot".  It will only be counted once when put into the cache and the frequency will not change any more.  Soon it will be replaced even though the entity is accessed frequently.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
ylwu-amzn: But we do add the reserved memory to AD total memory usage, right ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: I changed throughout the code.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/CacheBuffer.java
kaituo: yes

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
ylwu-amzn: Cool. So the `modelTtl` will be longer than 3 days? 
How about add some comments here so others will not get confused by not deleting `inActiveEntities` ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
ylwu-amzn: Cool. Is it reasonable to create separate `inactiveEntities` cache for different detectors ? Like this

```
// detectorId -> Cache
private Map<String, Cache<String, ModelState<EntityModel>>> inActiveEntities;
```

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
kaituo: How do we determine each inActiveEntities's max size?  If we give a uniform size, it is not fair since some detectors might only need a few entries, while another might need much more.

ylwu-amzn: Make sense. Build another dynamic algorithm to determine cache size for each detector seems not easy. Let's see if this late/eventual eviction will impact cluster performance. If the memory usage impacts a lot and we need to clean one detector's cache in time, some easy solution if iterating all keys in cache and remove keys belongs to the detector, definitely that will consume more CPU. 

Sample code

```
ConcurrentMap<String, Long> entityCacheMap = inActiveEntities.asMap();
entityCacheMap.forEach((k, v) -> {
            if (k.startsWith(detector.getName())) {
                cache.invalidate(k);
            }
});
```

ylwu-amzn (approved):
LGTM. Thanks for the change!

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java
kaituo: yes, agreed.  The total memory of inactive cache should be a few MBs.  Let's see the performance impact.

ylwu-amzn: Yes, one detector will consume few MBs, let's say 4MB, if 10 HC detectors be removed, it will be 40MB inactive cache, if user create another 10 HC detectors, then delete them, will be another 40MB, so on and so forth. Not sure how it will impact, let's monitor. Performance test can help.

kaituo: I meant the total inactive cache consumes few MBs in current implementation. After a detector is removed, it is still the same size.  After some time, these entries of removed detector is removed.

ylwu-amzn: Cool, misunderstood, memory usage should be ok. Will it impact new detector's inactive entities to be added into cache if the cache is full?

kaituo: it won't. Inactive entity cache is a LRU cache.

PR #256, opened 2020-10-14
Add multi-entity checkpoints read and write

kaituo:

*Issue #, if available:*

*Description of changes:*

We need checkpoints to save states and models on disk. In single-entity detectors, we store rcf and threshold models separately in different docs.  In multi-entity detectors, we need to store them together as we don't use distributed models anymore.  We also need to store recent sample history when the models are not ready.

This PR adds functions to serialize models and samples together in one doc and deserialize them when needed.  Also, we bulk indexing multi-entity detectors' checkpoints. Bulk requests will yield much better performance than single-document index requests  What's more, I add detectorId field in the checkpoint index to be able to query checkpoints by detector id.

Testing done:
1. added unit tests.
2. end-to-end testing

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
@@ -280,9 +280,7 @@ public class CheckpointDao {
 
     private void flush(BulkRequest bulkRequest) {
         clientUtil.<BulkRequest, BulkResponse>execute(BulkAction.INSTANCE, bulkRequest, ActionListener.wrap(r -> {
-            if (r.hasFailures() == false) {
-                logger.debug("Succeeded in bulking checkpoints");
-            } else {
+            if (r.hasFailures()) {
                 requests.addAll(BulkUtil.getIndexRequestToRetry(bulkRequest, r));
             }
         }, e -> {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
@@ -171,20 +171,7 @@ public class CheckpointDao {
         if (indexUtil.doesCheckpointIndexExist()) {
             saveModelCheckpointSync(source, modelId);
         } else {
-            indexUtil.initCheckpointIndex(ActionListener.wrap(initResponse -> {
-                if (initResponse.isAcknowledged()) {
-                    saveModelCheckpointSync(source, modelId);
-                } else {
-                    throw new RuntimeException("Creating checkpoint with mappings call not acknowledged.");
-                }
-            }, exception -> {
-                if (ExceptionsHelper.unwrapCause(exception) instanceof ResourceAlreadyExistsException) {
-                    // It is possible the index has been created while we sending the create request
-                    saveModelCheckpointSync(source, modelId);
-                } else {
-                    logger.error(String.format("Unexpected error creating index %s", indexName), exception);
-                }
-            }));
+            onCheckpointNotExist(source, modelId, false, null);
         }
     }
 
@@ -206,21 +193,33 @@ public class CheckpointDao {
         if (indexUtil.doesCheckpointIndexExist()) {
             saveModelCheckpointAsync(source, modelId, listener);
         } else {
-            indexUtil.initCheckpointIndex(ActionListener.wrap(initResponse -> {
-                if (initResponse.isAcknowledged()) {
-                    saveModelCheckpointAsync(source, modelId, listener);
-                } else {
-                    throw new RuntimeException("Creating checkpoint with mappings call not acknowledged.");
-                }
-            }, exception -> {
-                if (ExceptionsHelper.unwrapCause(exception) instanceof ResourceAlreadyExistsException) {
-                    // It is possible the index has been created while we sending the create request
-                    saveModelCheckpointAsync(source, modelId, listener);
-                } else {
-                    logger.error(String.format("Unexpected error creating index %s", indexName), exception);
-                }
-            }));
+            onCheckpointNotExist(source, modelId, true, listener);
         }
+    }
+
+    private void onCheckpointNotExist(Map<String, Object> source, String modelId, boolean isAsync, ActionListener<Void> listener) {
+        indexUtil.initCheckpointIndex(ActionListener.wrap(initResponse -> {
+            if (initResponse.isAcknowledged()) {
+                if (isAsync) {
+                    saveModelCheckpointAsync(source, modelId, listener);
+                } else {
+                    saveModelCheckpointSync(source, modelId);
+                }
+            } else {
+                throw new RuntimeException("Creating checkpoint with mappings call not acknowledged.");
+            }
+        }, exception -> {
+            if (ExceptionsHelper.unwrapCause(exception) instanceof ResourceAlreadyExistsException) {
+                // It is possible the index has been created while we sending the create request
+                if (isAsync) {
+                    saveModelCheckpointAsync(source, modelId, listener);
+                } else {
+                    saveModelCheckpointSync(source, modelId);
+                }
+            } else {
+                logger.error(String.format("Unexpected error creating index %s", indexName), exception);
+            }
+        }));
     }
 
     private void saveModelCheckpointAsync(Map<String, Object> source, String modelId, ActionListener<Void> listener) {
@@ -325,7 +324,7 @@ public class CheckpointDao {
             source.put(FIELD_MODEL, toCheckpoint(modelState.getModel()));
             source.put(TIMESTAMP, ZonedDateTime.now(ZoneOffset.UTC));
             requests.add(new IndexRequest(indexName).id(modelId).source(source));
-            modelState.setLastCheckpointTime(Instant.now());
+            modelState.setLastCheckpointTime(clock.instant());
             if (requests.size() >= maxBulkRequestSize) {
                 flush();
             }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
@@ -304,7 +304,9 @@ public class CheckpointDao {
     /**
      * Prepare bulking the input model state to the checkpoint index.
      * We don't save checkpoints within checkpointInterval again, except this
-     * is from cold start.
+     * is from cold start. This method will update the input state's last
+     *  checkpoint time if the checkpoint is staged (ready to be written in the
+     *  next batch).
      * @param modelState Model state
      * @param modelId Model Id
      * @param coldStart whether the checkpoint comes from cold start
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
@@ -15,35 +15,81 @@
 
 package com.amazon.opendistroforelasticsearch.ad.ml;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayDeque;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.elasticsearch.ExceptionsHelper;
+import org.elasticsearch.ResourceAlreadyExistsException;
 import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.DocWriteRequest;
+import org.elasticsearch.action.bulk.BulkAction;
+import org.elasticsearch.action.bulk.BulkItemResponse;
+import org.elasticsearch.action.bulk.BulkRequest;
+import org.elasticsearch.action.bulk.BulkResponse;
 import org.elasticsearch.action.delete.DeleteRequest;
 import org.elasticsearch.action.delete.DeleteResponse;
 import org.elasticsearch.action.get.GetRequest;
 import org.elasticsearch.action.get.GetResponse;
 import org.elasticsearch.action.index.IndexRequest;
 import org.elasticsearch.action.index.IndexResponse;
+import org.elasticsearch.action.support.IndicesOptions;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -98,14 +203,137 @@ public void putModelCheckpoint(String modelId, String modelCheckpoint, ActionLis
...
+            if (lock.isHeldByCurrentThread()) {
+                lock.unlock();
+            }
+        }
+    }
+
+    private void flush(BulkRequest bulkRequest) {
+        clientUtil.<BulkRequest, BulkResponse>execute(BulkAction.INSTANCE, bulkRequest, ActionListener.wrap(r -> {
+            if (r.hasFailures() == false) {
+                logger.debug("Succeeded in bulking checkpoints");
```
ohltyler, line 284: minor: remove this `debug` log

kaituo: Need this one when I am debugging :) Our default is Info. This won't impact things.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -167,9 +386,128 @@ public void deleteModelCheckpoint(String modelId) {
...
+            .setRequestsPerSecond(500); // throttle delete requests
+        logger.info("Delete checkpoints of detector {}", detectorID);
+        client.execute(DeleteByQueryAction.INSTANCE, deleteRequest, ActionListener.wrap(response -> {
+            if (response.isTimedOut() || !response.getBulkFailures().isEmpty() || !response.getSearchFailures().isEmpty()) {
+                logFailure(response, detectorID);
+            }
+            // if 0 docs get deleted, it means we cannot find matching docs
+            logger.info("{} " + DOC_GOT_DELETED_LOG_MSG, response.getDeleted());
+        }, exception -> {
+            if (exception instanceof IndexNotFoundException) {
```
ohltyler, line 419: I see some places are calling `ExceptionsHelper.unwrapCause(exception)`, but here it is not?

kaituo: ExceptionsHelper.unwrapCause is to get nested exception.  I don't see IndexNotFoundException is returned as nested.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -52,17 +98,60 @@
...
+
     /**
      * Constructor with dependencies and configuration.
      *
      * @param client ES search client
      * @param clientUtil utility with ES client
      * @param indexName name of the index for model checkpoints
+     * @param gson accessor to Gson functionality
+     * @param rcfSerde accessor to rcf serialization/deserialization
+     * @param thresholdingModelClass thresholding model's class
```
ohltyler, line 121: is there a reason for having a serialization/deserialization accessor for the RCF models but not for the thresholding models?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -167,9 +386,128 @@ public void deleteModelCheckpoint(String modelId) {
...
+                    rcf = rcfSerde.fromJson(json.getAsJsonPrimitive(ENTITY_RCF).getAsString());
+                }
+                ThresholdingModel threshold = null;
+                if (json.has(ENTITY_THRESHOLD)) {
+                    threshold = this.gson.fromJson(json.getAsJsonPrimitive(ENTITY_THRESHOLD).getAsString(), thresholdingModelClass);
+                }
+
+                String lastCheckpointTimeString = (String) (checkpoint.get(TIMESTAMP));
+                Instant timestamp = Instant.parse(lastCheckpointTimeString);
+                return new SimpleImmutableEntry<>(new EntityModel(modelId, samples, rcf, threshold), timestamp);
```
ohltyler, line 461: just to confirm, it is ok to pass null rcf and threshold models to the `EntityModel` constructor?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: rcf serialization/deserialization is more complex than threshold.  We have a separate package for it.  For threshold model, it is just standard gson.

src/main/resources/mappings/checkpoint.json
```diff
@@ -0,0 +1,18 @@
...
+  "_meta": {
+    "schema_version": 1
+  },
+  "properties": {
+    "detectorId": {
+      "type": "keyword"
+    },
+    "model": {
+      "type": "text"
+    },
```
ohltyler, line 12: to confirm, both rcf and threshold models will be stored under this mapping, correct?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: yes, it is ok

src/main/resources/mappings/checkpoint.json
kaituo: yes. One example:

"model":"{\"sp\":[],\"rcf\":\"{\\\"rng\\\":{... "th"...

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
ohltyler: I see, I wonder if there's a better way to standardize exceptions being nested or not, or if `unwrapCause()` already handles that. We have similar nested exception problems on the Kibana side. Can look at later.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
ohltyler: I see, makes sense.

src/main/resources/mappings/checkpoint.json
ohltyler: got it, makes sense. thanks

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: yeah, not sure.  If I want to look at A, but the nested is B.  I have to know A is top-level or not before checking.

kaituo: unwrapCause would give the innermost ElasticsearchException.

ohltyler: I see, yeah. no problem.

ohltyler (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -79,12 +168,28 @@ public void putModelCheckpoint(String modelId, String modelCheckpoint) {
...
+            }, exception -> {
+                if (ExceptionsHelper.unwrapCause(exception) instanceof ResourceAlreadyExistsException) {
+                    // It is possible the index has been created while we sending the create request
+                    saveModelCheckpointSync(source, modelId);
+                } else {
+                    logger.error(String.format("Unexpected error creating index %s", indexName), exception);
+                }
+            }));
+        }
+    }
```
yizheliu-amazon, line 189: looks like code is same as below. you may refactor them into a single method to avoid duplicate code.

kaituo: I tried, ended up giving up.  The difference is the save method: one is calling saveModelCheckpointSync with two parameters, while another is calling saveModelCheckpointAsync with threee parameters.  I created an funcitonal interface to consume three parameters (since JDK does not provide one) and make a generic method using functional interface.  I gave up because this interface and method are only used once inside the class and the amount of code is more.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -98,14 +203,135 @@ public void putModelCheckpoint(String modelId, String modelCheckpoint, ActionLis
...
+                            flush(bulkRequest);
+                        } else {
+                            throw new RuntimeException("Creating checkpoint with mappings call not acknowledged.");
+                        }
+                    }, exception -> {
+                        if (ExceptionsHelper.unwrapCause(exception) instanceof ResourceAlreadyExistsException) {
+                            // It is possible the index has been created while we sending the create request
+                            flush(bulkRequest);
+                        } else {
+                            logger.error(String.format("Unexpected error creating index %s", indexName), exception);
```
yizheliu-amazon, line 269: In which case we will call `write()` method of checkpointDao? Without knowing that, I am not sure whether it is okay to swallow the exception here.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -98,14 +203,135 @@ public void putModelCheckpoint(String modelId, String modelCheckpoint, ActionLis
...
+            if (lock.isHeldByCurrentThread()) {
+                lock.unlock();
+            }
+        }
+    }
+
+    private void flush(BulkRequest bulkRequest) {
+        clientUtil.<BulkRequest, BulkResponse>execute(BulkAction.INSTANCE, bulkRequest, ActionListener.wrap(r -> {
+            if (r.hasFailures()) {
+                requests.addAll(BulkUtil.getIndexRequestToRetry(bulkRequest, r));
```
yizheliu-amazon, line 284: is it possible that we retry infinitely?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -98,14 +203,135 @@ public void putModelCheckpoint(String modelId, String modelCheckpoint, ActionLis
...
+        // It is possible 2 states of the same model id gets saved: one overwrite another.
+        // This can happen if previous checkpoint hasn't been saved to disk, while the
+        // 1st one creates a new state without restoring.
+        if (modelState.getModel() != null) {
+            Map<String, Object> source = new HashMap<>();
+            source.put(DETECTOR_ID, modelState.getDetectorId());
+            source.put(FIELD_MODEL, toCheckpoint(modelState.getModel()));
+            source.put(TIMESTAMP, ZonedDateTime.now(ZoneOffset.UTC));
+            requests.add(new IndexRequest(indexName).id(modelId).source(source));
+            modelState.setLastCheckpointTime(Instant.now());
```
yizheliu-amazon, line 328: minor: if caller of this `write()` method expects `modelState` to be changed after `write()` is executed, I would prefer `write()` returns the updated `modelState`, instead of directly updating it, because usually `void` method is not expected to change state/values of input parameters. This is more for best practice. not blocking. feel free to skip it if no time to refactor.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
```diff
@@ -98,14 +203,135 @@ public void putModelCheckpoint(String modelId, String modelCheckpoint, ActionLis
...
+            if (requests.size() >= maxBulkRequestSize) {
```
yizheliu-amazon, line 329: How big `maxBulkRequestSize` could be? my concern is if `maxBulkRequestSize` is too big, and customer only has limited values for entity, is it possible that checkpointDao requests size never reach the max size, and never store these models?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: good idea.  Changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: we will call it whenever we need to save a checkpoint.  Any better option to not swallow it?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: It is possible when the system is not taking any more write requests.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: Could you give me a source that says void method should not change any input parameter's state?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
kaituo: This is autoflush logic.  The cache will manually call flush even though the number of requests does not exceed maxBulkRequestSize.

yizheliu-amazon: got it. Thanks

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
yizheliu-amazon: https://softwareengineering.stackexchange.com/questions/245767/is-modifying-an-incoming-parameter-an-antipattern
https://stackoverflow.com/questions/12019978/is-it-a-good-practice-to-change-arguments-in-java

There are discussions around this. But I think if you do need to modify input argument in a void method, at least make the method name clear or documentation clear that this method modifies which parameter. Again, not blocker issue.

kaituo: good to learn.  I'll add documentation to explain this.  I am using this smelly code throughout the code base.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/CheckpointDao.java
yizheliu-amazon: what if we throw the unexpected exception, will it break anything?

kaituo: not completely sure without testing.  I guess the upstream will eventually swallow it.

yizheliu-amazon: i see. thanks.

yizheliu-amazon (approved):

PR #257, opened 2020-10-14
Add state interfaces

kaituo:

*Issue #, if available:*

*Description of changes:*

We have different states in AD.  For better code readability and reuse, this PR adds three interfaces for our states:
* Expiring states: Represent a state that can be expired with a duration if not accessed.
* MaintenanceState: Represent a state that needs to maintain its metadata regularly
* CleanState: Represent a state organized via detectorId.  When deleting a detector's state, we can remove it from the state.

Testing done:
1. other unit tests cover them
2. end-to-end testing

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/CleanState.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/CleanState.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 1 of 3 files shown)


ohltyler:
Could you describe more on where/how these interfaces will be implemented with respect to HC detectors?

kaituo:
> Could you describe more on where/how these interfaces will be implemented with respect to HC detectors?

You can see examples where it is implemented here: https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/259/files

ohltyler (approved):

weicongs-amazon (approved):

PR #258, opened 2020-10-14
Add memory tracker

kaituo:

*Issue #, if available:*

*Description of changes:*

Previously, when creating a model, we evaluate all existing models and compare the total with the 10% heap memory limit.  If yes, we proceed to create the model.  Otherwise, we throw exceptions.  This does not work for multi-entity detectors.  First, there can be a lot of models in cache.  Reevaluating them every time we want to add a model is not efficient.  Second, we have two sources of memory usage now: single-entity and multi-entity detectors.  We need a central place to track memory usage across the board as we add more and more kinds of detectors.  This PR achieves the purpose.

This PR also updates RCF model size estimation.  Previously, we underestimated the size.

This PR also adds threshold model size estimation.  Previously, we didn't consider it.

This PR also adds a customized hashmap that can automatically consume and release memory.  This enables minimum change to our single-entity code as we just have to replace the map implementation.

Testing done:
1. added unit tests.
2. end-to-end testing pass.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
@@ -204,7 +204,7 @@ public class MemoryTracker {
      * Bytes to remove to keep AD memory usage within the limit
      * @return bytes to remove
      */
-    public long memoryToShed() {
+    public synchronized long memoryToShed() {
         return totalMemoryBytes - heapLimitBytes;
     }
 
@@ -234,12 +234,13 @@ public class MemoryTracker {
      * @param origin Origin
      * @param totalBytes total bytes from recomputing
      * @param reservedBytes reserved bytes from recomputing
+     * @return whether memory adjusted due to mismatch
      */
-    public synchronized void syncMemoryState(Origin origin, long totalBytes, long reservedBytes) {
+    public synchronized boolean syncMemoryState(Origin origin, long totalBytes, long reservedBytes) {
         long recordedTotalBytes = totalMemoryBytesByOrigin.getOrDefault(origin, 0L);
         long recordedReservedBytes = reservedMemoryBytesByOrigin.getOrDefault(origin, 0L);
         if (totalBytes == recordedTotalBytes && reservedBytes == recordedReservedBytes) {
-            return;
+            return false;
         }
         LOG
             .info(
@@ -261,5 +262,6 @@ public class MemoryTracker {
         long totalDiff = totalBytes - recordedTotalBytes;
         totalMemoryBytesByOrigin.put(origin, totalBytes);
         totalMemoryBytes += totalDiff;
+        return true;
     }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
@@ -1,0 +1,265 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad;
+
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.MODEL_MAX_SIZE_PERCENTAGE;
+
+import java.util.EnumMap;
+import java.util.Map;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.monitor.jvm.JvmService;
+
+import com.amazon.opendistroforelasticsearch.ad.common.exception.LimitExceededException;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.randomcutforest.RandomCutForest;
+
+/**
+ * Class to track AD memory usage.
+ *
+ */
+public class MemoryTracker {
+    private static final Logger LOG = LogManager.getLogger(MemoryTracker.class);
+
+    public enum Origin {
+        SINGLE_ENTITY_DETECTOR,
+        MULTI_ENTITY_DETECTOR
+    }
+
+    // memory tracker for total consumption of bytes
+    private long totalMemoryBytes;
+    private final Map<Origin, Long> totalMemoryBytesByOrigin;
+    // reserved for models. Cannot be deleted at will.
+    private long reservedMemoryBytes;
+    private final Map<Origin, Long> reservedMemoryBytesByOrigin;
+    private long heapSize;
+    private long heapLimitBytes;
+    private long desiredModelSize;
+    // we observe threshold model uses a fixed size array and the size is the same
+    private int thresholdModelBytes;
+    private int sampleSize;
+
+    /**
+     * Constructor
+     *
+     * @param jvmService Service providing jvm info
+     * @param modelMaxSizePercentage Percentage of heap for the max size of a model
+     * @param modelDesiredSizePercentage percentage of heap for the desired size of a model
+     * @param clusterService Cluster service object
+     * @param sampleSize The sample size used by stream samplers in a RCF forest
+     */
+    public MemoryTracker(
+        JvmService jvmService,
+        double modelMaxSizePercentage,
+        double modelDesiredSizePercentage,
+        ClusterService clusterService,
+        int sampleSize
+    ) {
+        this.totalMemoryBytes = 0;
+        this.totalMemoryBytesByOrigin = new EnumMap<Origin, Long>(Origin.class);
+        this.reservedMemoryBytes = 0;
+        this.reservedMemoryBytesByOrigin = new EnumMap<Origin, Long>(Origin.class);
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+        return deletedModeState;
+    }
+
+    @Override
+    public ModelState<RandomCutForest> put(K key, ModelState<RandomCutForest> value) {
+        ModelState<RandomCutForest> previousAssociatedState = super.put(key, value);
+        if (value != null && value.getModel() != null) {
+            long memoryToShed = memoryTracker.estimateModelSize(value.getModel());
+            memoryTracker.consumeMemory(memoryToShed, true, Origin.SINGLE_ENTITY_DETECTOR);
```
ohltyler, line 46: minor: rename `memoryToShed` to `memoryToConsume`

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
```diff
@@ -0,0 +1,265 @@
...
+     * @return estimated model size in bytes
+     */
+    public long estimateModelSize(AnomalyDetector detector, int numberOfTrees) {
+        return estimateModelSize(detector.getEnabledFeatureIds().size() * detector.getShingleSize(), numberOfTrees, sampleSize);
+    }
+
+    /**
+     * Gets the estimated size of an entity's model.
+     * RCF size:
+     * (Num_trees * num_samples * ( (16*dimensions + 84) + (24*dimensions + 48))）
```
ohltyler, line 176: minor: looks like this is the outdated formula for RCF size?

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
```diff
@@ -0,0 +1,265 @@
...
+        long rcfSize = totalSamples * (40 * dimension + 132);
+        long samplerSize = totalSamples * 36;
+        return rcfSize + samplerSize + thresholdModelBytes;
+    }
+
+    /**
+     * Bytes to remove to keep AD memory usage within the limit
+     * @return bytes to remove
+     */
+    public long memoryToShed() {
```
ohltyler, line 207: Make these and the below getters synchronized as well?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public ModelState<RandomCutForest> remove(Object key) {
+        ModelState<RandomCutForest> deletedModeState = super.remove(key);
+        if (deletedModeState != null && deletedModeState.getModel() != null) {
+            long memoryToShed = memoryTracker.estimateModelSize(deletedModeState.getModel());
+            memoryTracker.releaseMemory(memoryToShed, true, Origin.SINGLE_ENTITY_DETECTOR);
+        }
+        return deletedModeState;
```
ohltyler, line 38: 1. is it ok to return `null` here?
2. typo in `deletedModelState`

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class RCFMemoryAwareConcurrentHashmap<K> extends ConcurrentHashMap<K, ModelState<RandomCutForest>> {
+    private final MemoryTracker memoryTracker;
+
+    public RCFMemoryAwareConcurrentHashmap(MemoryTracker memoryTracker) {
+        this.memoryTracker = memoryTracker;
+    }
+
+    @Override
+    public ModelState<RandomCutForest> remove(Object key) {
```
ohltyler, line 32: could you explain why `put` and `remove` here are only being used to update the memory for single entity detectors?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
kaituo: I meant to the memory to get rid of, not to consume.

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
kaituo: It is not outdated.  For explanation's purpose, I separated two parts (16*dimensions + 84) and (24*dimensions + 48).  In the code, it is 40 * dimension + 132.

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
kaituo: Make sense.  Added synchronized to the first one.  We don't need it for the 2nd one as it is a constant.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
kaituo: 1. Yes, according to ConcurrenthashMap's doc:"the previous value associated with key, or null if there was no mapping for key"
2. good catch.  Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
kaituo: because multi-entity detectors don't use the hashmap.  I will update the doc to emphasize that.

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
ohltyler: I see, yeah makes sense

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
ohltyler: should this still be calling `consumeMemory()` then? why would this be memory to shed instead of memory to consume?

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
```diff
@@ -0,0 +1,265 @@
...
+        this.totalMemoryBytesByOrigin = new EnumMap<Origin, Long>(Origin.class);
+        this.reservedMemoryBytes = 0;
+        this.reservedMemoryBytesByOrigin = new EnumMap<Origin, Long>(Origin.class);
+        this.heapSize = jvmService.info().getMem().getHeapMax().getBytes();
+        this.heapLimitBytes = (long) (heapSize * modelMaxSizePercentage);
+        this.desiredModelSize = (long) (heapSize * modelDesiredSizePercentage);
+        clusterService
+            .getClusterSettings()
+            .addSettingsUpdateConsumer(MODEL_MAX_SIZE_PERCENTAGE, it -> this.heapLimitBytes = (long) (heapSize * it));
+        this.thresholdModelBytes = 180_000;
```
weicongs-amazon, line 83: how do we come out this number?

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
weicongs-amazon: how is this method used? do we need make it thread-safe?

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/RCFMemoryAwareConcurrentHashmap.java
kaituo: You are right.  Fixed the names.  Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * @param detectorId Detector Id, used in error message
+     * @param requiredBytes required bytes in memory
+     * @return whether there is memory required for AD
+     */
+    public synchronized boolean canAllocateReserved(String detectorId, long requiredBytes) {
+        if (reservedMemoryBytes + requiredBytes <= heapLimitBytes) {
+            return true;
+        } else {
+            throw new LimitExceededException(
```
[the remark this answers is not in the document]
kaituo: That's the old logic.  We will stop detector based on whether receiving the exception or not.

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
kaituo: You told me :)

kaituo: using jmap

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
kaituo: It is used here: https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/f6495c8657e3aa92addaf855167de1b2d470720b/src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityCache.java#L382

Yeah, I added synchronized after Tyler's comment

ohltyler (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/MemoryTracker.java
weicongs-amazon: :) got it. Here the **threshold** is for the threshold field in the RCF.

weicongs-amazon (approved):

PR #259, opened 2020-10-15
Rename TransportState to NodeState

kaituo:

*Issue #, if available:*

*Description of changes:*

Previously, we have TransportState and TransportStateManager to track states used by transport layers.  Now the state is not only used by the transport layer. Methods like getDetector are used by ModelManager etc.  This PR renames the class to reflect the fact.

This PR also modifies how we track whether a cold start is running or not.  Previously, the caller had to manually set it on and off.  And we have a code everywhere.  Now, we return a Releasable object that can be called automatically.  The current way is more concise and easier to avoid bugs.

This PR also adds states to track last index throttling time as we face index rejections issues.

Testing done:
1. will add unit tests.
2. end-to-end testing pass.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/NodeStateManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/NodeStateManager.java
@@ -13,12 +13,13 @@
  * permissions and limitations under the License.
  */
 
-package com.amazon.opendistroforelasticsearch.ad.transport;
+package com.amazon.opendistroforelasticsearch.ad;
 
 import static org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
 import java.time.Clock;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -29,6 +30,7 @@ import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.get.GetRequest;
 import org.elasticsearch.action.get.GetResponse;
 import org.elasticsearch.client.Client;
+import org.elasticsearch.common.lease.Releasable;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
 import org.elasticsearch.common.xcontent.NamedXContentRegistry;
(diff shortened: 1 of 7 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/NodeStateManager.java
```diff
@@ -127,7 +157,12 @@ public void getAnomalyDetector(String adID, ActionListener<Optional<AnomalyDetec
             ) {
                 ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser::getTokenLocation);
                 AnomalyDetector detector = AnomalyDetector.parse(parser, response.getId());
-                TransportState state = transportStates.computeIfAbsent(adID, id -> new TransportState(id, clock));
+                // end execution if all features are disabled
+                if (detector.getEnabledFeatureIds().isEmpty()) {
+                    listener.onFailure(new EndRunException(adID, CommonErrorMessages.ALL_FEATURES_DISABLED_ERR_MSG, true));
+                    return;
+                }
```
ohltyler, line 164: was this added just to give a more accurate exception and error message?

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/DeleteModelTransportActionTests.java
```diff
@@ -69,9 +71,10 @@ public void setUp() throws Exception {
 
         TransportService transportService = mock(TransportService.class);
         ActionFilters actionFilters = mock(ActionFilters.class);
-        TransportStateManager tarnsportStatemanager = mock(TransportStateManager.class);
+        NodeStateManager tarnsportStatemanager = mock(NodeStateManager.class);
         ModelManager modelManager = mock(ModelManager.class);
```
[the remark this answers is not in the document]
kaituo: good catch.  Changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/NodeStateManager.java
kaituo: This is added to to stop detector if all feature are disabled.

ohltyler: I misread this, I see. thanks

ohltyler (approved):

yizheliu-amazon (approved):

PR #262, opened 2020-10-15
Add data structure for entity model

kaituo:

*Issue #, if available:*

*Description of changes:*
This PR adds a new kind of model state called EntityModel.  An entity state contains RCF and threshold models, priority, and recent sets of values (128 sets at most) arising at AD job run time and their start/end timestamps.

Testing done:
1. These data structures are used throughout our code.  They are covered by their callers' unit tests.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelState.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelState.java
@@ -15,27 +15,35 @@
 
 package com.amazon.opendistroforelasticsearch.ad.ml;
 
+import java.time.Clock;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
 
+import com.amazon.opendistroforelasticsearch.ad.ExpiringState;
+
 /**
  * A ML model and states such as usage.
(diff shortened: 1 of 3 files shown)


yizheliu-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelState.java
```diff
@@ -44,15 +52,41 @@
...
-        this.lastUsedTime = lastUsedTime;
+        this.lastUsedTime = clock.instant();
+        // this is inaccurate until we find the last checkpoint time from disk
         this.lastCheckpointTime = Instant.MIN;
+        this.clock = clock;
+        this.priority = priority;
+    }
+
+    /**
+     * Create state with zero priority. Used in single-entity detector.
```
ohltyler, line 71: what is the purpose of adding priority here? How is that used by the HC detectors?

kaituo: It is used by our cache to determine whether to admit an entity state.  The higher the priority, the better chance.

ohltyler (approved):

PR #263, opened 2020-10-15
Add "main" function for model training and inference

kaituo:

*Issue #, if available:*

*Description of changes:*
This PR adds the "main" function for model training and inference.  If the model is in cache, the function asks the model directly for anomaly grade.  If the model is not in cache, the function triggers a cold start.

This PR also splits the model partition related functions to a separate class to break the circular dependency between ModelManager and NodeStateManager if we don't do so. This breakup also has better encapsulation as we have one class for one thing.

Testing done:
1.  Will add unit tests.
2. Manual tests pass.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
@@ -15,14 +15,11 @@
 
 package com.amazon.opendistroforelasticsearch.ad.ml;
 
-import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.MODEL_MAX_SIZE_PERCENTAGE;
-
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -31,6 +28,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -601,6 +514,7 @@ public void clear(String detectorId, ActionListener<Void> listener) {
         clearModels(detectorId, forests, ActionListener.wrap(r -> clearModels(detectorId, thresholds, listener), listener::onFailure));
     }
 
+    // TODO: we don't need to do iteration as this is a map.
```
ylwu-amzn, line 517: Can you explain more about this comment? From line 526, we need to iterate to get detector id from model id, then remove model from cache and checkpoint index.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
```diff
@@ -1046,6 +927,134 @@ public void getTotalUpdates(String modelId, String detectorId, ActionListener<Lo
...
+     *  result if the models are not ready
+     */
+    private ThresholdingResult maybeTrainBeforeScore(ModelState<EntityModel> modelState, String entityName) {
+        EntityModel model = modelState.getModel();
+        Queue<double[]> samples = model.getSamples();
+        String modelId = model.getModelId();
+        String detectorId = modelState.getDetectorId();
+        ThresholdingResult result = null;
+        if (model.getRcf() == null || model.getThreshold() == null) {
+            entityColdStarter.trainModel(samples, modelId, entityName, detectorId, modelState);
```
ylwu-amzn, line 1051: Just confirm, `entityColdStarter ` will train RCF model , even only threshold model is null ?

kaituo: yes

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: My original understanding is wrong.  Removed the comment.

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
ylwu-amzn: Is it reasonable to re-train threshold model only for that case?

kaituo: it seems yes as if threshold is null nothing can be detected.

ylwu-amzn: Seems we are talking two different things. I mean if RCF model is not null, threshold model is null, can we just train threshold model, not re-train RCF model?

ylwu-amzn (approved):
LGTM. Thanks for the change!

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ModelManager.java
kaituo: to train threshold model, we need data.  That's the expensive part.  After getting training data, it would be a waste to not use it with rcf and only use it with threshold model.

saratvemulapalli (approved):

PR #264, opened 2020-10-15
Add required config changes

kaituo:

*Issue #, if available:*

*Description of changes:*

This PR adds config changes required for multi-entity detectors.

Testing done:
1. Manually tested nothing breaks with these changes.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -20,6 +20,7 @@ import static java.util.Collections.unmodifiableList;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.time.Clock;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -36,6 +37,7 @@ import org.elasticsearch.client.Client;
 import org.elasticsearch.cluster.metadata.IndexNameExpressionResolver;
 import org.elasticsearch.cluster.node.DiscoveryNodes;
 import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.common.inject.Module;
 import org.elasticsearch.common.io.stream.NamedWriteableRegistry;
 import org.elasticsearch.common.settings.ClusterSettings;
 import org.elasticsearch.common.settings.IndexScopedSettings;
@@ -64,6 +66,9 @@ import org.elasticsearch.threadpool.ThreadPool;
 import org.elasticsearch.watcher.ResourceWatcherService;
 
 import com.amazon.opendistroforelasticsearch.ad.breaker.ADCircuitBreakerService;
+import com.amazon.opendistroforelasticsearch.ad.caching.CacheProvider;
+import com.amazon.opendistroforelasticsearch.ad.caching.EntityCache;
+import com.amazon.opendistroforelasticsearch.ad.caching.PriorityCache;
 import com.amazon.opendistroforelasticsearch.ad.cluster.ADClusterEventListener;
 import com.amazon.opendistroforelasticsearch.ad.cluster.HashRing;
 import com.amazon.opendistroforelasticsearch.ad.cluster.MasterEventListener;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -343,6 +360,86 @@ private static Void initGson() {
...
+            AnomalyDetectorSettings.MIN_PREVIEW_SIZE,
+            AnomalyDetectorSettings.HOURLY_MAINTENANCE,
+            AnomalyDetectorSettings.HOURLY_MAINTENANCE,
+            entityColdStarter,
+            modelPartitioner,
+            featureManager,
+            memoryTracker
+        );
+
+        // EntityCache cache = new LRUCache(
```
ylwu-amzn, line 415: Remove these commented code ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -223,13 +225,39 @@ private AnomalyDetectorSettings() {}
...
 
     // how many categorical fields we support
     public static final int CATEGORY_FIELD_LIMIT = 1;
 
+    public static final int MULTI_ENTITY_NUM_TREES = 10;
+
+    // cache related
+    public static final int DEDICATED_CACHE_SIZE = 10;
+
+    // We only keep priority (4 bytes float) in inactive cache. 1 million priorities
```
ylwu-amzn, line 248: So we store active entity's priority in inactive entity's cache ? If that's the case, why not name it as MAX_ENTITIES ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
kaituo: yeah, removed

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -239,4 +267,17 @@ private AnomalyDetectorSettings() {}
...
+        .intSetting("opendistro.anomaly_detection.max_primary_shards", 10, 0, Setting.Property.NodeScope, Setting.Property.Dynamic);
+
+    // max entity value's length
+    public static int MAX_ENTITY_LENGTH = 256;
+
+    // max number of index checkpoint requests in one bulk
+    public static int MAX_BULK_CHECKPOINT_SIZE = 1000;
+
+    // number of bulk checkpoints per second
+    public static double CHECKPOINT_BULK_PER_SECOND = 0.02;
```
ylwu-amzn, line 282: Can execute 1 bulk index checkpoint every 50 seconds?

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
kaituo: we don't do that.  Active entities' states has both models and priority, while inactive entities' states has only priority.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
kaituo: yeah,  1/60 is not a finite decimal representation.  It is fine.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
ylwu-amzn: Got it, misunderstanding, thought all priority of active&inactive entity will be stored in inactive cache.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -223,13 +225,39 @@ private AnomalyDetectorSettings() {}
...
+    public static final int MULTI_ENTITY_NUM_TREES = 10;
+
+    // cache related
+    public static final int DEDICATED_CACHE_SIZE = 10;
+
+    // We only keep priority (4 bytes float) in inactive cache. 1 million priorities
+    // take up 4 MB.
+    public static final int MAX_INACTIVE_ENTITIES = 1_000_000;
+
+    // TODO: check how much does 1 million insertion costs in memory
```
saratvemulapalli, line 252: Do we still have to do a performance test get the accurate number?

saratvemulapalli (approved):
The changes look good to me.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
kaituo: seems roughly 1 MB per 1 million from Weicong's jmap data.

ylwu-amzn (approved):
LGTM. Thanks for the change

PR #265, opened 2020-10-15
Add rcf score in ThresholdingResult

kaituo:

*Issue #, if available:*

*Description of changes:*
This PR adds an rcf score in ThresholdingResult so that we can use it when saving anomaly results.  Previously we don't need it since AnomalyResultTransportAction runs rcf first
and then threshold model. But for multi-entity result transport action, it sends the da
ta and gets back the ThresholdingResult, nothing in between.

Testing done:
1. Its callers have covered its usage in their unit tests.
2. Manual testing passes.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ThresholdingResult.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ThresholdingResult.java
@@ -24,16 +24,21 @@ public class ThresholdingResult {
 
     private final double grade;
     private final double confidence;
+    private final double rcfScore;
 
     /**
      * Constructor with all arguments.
      *
      * @param grade anomaly grade
      * @param confidence confidence for the grade
+     * @param rcfScore rcf score associated with the grade and confidence. Used
+     *   by multi-entity detector to differentiate whether the result is worth
+     *   saving or not.
      */
-    public ThresholdingResult(double grade, double confidence) {
+    public ThresholdingResult(double grade, double confidence, double rcfScore) {
         this.grade = grade;
         this.confidence = confidence;
+        this.rcfScore = rcfScore;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


yizheliu-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/ml/ThresholdingResult.java
```diff
@@ -24,16 +24,21 @@
...
      * Constructor with all arguments.
      *
      * @param grade anomaly grade
      * @param confidence confidence for the grade
+     * @param rcfScore rcf score associated with the grade and confidence. Used
+     *   by multi-entity detector to differentiate whether the result is worth
+     *   saving or not.
      */
-    public ThresholdingResult(double grade, double confidence) {
+    public ThresholdingResult(double grade, double confidence, double rcfScore) {
```
ohltyler, line 38: if this is only used for multi-entity detectors, shouldn't we still support the existing constructor, and have an additional constructor that takes in the `rcfScore` param?

kaituo: This is not only used for multi-entity detectors.  Single-entity detectors use it as well.

ohltyler: Sorry, I mean, if adding the `rcfScore` is only used for multi-entity detectors, shouldn't we keep the old constructor and add this new constructor that takes in the additional `rcfScore` param?

kaituo: Currently single-entity detectors get support from the constructor.  I don't break things.

ohltyler: Got it, wanted to confirm

ohltyler (approved):

PR #267, opened 2020-10-15
Add coordinating node logic

kaituo:

*Issue #, if available:*

*Description of changes:*

This PR adds coordinating node logic. To get feature data, the coordinating node aggregates log entries into multiple entity keys and their corresponding value vectors.  Then, the coordinating node sends the key-value pair to the model hosting nodes using consistent hashing.  Finally, it collects acknowledgements or exceptions and returns them to the job scheduler.

Testing done:
1. Will add unit tests.
2. Manual testing passes.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
@@ -15,12 +15,16 @@
 
 package com.amazon.opendistroforelasticsearch.ad.transport;
 
+import java.net.ConnectException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
```diff
@@ -234,14 +239,106 @@ protected void doExecute(Task task, ActionRequest actionRequest, ActionListener<
...
+                                        new ActionListenerResponseHandler<>(
+                                            new EntityResultListener(node.getId(), adID),
+                                            AcknowledgedResponse::new,
+                                            ThreadPool.Names.SAME
+                                        )
+                                    );
+                            });
+                    }
+
+                    listener.onResponse(new AnomalyResultResponse(0, 0, 0, new ArrayList<FeatureData>()));
```
ylwu-amzn, line 318: Always return empty anomaly result to job runner to persist ?

kaituo: We need to return sth to job runner.  Is it necessary to persist it? Currently I do as you told me to have a heartbeat.

saratvemulapalli (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportAction.java
ylwu-amzn: Heartbeat can be persisted in anomaly detection state. 
I'm ok with this currently if no impact to frontend.

kaituo: I did this for you.  If you think it is not necessary, I will remove it.

ylwu-amzn: Sure, you can remove it.

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #268, opened 2020-10-15
Add bulk indexing to result index handler

kaituo:

*Issue #, if available:*

*Description of changes:*
For multi-entities' results, we bulk indexing entities’ anomaly results of a detector instead of doing it one entity by another. Bulk requests will yield much better performance than single-document index requests.

This PR implements bulk indexing logic in MultiEntityResultHandler.

Testing done:
1. Will add unit tests.
2. Manual testing passes.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/AnomalyIndexHandler.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/AnomalyIndexHandler.java
@@ -49,25 +49,37 @@ import com.amazon.opendistroforelasticsearch.ad.util.RestHandlerUtils;
 
 public class AnomalyIndexHandler<T extends ToXContentObject> {
     private static final Logger LOG = LogManager.getLogger(AnomalyIndexHandler.class);
-
-    static final String CANNOT_SAVE_ERR_MSG = "Cannot save %s due to write block.";
     static final String FAIL_TO_SAVE_ERR_MSG = "Fail to save %s: ";
-    static final String RETRY_SAVING_ERR_MSG = "Retry in saving %s: ";
     static final String SUCCESS_SAVING_MSG = "Succeed in saving %s";
+    static final String CANNOT_SAVE_ERR_MSG = "Cannot save %s due to write block.";
+    static final String RETRY_SAVING_ERR_MSG = "Retry in saving %s: ";
 
     protected final Client client;
 
-    private final ThreadPool threadPool;
-    private final BackoffPolicy savingBackoffPolicy;
+    protected final ThreadPool threadPool;
+    protected final BackoffPolicy savingBackoffPolicy;
     protected final String indexName;
-    private final Consumer<ActionListener<CreateIndexResponse>> createIndex;
-    private final BooleanSupplier indexExists;
-    // whether save to a specific doc id or not
-    private final boolean fixedDoc;
+    protected final Consumer<ActionListener<CreateIndexResponse>> createIndex;
+    protected final BooleanSupplier indexExists;
+    // whether save to a specific doc id or not. False by default.
+    protected boolean fixedDoc;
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectionStateHandler.java
```diff
@@ -100,11 +100,11 @@ public DetectionStateHandler(
             DetectorInternalState.DETECTOR_STATE_INDEX,
             createIndex,
             indexExists,
-            true,
             clientUtil,
             indexUtils,
             clusterService
         );
+        this.fixedDoc = true;
```
ohltyler, line 107: why is `fixedDoc` true for `DetectionStateHandler`, and `false` for `AnomalyIndexHandler`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/AnomalyIndexHandler.java
```diff
@@ -90,12 +101,22 @@ public AnomalyIndexHandler(
...
         this.clusterService = clusterService;
     }
 
+    /**
+     * Since the constructor needs to provide injected value and Guice does not allow Boolean to be there
+     * (claiming it does not know how to instantiate it), caller needs to manually set it to true if
+     * it want to save to a specific doc.
+     * @param fixedDoc whether to save to a specific doc Id
+     */
+    public void setFixedDoc(boolean fixedDoc) {
```
weicongs-amazon, line 116: Minor: just FYI, Guice can support primitive type for injection. To do that, we need use bind with Named annotation.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/MultiEntityResultHandler.java
```diff
@@ -0,0 +1,154 @@
...
+            clientUtil,
+            indexUtils,
+            clusterService
+        );
+        this.nodeStateManager = nodeStateManager;
+        this.clock = clock;
+    }
+
+    public void write(AnomalyResult toSave, ADResultBulkRequest currentBulkRequest) {
+        currentBulkRequest.add(toSave);
```
weicongs-amazon, line 89: why do we need this simple method?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/DetectionStateHandler.java
kaituo: For DetectionStateHandler, we are writing to a specific doc.  By default, it is false.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/AnomalyIndexHandler.java
kaituo: Could you give me an example with code link?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/MultiEntityResultHandler.java
kaituo: removed

ohltyler (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/AnomalyIndexHandler.java
weicongs-amazon: similar like this: https://medium.com/@pavankumarreddy/injecting-dynamic-primitive-values-in-guice-26a1521b25b9

weicongs-amazon (approved):

PR #272, opened 2020-10-17
Added User support for background job and API Transport Actions

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Added User support for background job and API Transport Actions


src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
```diff
@@ -74,12 +76,14 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
...
             throw new IllegalStateException(CommonErrorMessages.DISABLED_ERR_MSG);
         }
+        String authHeader = request.header(ConfigConstants.AUTHORIZATION);
         SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
         searchSourceBuilder.parseXContent(request.contentOrSourceParamParser());
         searchSourceBuilder.fetchSource(getSourceContext(request));
         searchSourceBuilder.seqNoAndPrimaryTerm(true).version(true);
         SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(this.index);
-        return channel -> client.execute(actionType, searchRequest, search(channel));
+        SearchAnomalyRequest searchAnomalyRequest = new SearchAnomalyRequest(searchRequest, authHeader);
```
ohltyler, line 85: What was the reasoning for creating an entire new `SearchAnomalyRequest` class rather than adding the auth field to the `searchRequest` class?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -15,28 +15,100 @@
...
+        } else {
+            // Security is enabled and filter is enabled
+            Request authRequest = new AuthUserRequestBuilder(request.getAuthHeader()).build();
+            restClient.performRequestAsync(authRequest, new ResponseListener() {
+                @Override
+                public void onSuccess(Response response) {
+                    try {
+                        User user = new User(response);
+                        RestHandlerUtils.addFilter(user, request.getSearchRequest().source(), "user.backend_roles");
+                        logger.info("Filtering result by " + user.getBackendRoles());
```
ohltyler, line 96: minor: should this info log be removed?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
saratvemulapalli: Yeah good question, we need the authHeader to pass into the Transport layer. 
And existing SearchRequest couldn't support that.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
saratvemulapalli: Sure I'll move it to debug level.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/AbstractSearchAction.java
ohltyler: Cool, makes sense.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/DeleteAnomalyDetectorTransportAction.java
```diff
@@ -74,24 +75,31 @@ protected void doExecute(Task task, DeleteAnomalyDetectorRequest request, Action
...
+            }, exception -> {
+                if (exception instanceof IndexNotFoundException) {
+                    deleteDetectorStateDoc(detectorId, listener);
+                } else {
+                    LOG.error("Failed to delete anomaly detector job", exception);
+                    listener.onFailure(exception);
+                }
+            }));
+        } catch (Exception e) {
+            LOG.error(e);
```
skkosuri-amzn, line 101: Any particular reason you are not calling listener.onFailure(ex)?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/GetAnomalyDetectorTransportAction.java
```diff
@@ -102,16 +103,20 @@ protected void doExecute(Task task, GetAnomalyDetectorRequest request, ActionLis
...
+                MultiGetRequest multiGetRequest = new MultiGetRequest().add(adItem);
+                if (returnJob) {
+                    MultiGetRequest.Item adJobItem = new MultiGetRequest.Item(ANOMALY_DETECTOR_JOB_INDEX, detectorID).version(version);
+                    multiGetRequest.add(adJobItem);
+                }
+                client.multiGet(multiGetRequest, onMultiGetResponse(listener, returnJob, detectorID));
             }
-            client.multiGet(multiGetRequest, onMultiGetResponse(listener, returnJob, detectorID));
+        } catch (Exception e) {
+            LOG.error(e);
```
skkosuri-amzn, line 119: same comment log and call onFailure(..)

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportAction.java
```diff
@@ -72,25 +77,27 @@ protected void doExecute(Task task, IndexAnomalyDetectorRequest request, ActionL
...
+                maxMultiEntityAnomalyDetectors,
+                maxAnomalyFeatures,
+                method,
+                xContentRegistry,
+                restClient,
+                request.getAuthHeader()
+            );
+            indexAnomalyDetectorActionHandler.resolveUserAndStart();
         } catch (Exception e) {
             LOG.error(e);
```
skkosuri-amzn, line 102: same comment log and call onFailure(..)

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -15,28 +15,100 @@
...
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(FILTER_BY_BACKEND_ROLES, it -> filterEnabled = it);
     }
 
     @Override
-    protected void doExecute(Task task, SearchRequest request, ActionListener<SearchResponse> listener) {
+    protected void doExecute(Task task, SearchAnomalyRequest request, ActionListener<SearchResponse> listener) {
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            validateRole(request, listener);
+        } catch (Exception e) {
+            logger.error(e);
```
skkosuri-amzn, line 74: same comment log and call onFailure(..)

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/DeleteAnomalyDetectorTransportAction.java
saratvemulapalli: No reason, I missed it : )

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/GetAnomalyDetectorTransportAction.java
saratvemulapalli: Sure..done.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportAction.java
saratvemulapalli: Sure..done

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
saratvemulapalli: Sure..done

ohltyler (approved):
LGTM!

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -309,6 +313,12 @@ private static Void initGson() {
 
         double modelMaxSizePercent = AnomalyDetectorSettings.MODEL_MAX_SIZE_PERCENTAGE.get(settings);
 
+        try {
+            this.restClient = new SecureRestClientBuilder(settings, environment.configFile()).build();
+        } catch (IOException e) {
+            e.printStackTrace();
```
ylwu-amzn, line 319: change to log.error?

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
```diff
@@ -112,6 +120,8 @@
      * @param maxAnomalyFeatures      max features allowed per detector
      * @param method                  Rest Method type
      * @param xContentRegistry        Registry which is used for XContentParser
+     * @param restClient              RestClient used to talk to Security Plugin
```
ylwu-amzn, line 123: Can we only use this `restClient` to talk to security plugin ? Can we use it to call other plugins?

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
```diff
@@ -147,6 +147,9 @@ private AnomalyDetectorSettings() {}
             Setting.Property.Dynamic
         );
 
+    public static final Setting<Boolean> FILTER_BY_BACKEND_ROLES = Setting
+        .boolSetting("opendistro.anomaly_detection.filter_by_backend_roles", false, Setting.Property.NodeScope, Setting.Property.Dynamic);
```
ylwu-amzn, line 151: I see there are two roles fields in [User class](https://github.com/opendistro-for-elasticsearch/common-utils/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/commons/authuser/User.java#L46-L47).
What's the difference ? Should we add setting for them separately?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/RestHandlerUtils.java
```diff
@@ -101,6 +106,14 @@ public static String validateAnomalyDetector(AnomalyDetector anomalyDetector, in
         return null;
     }
 
+    public static void addFilter(User user, SearchSourceBuilder searchSourceBuilder, String fieldName) {
+        TermsQueryBuilder filterBackendRoles = QueryBuilders.termsQuery(fieldName, user.getBackendRoles());
+        if (searchSourceBuilder.query() instanceof BoolQueryBuilder) {
```
ylwu-amzn, line 111: What if the `searchSourceBuilder.query()` is not a `BoolQueryBuilder` ? 
For example, one user want to count all anomaly result. That will be just an aggregation query. For this case we will not add this `filterBackendRoles` term query to the original query, then user can get total anomaly result count for all users.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -15,28 +15,101 @@
...
+            listener.onFailure(e);
+        }
+    }
+
+    private void validateRole(SearchAnomalyRequest request, ActionListener<SearchResponse> listener) {
+        if (request.getAuthHeader() == null) {
+            // Auth Header is empty when 1. Security is disabled. 2. When user is super-admin
```
ylwu-amzn, line 81: Another possible case: old user created before FGAC enabled?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
```diff
@@ -15,28 +15,101 @@
...
+            // Proceed with search
+            search(request.getSearchRequest(), listener);
+        } else if (!filterEnabled) {
+            // Security is enabled and filter is disabled
+            // Proceed with search as user is already authenticated to hit this API.
```
ylwu-amzn, line 86: Security plugin will intercept and recheck user's permission on anomaly result index ? If not map read access of anomaly result index to this user role, security plugin will block and return error?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorTransportAction.java
```diff
@@ -15,28 +15,101 @@
...
+            // Auth Header is empty when 1. Security is disabled. 2. When user is super-admin
+            // Proceed with search
+            search(request.getSearchRequest(), listener);
+        } else if (!filterEnabled) {
+            // Security is enabled and filter is disabled
+            // Proceed with search as user is already authenticated to hit this API.
+            search(request.getSearchRequest(), listener);
+        } else {
+            // Security is enabled and filter is enabled
+            Request authRequest = new AuthUserRequestBuilder(request.getAuthHeader()).build();
```
ylwu-amzn, line 90: From this code, we can get `authHeader` from `SearchAnomalyRequest` and build authRequest with it, then perform secure REST API call.
Is it possible that we can build `authHeader` or `authRequest ` by ourselves without REST request for example in some schedule job ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
saratvemulapalli: Good point sure.

src/main/java/com/amazon/opendistroforelasticsearch/ad/settings/AnomalyDetectorSettings.java
saratvemulapalli: Good question.
Roles is something which is maintained by Security Plugin, you think of Internal roles.
Where as backend roles are determined from Authentication provider if the customer has integrated with other systems.
We decided to use backend role because it will be easier for customers so that their roles are automatically updated when there are changes in their organization rather opendistro internal roles.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
saratvemulapalli: Not really, this AuthHeader will not exist when security is disabled, this has nothing to do with FGAC. It is how security plugin works. Old user or any user will always have its credentials while logging in.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
saratvemulapalli: By the time the request comes to the Transport layer, the request is already authenticated i.e security plugin would have checked the permissions of the user.

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
saratvemulapalli: Good question, restClient can be used to talk to any rest endpoint.
We are using it to talk to our security plugin.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorTransportAction.java
```diff
@@ -15,28 +15,101 @@
...
+            restClient.performRequestAsync(authRequest, new ResponseListener() {
```
ylwu-amzn, line 91: I see the [AuthUserRequestBuilder](https://github.com/opendistro-for-elasticsearch/common-utils/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/commons/authuser/AuthUserRequestBuilder.java#L35) hard code auth URL in code. Is it possible/safe to let client define which URL to send request to? 
For example, 
1.User send REST request to AD
2.AD gets `authHeader` from user REST request
3.Build auth request to send request to alerting plugin's monitor creation URL?

saratvemulapalli: Definitely, RestClient is generic. 
We could use it to call any Rest endpoint from our transport.
Its just we are using a hardcoded API for our security plugin but we can have a generic request builder which we could use it for our monitors.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/RestHandlerUtils.java
```diff
@@ -101,6 +106,14 @@ public static String validateAnomalyDetector(AnomalyDetector anomalyDetector, in
         return null;
     }
 
+    public static void addFilter(User user, SearchSourceBuilder searchSourceBuilder, String fieldName) {
+        TermsQueryBuilder filterBackendRoles = QueryBuilders.termsQuery(fieldName, user.getBackendRoles());
```
ylwu-amzn, line 110: user backend roles is a list of roles, can this term query work for this case
1.User created a detector and runs it. At this time, user has roles [ad_view, ad_run, monitor_view]
2.After 1 week, user gets monitor creation and run permission, and his "ad_run" get revoked. So now user has roles [ad_view, monitor_view, monitor_create, monitor_run]

Question:
1.Will the detector still run after the "detector_run" permission get revoked ? Will we check permission for every job execution?
2.Can this term query work for this case ? That means we need to query ad user roles with current role `[ad_view, monitor_view, monitor_create, monitor_run]`

ylwu-amzn: Sorry, misunderstand, you are using termsQuery, so it should work from question2. Only need to consider question1.

saratvemulapalli: Correct.
Specifically to answer Q1, we will inject user role for the background job. This will check permissions for every run. 
Q2. Exactly terms query will take care of handling multiple roles.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/RestHandlerUtils.java
ylwu-amzn: Any updates? If it's not easy to build a flexible solution. How about we start from supporting `BoolQueryBuilder` like alerting ? We can support more query types in future.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyResultTransportAction.java
ylwu-amzn: Got it, so we are safe to stash thread context on line 71 to pass transport layer check, otherwise, security plugin will recheck user role permission on the index?

saratvemulapalli: Correct. Thats exactly how it works.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorTransportAction.java
saratvemulapalli: Yeah, its both yes and no : ).
Basically when the security plugin is used, any user who sends a request, the auth header will be set.
For the scheduled job we can do the same by injecting the role and making a rest request.
Instead we are directly calling transport, either way the permissions are being checked.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/RestHandlerUtils.java
saratvemulapalli: Yeah, its getting tricky support other queries, I spent some time to figure out if there is a cleaner way of doing it. 
Like you said, lets start supporting the bool query and we can add support to the other queries if needed later.

I will handle queries with an error for now.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorTransportAction.java
```diff
@@ -122,4 +125,14 @@ public void onFailure(Exception e) {
             }
         });
     }
+
+    private void addFilter(User user, SearchSourceBuilder searchSourceBuilder, String fieldName) {
+        TermsQueryBuilder filterBackendRoles = QueryBuilders.termsQuery(fieldName, user.getBackendRoles());
+        // For search detector queries, non BoolQuery is only used to find if the new detector name being created is
```
ylwu-amzn, line 131: If user use curl to search detector, and they don't use bool query, they can search other user's detector data. So we may have security leak.

How about we build a dedicated API to check if detector name is unique or not. The input is simple a detector name string, the output is true or false. And we control access in java code by ourselves. This need to change AD Kibana code to switch to this new API.

saratvemulapalli: Yeah sure that sounds good. I will work with Tyler and work on these changes.

ylwu-amzn: Sounds good. Will approve this one. Put the new API in another PR.

ylwu-amzn (approved):

PR #277, opened 2020-10-19
Fix nested field issue

kaituo:
*Issue #, if available:*

*Description of changes:*

Previously, our validation logic cannot parse nested fields from the getFieldMapping response because nested fields don't have the full name in the response.  For example, the field nested_host.host only has "host" in the type mapping instead of "nested_host.host".  This PR fixes that by not depending on field name when parsing the name type mapping.

Testing done:
1. manual testing passes.


ohltyler:
Have manually confirmed this works for both flat and nested category fields.

ohltyler (approved):
Thanks for fixing!

ylwu-amzn (approved):

yizheliu-amazon (approved):
thanks for the quick fix

PR #278, opened 2020-10-19
Upgrade mapping

kaituo:
*Issue #, if available:*

*Description of changes:*
This PR updates the current AD index mapping if a new version is detected.  Previously, we didn't do so.  This will cause issues when upgrading AD binary version: newly added fields will not be searchable, though we can still send get requests to fetch them.

This PR also makes the cache miss handling setting a dynamic setting.

Testing done:
1. manual upgrade testing passes with fgac on and off.
2. will add unit tests later.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestIndexAnomalyDetectorAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestIndexAnomalyDetectorAction.java
@@ -51,7 +51,6 @@ import org.elasticsearch.rest.action.RestResponseListener;
 
 import com.amazon.opendistroforelasticsearch.ad.AnomalyDetectorPlugin;
 import com.amazon.opendistroforelasticsearch.ad.constant.CommonErrorMessages;
-import com.amazon.opendistroforelasticsearch.ad.indices.AnomalyDetectionIndices;
 import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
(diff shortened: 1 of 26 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestIndexAnomalyDetectorAction.java
```diff
@@ -124,6 +126,7 @@ protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient cli
             ? WriteRequest.RefreshPolicy.parse(request.param(REFRESH))
             : WriteRequest.RefreshPolicy.IMMEDIATE;
         RestRequest.Method method = request.getHttpRequest().method();
+        anomalyDetectionIndices.updateMappingIfNecessary();
```
ylwu-amzn, line 129: Will this fail with FGAC ?

saratvemulapalli: Yeah..lets move this to transport layer. Since Anomaly Detection indices will be part of system index, we cannot update them.

kaituo: yeah, moved to transport layer.

ylwu-amzn (approved):
LGTM. Thanks for the change

yizheliu-amazon (approved):

PR #283, opened 2020-10-20
Add checkpoint index retention for multi entity detector

weicongs-amazon:
*Issue #, if available:*

*Description of changes:*
Add checkpoint index retention logic for multi entity detectors.

*Testing:*
* UT - Done
* Functional testing in test domain - Done

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/IndexCleanup.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/IndexCleanup.java
@@ -28,6 +28,7 @@ import org.elasticsearch.action.admin.indices.stats.ShardStats;
 import org.elasticsearch.action.support.IndicesOptions;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.common.util.concurrent.ThreadContext;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.reindex.DeleteByQueryAction;
 import org.elasticsearch.index.reindex.DeleteByQueryRequest;
@@ -113,10 +114,14 @@ public class IndexCleanup {
             .setQuery(queryForDeleteByQueryRequest)
             .setIndicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN)
             .setRefresh(true);
-        clientUtil.execute(DeleteByQueryAction.INSTANCE, deleteRequest, ActionListener.wrap(response -> {
-            // if 0 docs get deleted, it means our query cannot find any matching doc
-            LOG.info("{} docs are deleted for index:{}", response.getDeleted(), indexName);
-            listener.onResponse(response.getDeleted());
-        }, listener::onFailure));
+
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            clientUtil.execute(DeleteByQueryAction.INSTANCE, deleteRequest, ActionListener.wrap(response -> {
+                // if 0 docs get deleted, it means our query cannot find any matching doc
+                LOG.info("{} docs are deleted for index:{}", response.getDeleted(), indexName);
+                listener.onResponse(response.getDeleted());
+            }, listener::onFailure));
+        }
+
     }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/IndexCleanup.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/IndexCleanup.java
@@ -15,7 +15,9 @@
 
 package com.amazon.opendistroforelasticsearch.ad.cluster.diskcleanup;
 
-import com.amazon.opendistroforelasticsearch.ad.util.ClientUtil;
+import java.util.Arrays;
+import java.util.Objects;
+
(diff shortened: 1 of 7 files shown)


kaituo (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/IndexCleanup.java
```diff
@@ -0,0 +1,122 @@
...
+     * @param indexName index name
+     * @param queryForDeleteByQueryRequest query request
+     * @param listener action listener
+     */
+    public void deleteDocsByQuery(String indexName, QueryBuilder queryForDeleteByQueryRequest, ActionListener<Long> listener) {
+        DeleteByQueryRequest deleteRequest = new DeleteByQueryRequest(indexName)
+            .setQuery(queryForDeleteByQueryRequest)
+            .setIndicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN)
+            .setRefresh(true);
+        clientUtil.execute(DeleteByQueryAction.INSTANCE, deleteRequest, ActionListener.wrap(response -> {
```
ylwu-amzn, line 116: Will this work with FGAC? We have moved AD indices to system indices list. May not be able to delete directly.

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/ModelCheckpointIndexRetention.java
```diff
@@ -0,0 +1,111 @@
...
+                            .lte(clock.millis() - cleanUpTtl.toMillis())
+                            .format(CommonName.EPOCH_MILLIS_FORMAT)
+                    ),
+                ActionListener.wrap(cleanupNeeded -> {
+                    if (cleanupNeeded) {
+                        if (cleanUpTtl.equals(MINIMUM_CHECKPOINT_TTL)) {
+                            return;
+                        }
+
+                        Duration nextCleanupTtl = cleanUpTtl.minusDays(1);
```
ylwu-amzn, line 100: Why need to minus 1 day for next cleanup TTL? If the `defaultCheckpointTtl` is 5 days at first, the TTL will become shorter and shorter until reach 1 day ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/ModelCheckpointIndexRetention.java
```diff
@@ -0,0 +1,111 @@
...
+                    .filter(
+                        QueryBuilders
+                            .rangeQuery(CheckpointDao.TIMESTAMP)
+                            .lte(clock.millis() - defaultCheckpointTtl.toMillis())
+                            .format(CommonName.EPOCH_MILLIS_FORMAT)
+                    ),
+                ActionListener
+                    .wrap(
+                        response -> { cleanupBasedOnShardSize(defaultCheckpointTtl.minusDays(1)); },
+                        exception -> LOG.error("delete docs by query fails for checkpoint index", exception)
```
ylwu-amzn, line 75: If any exception happens, will stop clean up cron job? It's possible that high load causes failure and next run may succeed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/ModelCheckpointIndexRetention.java
weicongs-amazon: Yes. But this logic is only triggered whey the shard size is bigger than 50G.

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/ModelCheckpointIndexRetention.java
weicongs-amazon: It's possible. This error is acceptable since this is scheduled job. The process will continue in next schedule. Will add comment there

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/ModelCheckpointIndexRetention.java
ylwu-amzn: Got it, so the model checkpoint TTL is not hard constraint, we always prioritize the disk usage than checkpoint TTL. If the shard size still greater than 50GB after cleaning old checkpoints > 5days, will continue to clean old checkpoints older than 4 days, 3 days, until 1day.

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/ModelCheckpointIndexRetention.java
ylwu-amzn: Make sense.

src/main/java/com/amazon/opendistroforelasticsearch/ad/cluster/diskcleanup/IndexCleanup.java
weicongs-amazon: Good catch, thanks.  It still works with current solution. But still add the stash-threadcontext logic to be safe.

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #284, opened 2020-10-20
Renaming all Actions to distinguish between internal and external facing

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Renaming all actions to distinguish between Internal facing vs external facing.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #286, opened 2020-10-22
Adding new Search Detector Info API 

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Adding new Search Detector Info API to query total number of detectors and match name with existing detectors.
This API will search across all the detectors in detector index irrespective of User context.

Count Request: 
`GET _opendistro/_anomaly_detection/detectors/count`
Response: 
`{
  "count" : 4,
  "match" : false
}`

Match Request:
`GET _opendistro/_anomaly_detection/detectors/match?name=opendistro_ad`
Response:
`{
  "count" : 0,
  "match" : false
}`

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.transport;
+
+import org.elasticsearch.action.ActionType;
+
+public class SearchAnomalyDetectorInfoAction extends ActionType<SearchAnomalyDetectorInfoResponse> {
+    public static final SearchAnomalyDetectorInfoAction INSTANCE = new SearchAnomalyDetectorInfoAction();
+    public static final String NAME = "cluster:admin/opendistro/ad/detector/info";
+
+    private SearchAnomalyDetectorInfoAction() {
+        super(NAME, SearchAnomalyDetectorInfoResponse::new);
+    }
+
+}
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
@@ -17,9 +17,12 @@ package com.amazon.opendistroforelasticsearch.ad.transport;
 
 import org.elasticsearch.action.ActionType;
 
+import com.amazon.opendistroforelasticsearch.ad.constant.CommonValue;
+
 public class SearchAnomalyDetectorInfoAction extends ActionType<SearchAnomalyDetectorInfoResponse> {
+    // External Action which used for public facing RestAPIs.
+    public static final String NAME = CommonValue.EXTERNAL_ACTION_PREFIX + "detector/info";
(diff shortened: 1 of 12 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoTransportAction.java
```diff
@@ -0,0 +1,117 @@
...
+                client.search(searchRequest, new ActionListener<SearchResponse>() {
+
+                    @Override
+                    public void onResponse(SearchResponse searchResponse) {
+                        boolean nameExists = false;
+                        if (searchResponse.getHits().getTotalHits().value > 0) {
+                            nameExists = true;
+                        } else {
+                            nameExists = false;
+                        }
```
ohltyler, line 101: minor: could condense this into one line: `boolean nameExists = searchResponse.getHits().getTotalHits().value > 0`

ohltyler:
Have you discussed with the tech writer about adding these to the AD Open Distro docs?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoTransportAction.java
saratvemulapalli: Sure, that a nice idea.

saratvemulapalli:
> Have you discussed with the tech writer about adding these to the AD Open Distro docs?

Good question, I will work with the tech writer as we need to update bunch of other information as well for FGAC.

ohltyler:
> > Have you discussed with the tech writer about adding these to the AD Open Distro docs?
> 
> Good question, I will work with the tech writer as we need to update bunch of other information as well for FGAC.

Cool, don't see any other blockers, will approve.

ohltyler (approved):

ohltyler:
One more edge case here - in the case of the `.opendistro-anomaly-detectors` index not existing yet, can we return `0` here rather than an error? Tested in Kibana dev console

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
```diff
@@ -0,0 +1,28 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.transport;
+
+import org.elasticsearch.action.ActionType;
+
+public class SearchAnomalyDetectorInfoAction extends ActionType<SearchAnomalyDetectorInfoResponse> {
+    public static final SearchAnomalyDetectorInfoAction INSTANCE = new SearchAnomalyDetectorInfoAction();
```
ylwu-amzn, line 21: I see you switched NAME and INSTANCE sequence in this PR #284 , why we need to do that? Should we do the same thing here?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
```diff
@@ -0,0 +1,28 @@
...
+    public static final String NAME = "cluster:admin/opendistro/ad/detector/info";
```
ylwu-amzn, line 22: I see PR #284 merged,  change to use `CommonValue.INTERNAL_ACTION_PREFIX `?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
saratvemulapalli: Sure makes sense. 
I''ll rebase my changes and update it.

saratvemulapalli: I just saw the first part of the question, JVM is not smart enough, since INSTANCE depends on ACTION it must be defined first and if defined later string concat does not work. So I had to move it above.

saratvemulapalli:
> One more edge case here - in the case of the `.opendistro-anomaly-detectors` index not existing yet, can we return `0` here rather than an error? Tested in Kibana dev console

Sure had a chat offline, it makes sense to return 0, false when index doesnt exist.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchAnomalyDetectorInfoAction.java
saratvemulapalli: Sure, will take care of it.

ohltyler (approved):
Cool, thanks for the fix. LGTM

ylwu-amzn (approved):

PR #291, opened 2020-10-23
set user as nested type to support exists query

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
We need to set the user type as nested, then we can use exists query to check if `user` and `user.backend_roles` exists


kaituo (approved):

saratvemulapalli (approved):
Looks good to me!

PR #293, opened 2020-10-24
update user role filter to use nested query

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
We changed AD index mapping to use nested type for user field (#291 ). This PR is to update user role filter to nested query.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
@@ -28,15 +28,21 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.regex.Matcher;
 
+import org.apache.lucene.search.join.ScoreMode;
+import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.common.ParsingException;
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
 import org.elasticsearch.common.xcontent.NamedXContentRegistry;
 import org.elasticsearch.common.xcontent.XContentParser;
 import org.elasticsearch.common.xcontent.XContentType;
 import org.elasticsearch.index.query.BoolQueryBuilder;
+import org.elasticsearch.index.query.ExistsQueryBuilder;
+import org.elasticsearch.index.query.NestedQueryBuilder;
+import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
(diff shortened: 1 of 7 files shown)


ohltyler (approved):

yizheliu-amazon (approved):
thanks for the change

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -390,4 +397,41 @@ public static SearchSourceBuilder generateEntityColdStartQuery(
...
+        String userFieldName = "user";
+        String userBackendRoleFieldName = "user.backend_roles.keyword";
+        if (user == null) {
+            // For old monitor and detector, they have no user field, user = null
+            ExistsQueryBuilder userRolesFilterQuery = QueryBuilders.existsQuery(userFieldName);
+            NestedQueryBuilder nestedQueryBuilder = new NestedQueryBuilder(userFieldName, userRolesFilterQuery, ScoreMode.None);
+            boolQueryBuilder.mustNot(nestedQueryBuilder);
+        } else if (user.getBackendRoles() == null || user.getBackendRoles().size() == 0) {
+            // For simple FGAC user, they may have no backend roles, these users should be able to see detectors
+            // of other users whose backend role is empty. user != null, user.backend_role == null
```
kaituo, line 412: In this case, why don't we filter based on users' roles or just user name?

ylwu-amzn: Good question.
1.Why not use user's roles?
It's ok if all users use our predefined roles. But user can create custom role with arbitrary name. It's impossible to guess what permission the user role covers. Maybe role "abc" has exactly the same permission with "123".

2.Why not use user name ?
The main use case is for enterprise org/team. For one team, team members should have same permission and they can access team's resources no matter it's created by themselves or other team members. If limit resource access to only the creator, they have to share one account across team which is not good.

kaituo: 1. does filtering using backend role has the same issue?  backend role "abc" has the same permission with "123"?

2. user name based solution is opposite to your current solution: it is overly restrictive while your current solution is overly permissive.

ylwu-amzn: 1.Backend role has no mapping to permission. They just map to individuals or orgs. So if user1 and usr2 has same backend role, they are in same org.

2.Agree.

PR #294, opened 2020-10-26
fix no permission when preview detector with detector id

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
When FGAC enabled and preview detector with detector id, it throws no permission exception as can't read detector config index.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestExecuteAnomalyDetectorAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestExecuteAnomalyDetectorAction.java
@@ -36,6 +36,7 @@ import org.elasticsearch.client.node.NodeClient;
 import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.unit.TimeValue;
+import org.elasticsearch.common.util.concurrent.ThreadContext;
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
 import org.elasticsearch.common.xcontent.XContentBuilder;
 import org.elasticsearch.common.xcontent.XContentParser;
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestExecuteAnomalyDetectorAction.java
```diff
@@ -168,7 +169,12 @@ private String validateAdExecutionInput(AnomalyDetectorExecutionInput input) {
     private void preivewAnomalyDetector(NodeClient client, RestChannel channel, AnomalyDetectorExecutionInput input) {
         if (!StringUtils.isBlank(input.getDetectorId())) {
             GetRequest getRequest = new GetRequest(AnomalyDetector.ANOMALY_DETECTORS_INDEX).id(input.getDetectorId());
-            client.get(getRequest, onGetAnomalyDetectorResponse(channel, input));
+            try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+                client.get(getRequest, onGetAnomalyDetectorResponse(channel, input));
+            } catch (Exception e) {
+                logger.error("Fail to get detector for preview", e);
+                channel.sendResponse(new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
```
ohltyler, line 176: Do we always throw `RestStatus.INTERNAL_SERVER_ERROR` or is there any one specific to FGAC permissions errors? I'm just wondering if the frontend will propagate this correctly as a permissions issue.

ylwu-amzn: Good catch. This change will run without any user info, so if any exception happen, it should not be permission error. For permission error we should return "RestStatus.FORBIDDEN". 
It's hard to test this error as admin can always get detector. Let's tune the error message part later if it really happens.

lezzago (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestExecuteAnomalyDetectorAction.java
ohltyler: I see, makes sense

ohltyler (approved):

saratvemulapalli (commented):
This might be a problem. 
We are not validating user's permission to preview the detector. 
Which means any user can do a preview. Is that safe to assume?
Also Stashing User context in Rest API will lead into security vulnerabilities. 

Longer term I think we have to move preview to transport layer and have a dedicated action for it.

ylwu-amzn:
> This might be a problem.
> We are not validating user's permission to preview the detector.
> Which means any user can do a preview. Is that safe to assume?
> Also Stashing User context in Rest API will lead into security vulnerabilities.
> 
> Longer term I think we have to move preview to transport layer and have a dedicated action for it.

Preview returns sampled data, most results are interpolated values. Definitely, this is some short term solution to quick fix the alerting side user experience. We should move preview to transport action in long term.

PR #296, opened 2020-10-27
Fix edge case where entities found for preview is empty

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Fix edge case where entities found for preview is empty.

Have verified on test domain that empty result is returned with this fix.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorRunner.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorRunner.java
@@ -72,6 +72,9 @@ public final class AnomalyDetectorRunner {
         if (categoryField != null && !categoryField.isEmpty()) {
             featureManager.getPreviewEntities(detector, startTime.toEpochMilli(), endTime.toEpochMilli(), ActionListener.wrap(entities -> {
 
+                if (entities == null || entities.isEmpty()) {
+                    listener.onResponse(Collections.emptyList());
+                }
                 ActionListener<EntityAnomalyResult> entityAnomalyResultListener = ActionListener
                     .wrap(
                         entityAnomalyResult -> { listener.onResponse(entityAnomalyResult.getAnomalyResults()); },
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorRunner.java
```diff
@@ -72,6 +72,9 @@ public void executeDetector(AnomalyDetector detector, Instant startTime, Instant
         if (categoryField != null && !categoryField.isEmpty()) {
             featureManager.getPreviewEntities(detector, startTime.toEpochMilli(), endTime.toEpochMilli(), ActionListener.wrap(entities -> {
 
+                if (entities == null || entities.isEmpty()) {
+                    listener.onResponse(Collections.emptyList());
```
ylwu-amzn, line 76: Should we return empty results or some failure like this https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java#L525

yizheliu-amazon: Such failure will be caught and empty result is still returned instead in [onFailure()](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorRunner.java#L117-L119)

I think it is okay to just return empty result

ylwu-amzn (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorRunner.java
kaituo: Could you tell me why it would end up calling onFailure?  listener is the rest channel.

yizheliu-amazon: If entities is empty, code execution won't go into the for loop [below](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorRunner.java#L87-L103), and `listener` has no chance to call `onFailure` or `onResponse`

kaituo: I feel it is better to return some failure and show that on kibana than return empty result.  Customers may wonder why it returns nothing.

yizheliu-amazon: if empty list is returned, existing Kibana will show corresponding message.

kaituo: Can you add a comment explaining why we want to return empty responses instead of errors?  It helps code readability.

kaituo (approved):

PR #301, opened 2020-10-30
fix null user in detector

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
Detector has null user as we can't get user from thread context and we don't set user in detector when create/update detector.

Test locally and on test cluster.


saratvemulapalli (approved):

skkosuri-amzn (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
```diff
@@ -448,7 +448,7 @@ private void indexAnomalyDetector(String detectorId) throws IOException {
             anomalyDetector.getSchemaVersion(),
             Instant.now(),
             anomalyDetector.getCategoryField(),
-            anomalyDetector.getUser()
+            user
```
ylwu-amzn, line 451: We didn’t put new user from thread context to detector, so the save will store null value

PR #306, opened 2020-11-04
filter out empty value for entity features

ylwu-amzn:
*Issue #, if available:*

*Description of changes:*
HC detector stays on initialization for days. We can see a lot of “NoSuchElementException” exception which caused by `Optional.EMPTY.get()`. This caused by null feature aggregation result (the category field value is not null). So if some entities have null feature data, will throw `NoSuchElementException` exception and skip other normal entities.


zhanghg08 (approved):
LGTM, thanks for the change.

yizheliu-amazon (approved):

PR #309, opened 2020-11-09
Fix for upgrading mapping

kaituo:
*Issue #, if available:*

*Description of changes:*

This PR fixes two bugs of upgrading mapping:
First, previously we only tried once when upgrading mapping.  It is possible we need to try a few more times.  The drawback is that if the code has a bug, we retry endlessly.  I should fix it in the future.
Second, we enclose the upgrade mapping function call under a user context in a recent commit.  Upgrade mapping will fail when security plugin is enabled since a regular user cannot upgrade mappings of system indices.

The PR also adds upper bounds to our dynamic settings if reasonable.
Testing done:
1. Tested that upgrade mapping succeeds with new changes.
2. Tested the upper bounds of the changed settings are in effect.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
@@ -581,7 +581,7 @@ public class AnomalyDetectionIndices implements LocalNodeMasterListener {
             ActionListener.wrap(r -> updateRunning.set(false), exception -> {
                 // TODO: don't retry endlessly. Can be annoying if there are too many exception logs.
                 updateRunning.set(false);
-                logger.error("Fail to updatea all mappings");
+                logger.error("Fail to update AD indices' mappings");
             }),
             updates.size()
         );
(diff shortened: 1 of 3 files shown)


saratvemulapalli (approved):
Thanks for the change, didnt realize the update mapping would fail.

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -578,7 +578,11 @@ public void updateMappingIfNecessary() {
         }
 
         final GroupedActionListener<Void> conglomerateListeneer = new GroupedActionListener<>(
-            ActionListener.wrap(r -> updateRunning.set(false), exception -> logger.error("Fail to updatea all mappings")),
+            ActionListener.wrap(r -> updateRunning.set(false), exception -> {
+                // TODO: don't retry endlessly. Can be annoying if there are too many exception logs.
+                updateRunning.set(false);
```
ylwu-amzn, line 583: If remove this line, will not update index mapping if any exception happens?

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -578,7 +578,11 @@ public void updateMappingIfNecessary() {
...
+                logger.error("Fail to updatea all mappings");
```
ylwu-amzn, line 584: Add ad to the error message.

kaituo: added

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
kaituo: if removing this line, we will only try to update mapping once per node.

ylwu-amzn (approved):
LGTM. Thanks for the change!

PR #312, opened 2020-11-17
Fix issue where data hole exists for Preview API

yizheliu-amazon:
*Issue #, if available:*

*Description of changes:*
Fix issue where data hole exists for Preview API. For now, we only get cold start samples with non-zero doc, and we add samples to corresponding date range by [index of cold start samples](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/2e7e39d01631e27822b64cdd93e75956d72efdf2/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java#L607-L613). This can cause issue because when source data has no doc in periods of time, size of `featureSamples` is less than `sampleRanges`, then only the some beginning date ranges of `sampleRanges` have data returned.

Fix is simple: just include zero count bucket in the response from ES. 

Example: data only exists in source starting from 11-13-2020
Before:
Screen Shot 2020-11-14 at 5 24 12 PM

After:
Screen Shot 2020-11-14 at 5 22 53 PM

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -594,7 +594,13 @@ public class FeatureManager implements CleanState {
         ActionListener<Entry<List<Entry<Long, Long>>, double[][]>> listener
     ) throws IOException {
         searchFeatureDao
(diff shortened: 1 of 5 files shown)


yizheliu-amazon:
Checks are failing because `set-env` is disabled. Have run gradle build locally to make sure all tests pass

kaituo:
> Checks are failing because `set-env` is disabled. Have run gradle build locally to make sure all tests pass

Weicong's PR fixed it.

kaituo (approved):

ylwu-amzn (approved):

PR #325, opened 2020-12-07
Adding role based filtering for rest of APIs

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Adding backend user role based filtering to Delete, Start/Stop, Get/Profile and Update APIs.
Also added support to block users creating detectors when AD Backend Role based filtering is turned on and user does not have backend roles. 

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobTransportAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobTransportAction.java
@@ -15,9 +15,10 @@
 
 package com.amazon.opendistroforelasticsearch.ad.transport;
 
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES;
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.REQUEST_TIMEOUT;
-
-import java.io.IOException;
+import static com.amazon.opendistroforelasticsearch.ad.util.ParseUtils.getDetector;
+import static com.amazon.opendistroforelasticsearch.ad.util.ParseUtils.getUserContext;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -25,6 +26,7 @@ import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.support.ActionFilters;
 import org.elasticsearch.action.support.HandledTransportAction;
 import org.elasticsearch.client.Client;
+import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.common.inject.Inject;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.unit.TimeValue;
@@ -34,64 +36,104 @@ import org.elasticsearch.tasks.Task;
 import org.elasticsearch.transport.TransportService;
 
 import com.amazon.opendistroforelasticsearch.ad.indices.AnomalyDetectionIndices;
+import com.amazon.opendistroforelasticsearch.ad.rest.handler.AnomalyDetectorFunction;
 import com.amazon.opendistroforelasticsearch.ad.rest.handler.IndexAnomalyDetectorJobActionHandler;
+import com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings;
 import com.amazon.opendistroforelasticsearch.ad.util.RestHandlerUtils;
+import com.amazon.opendistroforelasticsearch.commons.authuser.User;
 
 public class AnomalyDetectorJobTransportAction extends HandledTransportAction<AnomalyDetectorJobRequest, AnomalyDetectorJobResponse> {
     private final Logger logger = LogManager.getLogger(AnomalyDetectorJobTransportAction.class);
 
     private final Client client;
+    private final ClusterService clusterService;
     private final Settings settings;
     private final AnomalyDetectionIndices anomalyDetectionIndices;
     private final NamedXContentRegistry xContentRegistry;
+    private volatile Boolean filterByEnabled;
 
     @Inject
     public AnomalyDetectorJobTransportAction(
         TransportService transportService,
         ActionFilters actionFilters,
         Client client,
+        ClusterService clusterService,
         Settings settings,
         AnomalyDetectionIndices anomalyDetectionIndices,
         NamedXContentRegistry xContentRegistry
     ) {
         super(AnomalyDetectorJobAction.NAME, transportService, actionFilters, AnomalyDetectorJobRequest::new);
         this.client = client;
+        this.clusterService = clusterService;
         this.settings = settings;
         this.anomalyDetectionIndices = anomalyDetectionIndices;
         this.xContentRegistry = xContentRegistry;
+        filterByEnabled = AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES.get(settings);
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(FILTER_BY_BACKEND_ROLES, it -> filterByEnabled = it);
     }
 
     @Override
     protected void doExecute(Task task, AnomalyDetectorJobRequest request, ActionListener<AnomalyDetectorJobResponse> listener) {
+        String detectorId = request.getDetectorID();
+        // By the time request reaches here, the user permissions are validated by Security plugin.
+        User user = getUserContext(client);
+        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
+            resolveUserAndExecute(user, detectorId, listener, () -> adJobExecute(request, listener));
+        } catch (Exception e) {
+            logger.error(e);
+            listener.onFailure(e);
+        }
+    }
+
+    private void adJobExecute(AnomalyDetectorJobRequest request, ActionListener<AnomalyDetectorJobResponse> listener) {
         String detectorId = request.getDetectorID();
         long seqNo = request.getSeqNo();
         long primaryTerm = request.getPrimaryTerm();
         String rawPath = request.getRawPath();
         TimeValue requestTimeout = REQUEST_TIMEOUT.get(settings);
 
-        // By the time request reaches here, the user permissions are validated by Security plugin.
-        // Since the detectorID is provided, this can only happen if User is part of a role which has access
-        // to the detector. This is filtered by our Search Detector API.
-
-        try (ThreadContext.StoredContext context = client.threadPool().getThreadContext().stashContext()) {
-            IndexAnomalyDetectorJobActionHandler handler = new IndexAnomalyDetectorJobActionHandler(
-                client,
-                listener,
-                anomalyDetectionIndices,
-                detectorId,
-                seqNo,
-                primaryTerm,
-                requestTimeout,
-                xContentRegistry
-            );
+        IndexAnomalyDetectorJobActionHandler handler = new IndexAnomalyDetectorJobActionHandler(
+            client,
+            listener,
+            anomalyDetectionIndices,
+            detectorId,
+            seqNo,
+            primaryTerm,
+            requestTimeout,
+            xContentRegistry
+        );
+        try {
             if (rawPath.endsWith(RestHandlerUtils.START_JOB)) {
                 handler.startAnomalyDetectorJob();
             } else if (rawPath.endsWith(RestHandlerUtils.STOP_JOB)) {
                 handler.stopAnomalyDetectorJob(detectorId);
             }
-        } catch (IOException e) {
+        } catch (Exception e) {
             logger.error(e);
             listener.onFailure(e);
+        }
+    }
+
+    private void resolveUserAndExecute(
+        User requestedUser,
+        String detectorId,
+        ActionListener<AnomalyDetectorJobResponse> listener,
+        AnomalyDetectorFunction function
+    ) {
+        if (requestedUser == null) {
+            // Security is disabled or user is superadmin
+            function.execute();
+        } else if (!filterByEnabled) {
+            // security is enabled and filterby is disabled.
+            function.execute();
+        } else {
+            // security is enabled and filterby is enabled.
+            // Get detector and check if the user has permissions to access the detector
+            try {
+                getDetector(requestedUser, detectorId, listener, function, client, clusterService, xContentRegistry);
+            } catch (Exception e) {
+                listener.onFailure(e);
+            }
         }
     }
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
@@ -17,6 +17,7 @@ package com.amazon.opendistroforelasticsearch.ad.util;
 
 import static com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector.QUERY_PARAM_PERIOD_END;
 import static com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector.QUERY_PARAM_PERIOD_START;
+import static org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 import static org.elasticsearch.search.aggregations.AggregationBuilders.dateRange;
 import static org.elasticsearch.search.aggregations.AggregatorFactories.VALID_AGG_NAME;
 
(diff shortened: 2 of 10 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobTransportAction.java
```diff
@@ -34,64 +36,104 @@
...
+        } else {
+            // security is enabled and filterby is enabled.
+            // Get detector and check if the user has permissions to access the detector
+            try {
+                getDetector(requestedUser, detectorId, listener, function, client, clusterService, xContentRegistry);
+            } catch (Exception e) {
+                listener.onFailure(e);
+            }
+        }
+    }
```
yizheliu-amazon, line 138: I have seen multiple `resolveUserAndExecute` for different TransportAction, is it possible to change it to a common method like below?

```
resolveUserAndExecute(User requestedUser,
        String detectorId,
        ActionListener<AnomalyDetectorJobResponse> listener,
        AnomalyDetectorFunction function,
        AnomalyDetectorFunction functionWithSecurityAndFilterBy) // a better name is needed. :)
 {

        if (requestedUser == null) {
            // Security is disabled or user is superadmin
            function.execute();
        } else if (!filterByEnabled) {
            // security is enabled and filterby is disabled.
            function.execute();
        } else {
            // security is enabled and filterby is enabled.
            // Get detector and check if the user has permissions to access the detector
            try {
                functionWithSecurityAndFilterBy.execute();
            } catch (Exception e) {
                listener.onFailure(e);
            }
        }
}

```

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -450,6 +457,99 @@ public static User getUserContext(Client client) {
...
+                    function.execute();
+                } else {
+                    logger.debug("User: " + requestUser.getName() + " does not have permissions to access detector: " + detectorId);
+                    listener.onFailure(new ElasticsearchException("User does not have permissions to access detector: " + detectorId));
+                }
+            } catch (Exception e) {
+                listener.onFailure(new ElasticsearchException("Unable to get user information from detector " + detectorId));
+            }
+        } else {
+            listener.onFailure(new ElasticsearchException("Could not find detector " + detectorId));
```
yizheliu-amazon, line 514: maybe replace with ResourceNotFound exception

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobTransportAction.java
saratvemulapalli: Yeah good point. I thought about it as well. I backed off because IndexAnomalyDetectorTransportAction needed a slightly different logic within this function. 

What I can do is stub this out to a common function and leave it for IndexAnomalyDetectorTransportAction.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
saratvemulapalli: +1 I like that sure.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -450,6 +458,126 @@ public static User getUserContext(Client client) {
...
+                        .wrap(
+                            response -> onGetAdResponse(response, requestUser, detectorId, listener, function, xContentRegistry),
+                            exception -> {
+                                logger.error("Failed to get anomaly detector: " + detectorId, exception);
+                                listener.onFailure(exception);
+                            }
+                        )
+                );
+        } else {
+            listener.onFailure(new ElasticsearchException("Failed to get anomaly detector: " + detectorId));
```
ylwu-amzn, line 512: How about change this error message to "Anomaly detector index not found"? "Failed to get anomaly detector" is too general and doesn't show exactly this error.

saratvemulapalli: Sure that makes sense as well. It's probably actionable for the user.

yizheliu-amazon (approved):
thanks for the change

ylwu-amzn (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -450,6 +458,129 @@ public static User getUserContext(Client client) {
         return User.parse(userStr);
     }
 
+    public static void resolveUserAndExecute(
```
kaituo, line 461: ParseUtils are utility functions to parse ES responses.  Can we move these functions to sth like SecurityUtils?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -450,6 +458,129 @@ public static User getUserContext(Client client) {
...
+                }
+            } catch (Exception e) {
+                listener.onFailure(new ElasticsearchException("Unable to get user information from detector " + detectorId));
+            }
+        } else {
+            listener.onFailure(new ResourceNotFoundException("Could not find detector " + detectorId));
+        }
+    }
+
+    public static boolean checkUserPermissions(User requestedUser, User resourceUser, String detectorId) throws Exception {
```
kaituo, line 549: Since this is a public function, what if the caller passes in a null requestedUser or resourceUser?  Also, we are assuming backend role filter is on before calling this function, what if users forget that?  If it is only used by other functions in the class, can we make it private?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -450,6 +458,129 @@ public static User getUserContext(Client client) {
...
+                            + " permissions to access detector: "
+                            + detectorId
+                    );
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public static boolean checkFilterByBackendRoles(User requestedUser, ActionListener listener) {
```
kaituo, line 571: You may want to add requestedUsers != null if this is a public method.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
```diff
@@ -43,11 +55,26 @@
...
+        org.elasticsearch.threadpool.ThreadPool mockThreadPool = mock(ThreadPool.class);
+        when(client.threadPool()).thenReturn(mockThreadPool);
+        when(mockThreadPool.getThreadContext()).thenReturn(threadContext);
 
         action = new AnomalyDetectorJobTransportAction(
             mock(TransportService.class),
             mock(ActionFilters.class),
-            client(),
+            client,
+            clusterService,
```
kaituo, line 77: Not related to this line, when I read testStartAdJobTransportAction and testStopAdJobTransportAction, there are no asserts or verification.  Is it correct?

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
```diff
@@ -43,11 +55,26 @@
...
         super.setUp();
+        ClusterService clusterService = mock(ClusterService.class);
+        ClusterSettings clusterSettings = new ClusterSettings(
+            Settings.EMPTY,
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList(AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES)))
+        );
+
+        Settings build = Settings.builder().build();
+        ThreadContext threadContext = new ThreadContext(build);
+        threadContext.putTransient(ConfigConstants.OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT, "alice");
```
kaituo, line 66: do we have code verifying alice has permission to do sth but not others?

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
```diff
@@ -83,6 +105,52 @@ public void testIndexTransportAction() {
...
+        IndexAnomalyDetectorTransportAction transportAction = new IndexAnomalyDetectorTransportAction(
+            mock(TransportService.class),
+            mock(ActionFilters.class),
+            client,
+            clusterService,
+            settings,
+            mock(AnomalyDetectionIndices.class),
+            xContentRegistry()
+        );
+        transportAction.doExecute(task, request, response);
```
kaituo, line 128: any verification after the call?

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
```diff
@@ -83,6 +105,52 @@ public void testIndexTransportAction() {
...
```
kaituo, line 151: I didn't find any any verification in the test case.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
saratvemulapalli: Sure, we can make this private, which actually makes sense as no other callers exist at this point of time.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
saratvemulapalli: Sure we can add that it doesnt hurt, but also we only use this when the user is not null.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
saratvemulapalli: That will not be covered in the unit tests, integration tests will cover them.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
saratvemulapalli: Correct, the asserts are in the listener, when the transport action is called. And the listener is common to all of them.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
saratvemulapalli: Verification is done at line 93 and line 98.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
saratvemulapalli: Same here, the listener gets the verification.
Please see line 93 and line 98.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
saratvemulapalli: Thats a good point. I thought about it as well. Will make a change as part of the integration tests coming up in the next PR.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
kaituo: We can still do it in unit tests by mocking the output of resourceUser.getBackendRoles().

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
kaituo: When you add security plugin, how would the gradle build work when we upgrade versions?  Would the test depend on the availability of the new security plugin version?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
kaituo: yeah, understood.  When others call this method, they have to have the same level of familiarity to the code like you.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
kaituo: Are these the asserts you referred to:

```
response = new ActionListener<AnomalyDetectorJobResponse>() {
            @Override
            public void onResponse(AnomalyDetectorJobResponse adResponse) {
                // Will not be called as there is no detector
                Assert.assertTrue(true);
            }
            }

            @Override
            public void onFailure(Exception e) {
                // Will not be called as there is no detector
                Assert.assertTrue(true);
            }
        };
```

If yes, we always pass the test?

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
kaituo: I see. So we tested 1 branch (index does not exist).  You might plan testing other branches using integration test.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
kaituo: got it.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
kaituo: There can be a race condition for verification. doExecute is an async call and the assert won't be callled (since it would be called by another thread).  So Junit would always declare victory after the doExecute call.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
saratvemulapalli: Sure makes sense.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
saratvemulapalli: Nice catch, onResponse should have been false.

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
saratvemulapalli: Good question. There are 2 parts to answer:
1. How do we test the existing functionality is not broken by changes in AD.
2. How do we test new changes in security plugin that our existing functionality is not broken in AD.

1. We will use the tests where security is enabled as part of our CI. This will pull in the latest docker image and will test what ever is available.
2. The next step is to take the tip of Security Plugin and test our changes against it. I am exploring different options to do this and we will have other dependency issues like our CI being blocked due to security plugin changes. 
Also, doing step 1. allows testing against local clusters which is a good start as part of our release even though the latest docker is not available.

I hope that helps.

kaituo: Yes, we need to use stable release version of security plugin to test so that security plugin bugs won't block us that often as you said.

Also, if new versions are not available during release preparation, we need to skip the test.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/IndexAnomalyDetectorTransportActionTests.java
saratvemulapalli: Had a chat with Kaituo offline. We did see its being run in a single thread. Its probably safe for us to assume that as listener onFailure is being called and no async calls are made.

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyDetectorJobActionTests.java
saratvemulapalli: Definitely, the problem with verifying Alice permissions is that we have to create a detector with a user, then probably verify if alice really has access. We are literally spending a lot of time mocking some parts rather than unit testing, integration tests will cleanly cover them. I understand we dont have it as part of Jacoco for integration tests, I'll spend sometime to get that working, its probably worth the effort.

kaituo: Thanks Sarat.

kaituo (approved):

PR #329, opened 2020-12-11
add AD task and tune detector&AD result data model

ylwu-amzn:
*Issue #, if available:*

## Description of changes:
Currently we only have realtime detector which can take streaming data and detect anomaly in realtime way. For historical detector, user can feed historical data and detect historical anomalies. When user start historical detector, we will start an AD task at backend and run it asynchronously. This PR is to add AD task and tune detector and AD result to support historical detector.

1. AD Task
    When user start a historical detector, we will create a new AD task. We will record task state, progress and detector snapshot in task.

2. Detector changes
    * Add detection date range: user can specify the historical date range to run anomaly detection
    * Add detector type: this is to make the search and aggregation easier. When create detector, user doesn't need to fill this field, we will check `detection_date_range` and `catetory_field` to decide detector type automatically.

3. AD result changes
    * Add task id: user can run historical detector multiple times, and for each run we will create a new AD task. We store task id in AD result, so user can query the latest AD task result with task id.

## Test
1.`./gradlew build`
2.`./gradlew integTest -PnumNodes=3`

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.model;
+
+public enum ADTaskState {
+    CREATED,
+    INIT,
+    RUNNING,
+    FAILED,
+    STOPPED,
+    FINISHED
+}
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java
@@ -15,6 +15,40 @@
 
 package com.amazon.opendistroforelasticsearch.ad.model;
 
+/**
+ * AD task states.
+ * <ul>
+ * <li><code>CREATED</code>:
+ *     When user start a historical detector, we will create one task to track the detector
+ *     execution and set its state as CREATED
+ *
+ * <li><code>INIT</code>:
+ *     After task created, coordinate node will gather all eligible node’s state and dispatch
+ *     task to the worker node with lowest load. When the worker node receives the request,
+ *     it will set the task state as INIT immediately, then start to run cold start to train
+ *     RCF model. We will track the initialization progress in task.
+ *     Init_Progress=ModelUpdates/MinSampleSize
+ *
+ * <li><code>RUNNING</code>:
+ *     If RCF model gets enough data points and passed training, it will start to detect data
+ *     normally and output positive anomaly scores. Once the RCF model starts to output positive
+ *     anomaly score, we will set the task state as RUNNING and init progress as 100%. We will
+ *     track task running progress in task: Task_Progress=DetectedPieces/AllPieces
+ *
+ * <li><code>FINISHED</code>:
+ *     When all historical data detected, we set the task state as FINISHED and task progress
+ *     as 100%.
+ *
+ * <li><code>STOPPED</code>:
+ *     User can cancel a running task by stopping detector, for example, user want to tune
+ *     feature and reran and don’t want current task run any more. When a historical detector
+ *     stopped, we will mark the task flag cancelled as true, when run next piece, we will
+ *     check this flag and stop the task. Then task stopped, will set its state as STOPPED
+ *
+ * <li><code>FAILED</code>:
+ *     If any exception happen, we will set task state as FAILED
+ * </ul>
+ */
 public enum ADTaskState {
     CREATED,
     INIT,
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
@@ -43,7 +43,6 @@ import org.elasticsearch.common.xcontent.XContentBuilder;
 import org.elasticsearch.common.xcontent.XContentParser;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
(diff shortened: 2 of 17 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -596,7 +658,27 @@ public void setUser(User user) {
...
+        return categoryFields != null && categoryFields.size() > 0;
+    }
+
+    public boolean isRealTimeDetector() {
+        return AnomalyDetector.isRealTimeDetector(getDetectionDateRange());
+    }
+
+    private static boolean isRealTimeDetector(DetectionDateRange detectionDateRange) {
+        return detectionDateRange == null || detectionDateRange.getEndTime() == null;
     }
```
yizheliu-amazon, line 683: why implement static method, instead of just using public method like `this.isRealTimeDetector()`?

yizheliu-amazon (approved):
No major concerns from me.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
ylwu-amzn: We need to use this static method to check if a detector is realtime or not before create detector instance. Check here: https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/329/files#diff-ab4bf6eb04e845b21b1182e30afb84c73b282bae5cc26dbcfef2e846211bcdecR475

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java
```diff
@@ -0,0 +1,25 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.model;
+
+public enum ADTaskState {
+    CREATED,
+    INIT,
```
weicongs-amazon, line 20: what's different between created and init?  Is INIT for the model training state?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorType.java
```diff
@@ -0,0 +1,23 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.model;
+
+public enum AnomalyDetectorType {
+    REALTIME_SIGLE_ENTITY,
+    REALTIME_MULTI_ENTITY,
+    HISTORICAL_SIGLE_ENTITY,
```
weicongs-amazon, line 21: typo: SINGLE?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/DetectionDateRange.java
```diff
@@ -0,0 +1,131 @@
...
+        if (startTime == null) {
+            throw new IllegalArgumentException("Detection data range's start time must not be null");
+        }
+        if (endTime == null) {
+            throw new IllegalArgumentException("Detection data range's end time must not be null");
+        }
+    }
+
+    public DetectionDateRange(StreamInput in) throws IOException {
+        this.startTime = in.readOptionalInstant();
```
weicongs-amazon, line 55: what's the reason that there is a null check in the first method, not in the second one?

weicongs-amazon (commented):
just double confirm. for the schema change, we will have a test plan for it, right?  We always need be cautious for any schema related change

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java
ylwu-amzn: **Created**: when user start a historical detector, we will create one task to track the detector execution and set its state as CREATED
**Initializing**: after task created, coordinate node will gather all node’s state and dispatch task to the worker node with lowest load. When the worker node receives the request, it will set the task state as INIT immediately, then start to run cold start to train RCF model.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetectorType.java
ylwu-amzn: Good catch, will fix

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/DetectionDateRange.java
ylwu-amzn: Good catch, should change the `readOptionalInstant` to `readInstant`, and `readInstant` will throw exception if it's null.

ylwu-amzn:
> just double confirm. for the schema change, we will have a test plan for it, right? We always need be cautious for any schema related change

Yes, agree. Have done backward compatibility test based on previous test plans to make sure this change not break current functions.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
yizheliu-amazon: I see. got it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/ADTaskState.java
weicongs-amazon: Thanks for the explanation. we can add some java docs for it later

ylwu-amzn: Good point, will add java docs for these stats.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -596,7 +661,27 @@ public void setUser(User user) {
...
+    private static boolean isMultientityDetector(List<String> categoryFields) {
+        return categoryFields != null && categoryFields.size() > 0;
+    }
+
+    public boolean isRealTimeDetector() {
+        return AnomalyDetector.isRealTimeDetector(getDetectionDateRange());
+    }
+
+    private static boolean isRealTimeDetector(DetectionDateRange detectionDateRange) {
+        return detectionDateRange == null || detectionDateRange.getEndTime() == null;
```
weicongs-amazon, line 685: endTime can't be  null, right?

weicongs-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
ylwu-amzn: Yes, for initial release, we plan to separate historical detector and realtime detector flow, so it's true the endTime can't be null. Next phase we will build universal flow, then the endTime could be null. If the endTime is null, we will pre-train model with historical data from startTime to now, then start realtime detection with the pre-trained model.

PR #331, opened 2020-12-15
Adding support for Security Test Framework

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Adding support for Security tests for AD.
These tests will be run as part of CI and also can be locally tested with a test cluster.

Running locally:
`-Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster" -Dhttps=true -Duser=admin -Dpassword=admin --tests "com.amazon.opendistroforelasticsearch.ad.rest.SecureADRestIT.*"`

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
@@ -1,0 +1,243 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.rest;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import org.apache.http.HttpHost;
+import org.elasticsearch.client.Response;
+import org.elasticsearch.client.RestClient;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+
+import com.amazon.opendistroforelasticsearch.ad.AnomalyDetectorRestTestCase;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.opendistroforelasticsearch.commons.rest.SecureRestClientBuilder;
+
+public class SecureADRestIT extends AnomalyDetectorRestTestCase {
+    String aliceUser = "alice";
+    RestClient aliceClient;
+    String bobUser = "bob";
+    RestClient bobClient;
+    String catUser = "cat";
+    RestClient catClient;
+    String dogUser = "dog";
+    RestClient dogClient;
+
+    @Before
+    public void setupSecureTests() throws IOException {
+        if (!isHttps())
+            return;
+        createIndexRole("index_all_access", "*");
+        createUser(aliceUser, aliceUser, new ArrayList<>(Arrays.asList("odfe")));
+        aliceClient = new SecureRestClientBuilder(getClusterHosts().toArray(new HttpHost[0]), isHttps(), aliceUser, aliceUser)
+            .setSocketTimeout(60000)
+            .build();
+
+        createUser(bobUser, bobUser, new ArrayList<>(Arrays.asList("odfe")));
+        bobClient = new SecureRestClientBuilder(getClusterHosts().toArray(new HttpHost[0]), isHttps(), bobUser, bobUser)
+            .setSocketTimeout(60000)
+            .build();
+
+        createUser(catUser, catUser, new ArrayList<>(Arrays.asList("aes")));
+        catClient = new SecureRestClientBuilder(getClusterHosts().toArray(new HttpHost[0]), isHttps(), catUser, catUser)
+            .setSocketTimeout(60000)
+            .build();
+
+        createUser(dogUser, dogUser, new ArrayList<>(Arrays.asList()));
+        dogClient = new SecureRestClientBuilder(getClusterHosts().toArray(new HttpHost[0]), isHttps(), dogUser, dogUser)
+            .setSocketTimeout(60000)
+            .build();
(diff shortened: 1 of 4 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
```diff
@@ -0,0 +1,243 @@
...
+            Assert.assertEquals(response.getStatusLine().toString(), "HTTP/1.1 500 Internal Server Error");
+        } catch (IOException e) {
+            if (!e.getMessage().contains("no permissions for [cluster:admin/opendistro/ad/detector/jobmanagement]")) {
+                Assert.assertTrue(false);
+            }
+        }
+    }
+
+    public void testStartDetectorForWriteUser() {
+        if (!isHttps())
```
yizheliu-amazon, line 130: minor: we may throw exception or fail the test if `isHttps` is false but SecureTest is running, just in case of wrong CI set up. Wrong CI set up like testing against SecureTest with `isHttp` true may get passing test, but actually test is not being run.

yizheliu-amazon (approved):
Thanks for the change. Make sure we have corresponding CI set up as next step.

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
weicongs-amazon: +1. we can throw exceptions in the @before method, and remove the checks in other places. To make it work, we  might need exclude this test case in the gradle.build

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
```diff
@@ -0,0 +1,243 @@
...
+        try {
+            // User Bob has AD read access, should not be able to modify a detector
+            AnomalyDetector aliceDetector = createRandomAnomalyDetector(false, false, aliceClient);
+            Assert.assertNotNull(aliceDetector.getDetectorId());
+            Response response = startAnomalyDetector(aliceDetector.getDetectorId(), bobClient);
+            Assert.assertEquals(response.getStatusLine().toString(), "HTTP/1.1 500 Internal Server Error");
+        } catch (IOException e) {
+            if (!e.getMessage().contains("no permissions for [cluster:admin/opendistro/ad/detector/jobmanagement]")) {
+                Assert.assertTrue(false);
+            }
```
weicongs-amazon, line 125: not very clear for the goal of this test case.  If the exception is expected, then 121L assert  can't be executed.  If not, then why is this exception is verified?

weicongs-amazon: See a couple of similar logic in other test cases.

saratvemulapalli: Yeah it should always land up in the exception, the idea was just incase it doesn't we should Assert and make sure it doesnt silently exit. Probably what I can do is is that have an assert without a condition.

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
```diff
@@ -0,0 +1,243 @@
...
+import org.elasticsearch.client.RestClient;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+
+import com.amazon.opendistroforelasticsearch.ad.AnomalyDetectorRestTestCase;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.opendistroforelasticsearch.commons.rest.SecureRestClientBuilder;
+
+public class SecureADRestIT extends AnomalyDetectorRestTestCase {
```
ylwu-amzn, line 33: How to run this secure REST integration test? Should we start a local ES cluster with security plugin installed, then run this IT?

saratvemulapalli: Yeah Correct, have security plugin installed and then run these tests. You can run it using `./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster" -Dhttps=true -Duser=admin -Dpassword=admin --tests "com.amazon.opendistroforelasticsearch.ad.rest.SecureADRestIT.*`

This is also run as part of our CI for every build: https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/.github/workflows/CI.yml#L77

saratvemulapalli:
> Thanks for the change. Make sure we have corresponding CI set up as next step.

Sure this already runs as part of the CI today, we are just adding a full new suite of tests.
This is also run as part of our CI for every build: https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/.github/workflows/CI.yml#L77

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
weicongs-amazon: then it's better to use expected exception check. Currently the test case will be possibly passed when no exception is thrown

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
ylwu-amzn: Cool, can you also add this to README doc?

saratvemulapalli: Thats a good point sure.

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
saratvemulapalli: Thats a good point, sure. I'll make that change.

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
saratvemulapalli: Thats a good point, sure. What I can do is exclude this test in build and throw an exception when isHttps is not set in the `@before` method

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
saratvemulapalli: Sent out a PR: https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/334.

weicongs-amazon (approved):

PR #334, opened 2020-12-17
Updating Readme to include Secure tests

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Updating readme to include documentation of Secure tests


weicongs-amazon (approved):

weicongs-amazon:
is there a way to install security plugin, similar as job scheduler one, to run these tests locally without the need of docker image?

saratvemulapalli:
> is there a way to install security plugin, similar as job scheduler one, to run these tests locally without the need of docker image?

Thats a good question, unfortunately ES Test Framework does not support Opendistro Security plugin, from what I hear from the security team is that it collides with X-Pack.

ylwu-amzn (approved):

PR #337, opened 2020-12-20
add AD task cache

ylwu-amzn:
*Issue #, if available:*

## Description of changes:
Add AD task cache. We will put RCF&threshold model, shingle data, threshold model training data in cache. 

##
1.  `./gradlew build`

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.task;
+
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.NUM_MIN_SAMPLES;
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.NUM_SAMPLES_PER_TREE;
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.NUM_TREES;
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.THRESHOLD_MODEL_TRAINING_SIZE;
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.TIME_DECAY;
+
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+import com.amazon.opendistroforelasticsearch.ad.ml.HybridThresholdingModel;
+import com.amazon.opendistroforelasticsearch.ad.ml.ThresholdingModel;
+import com.amazon.opendistroforelasticsearch.ad.model.ADTask;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings;
+import com.amazon.randomcutforest.RandomCutForest;
+
+/**
+ * AD batch task cache which will hold RCF, threshold model, shingle and training data.
+ */
+public class ADBatchTaskCache {
+    private final String detectorId;
+    private RandomCutForest rcfModel;
+    private ThresholdingModel thresholdModel;
+    private boolean thresholdModelTrained;
+    private Deque<Map.Entry<Long, Optional<double[]>>> shingle;
+    private List<Double> thresholdModelTrainingData;
+    private AtomicBoolean cancelled = new AtomicBoolean(false);
+    private AtomicLong cacheMemorySize = new AtomicLong(0);
+    private String cancelReason;
+    private String cancelledBy;
+
+    protected ADBatchTaskCache(ADTask adTask) {
+        this.detectorId = adTask.getDetectorId();
+
+        AnomalyDetector detector = adTask.getDetector();
+        rcfModel = RandomCutForest
+            .builder()
+            .dimensions(detector.getShingleSize() * detector.getEnabledFeatureIds().size())
+            .numberOfTrees(NUM_TREES)
+            .lambda(TIME_DECAY)
+            .sampleSize(NUM_SAMPLES_PER_TREE)
+            .outputAfter(NUM_MIN_SAMPLES)
+            .parallelExecutionEnabled(false)
+            .build();
+
+        this.thresholdModel = new HybridThresholdingModel(
+            AnomalyDetectorSettings.THRESHOLD_MIN_PVALUE,
+            AnomalyDetectorSettings.THRESHOLD_MAX_RANK_ERROR,
+            AnomalyDetectorSettings.THRESHOLD_MAX_SCORE,
+            AnomalyDetectorSettings.THRESHOLD_NUM_LOGNORMAL_QUANTILES,
+            AnomalyDetectorSettings.THRESHOLD_DOWNSAMPLES,
+            AnomalyDetectorSettings.THRESHOLD_MAX_SAMPLES
+        );
+        this.thresholdModelTrainingData = new ArrayList<>(THRESHOLD_MODEL_TRAINING_SIZE);
+        this.thresholdModelTrained = false;
+        this.shingle = new ArrayDeque<>(detector.getShingleSize());
+    }
+
+    protected String getDetectorId() {
+        return detectorId;
+    }
+
+    protected RandomCutForest getRcfModel() {
+        return rcfModel;
+    }
+
+    protected Deque<Map.Entry<Long, Optional<double[]>>> getShingle() {
+        return shingle;
+    }
+
+    protected ThresholdingModel getThresholdModel() {
+        return thresholdModel;
+    }
+
+    protected void setThresholdModelTrained(boolean thresholdModelTrained) {
+        this.thresholdModelTrained = thresholdModelTrained;
+    }
+
+    protected boolean isThresholdModelTrained() {
+        return thresholdModelTrained;
+    }
+
+    protected List<Double> getThresholdModelTrainingData() {
+        return thresholdModelTrainingData;
+    }
+
+    protected AtomicLong getCacheMemorySize() {
+        return cacheMemorySize;
+    }
+
+    protected boolean isCancelled() {
+        return cancelled.get();
+    }
+
+    protected String getCancelReason() {
+        return cancelReason;
+    }
+
+    protected String getCancelledBy() {
+        return cancelledBy;
+    }
+
+    protected void cancel(String reason, String userName) {
+        this.cancelled.compareAndSet(false, true);
+        this.cancelReason = reason;
+        this.cancelledBy = userName;
+    }
+}
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
@@ -22,12 +22,11 @@ import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorS
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.TIME_DECAY;
 
 import java.util.ArrayDeque;
-import java.util.ArrayList;
 import java.util.Deque;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 import com.amazon.opendistroforelasticsearch.ad.ml.HybridThresholdingModel;
@@ -46,7 +45,8 @@ public class ADBatchTaskCache {
     private ThresholdingModel thresholdModel;
     private boolean thresholdModelTrained;
     private Deque<Map.Entry<Long, Optional<double[]>>> shingle;
-    private List<Double> thresholdModelTrainingData;
+    private AtomicInteger thresholdModelTrainingDataSize = new AtomicInteger(0);
+    private double[] thresholdModelTrainingData;
     private AtomicBoolean cancelled = new AtomicBoolean(false);
     private AtomicLong cacheMemorySize = new AtomicLong(0);
     private String cancelReason;
@@ -74,7 +74,7 @@ public class ADBatchTaskCache {
             AnomalyDetectorSettings.THRESHOLD_DOWNSAMPLES,
             AnomalyDetectorSettings.THRESHOLD_MAX_SAMPLES
         );
-        this.thresholdModelTrainingData = new ArrayList<>(THRESHOLD_MODEL_TRAINING_SIZE);
+        this.thresholdModelTrainingData = new double[THRESHOLD_MODEL_TRAINING_SIZE];
         this.thresholdModelTrained = false;
         this.shingle = new ArrayDeque<>(detector.getShingleSize());
     }
@@ -103,8 +103,17 @@ public class ADBatchTaskCache {
         return thresholdModelTrained;
     }
 
-    protected List<Double> getThresholdModelTrainingData() {
+    protected double[] getThresholdModelTrainingData() {
         return thresholdModelTrainingData;
+    }
+
+    protected void clearTrainingData() {
+        this.thresholdModelTrainingData = null;
+        this.thresholdModelTrainingDataSize.set(0);
+    }
+
+    public AtomicInteger getThresholdModelTrainingDataSize() {
+        return thresholdModelTrainingDataSize;
     }
 
     protected AtomicLong getCacheMemorySize() {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
@@ -58,6 +58,9 @@ public class ADTaskCacheManager {
 
     /**
      * Put AD task into cache.
+     * If AD task is already in cache, will throw {@link IllegalArgumentException}
+     * If there is one AD task in cache for detector, will throw {@link IllegalArgumentException}
+     * If there is no enough memory for this AD task, will throw {@link LimitExceededException}
      *
      * @param adTask AD task
      */
@@ -83,6 +86,8 @@ public class ADTaskCacheManager {
     /**
      * check if current running batch task on current node exceeds
      * max running task limitation.
+     * If executing task count exceeds limitation, will throw
+     * {@link LimitExceededException}
      */
     public void checkRunningTaskLimit() {
         if (size() >= maxAdBatchTaskPerNode) {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
@@ -154,7 +154,7 @@ public class ADTaskCacheManager {
             long cacheSize = trainingDataMemorySize(size);
             taskCache.getThresholdModelTrainingData().clear();
             taskCache.getCacheMemorySize().getAndAdd(-cacheSize);
-            memoryTracker.releaseMemory(cacheSize, false, HISTORICAL_SINGLE_ENTITY_DETECTOR);
+            memoryTracker.releaseMemory(cacheSize, true, HISTORICAL_SINGLE_ENTITY_DETECTOR);
         }
     }
 
@@ -220,7 +220,7 @@ public class ADTaskCacheManager {
      */
     public void remove(String taskId) {
         if (contains(taskId)) {
-            memoryTracker.releaseMemory(getBatchTaskCache(taskId).getCacheMemorySize().get(), false, HISTORICAL_SINGLE_ENTITY_DETECTOR);
+            memoryTracker.releaseMemory(getBatchTaskCache(taskId).getCacheMemorySize().get(), true, HISTORICAL_SINGLE_ENTITY_DETECTOR);
             taskCaches.remove(taskId);
         }
     }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
@@ -21,11 +21,11 @@ import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorS
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.THRESHOLD_MODEL_TRAINING_SIZE;
 
 import java.util.Deque;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.common.settings.Settings;
@@ -34,6 +34,7 @@ import com.amazon.opendistroforelasticsearch.ad.MemoryTracker;
 import com.amazon.opendistroforelasticsearch.ad.common.exception.LimitExceededException;
 import com.amazon.opendistroforelasticsearch.ad.ml.ThresholdingModel;
 import com.amazon.opendistroforelasticsearch.ad.model.ADTask;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
 import com.amazon.randomcutforest.RandomCutForest;
 
 public class ADTaskCacheManager {
@@ -125,8 +126,17 @@ public class ADTaskCacheManager {
      * @param taskId AD task id
      * @return threshold model training data
      */
-    public List<Double> getThresholdModelTrainingData(String taskId) {
+    public double[] getThresholdModelTrainingData(String taskId) {
         return getBatchTaskCache(taskId).getThresholdModelTrainingData();
+    }
+
+    public int addThresholdModelTrainingData(String taskId, double... data) {
+        ADBatchTaskCache taskCache = getBatchTaskCache(taskId);
+        double[] thresholdModelTrainingData = taskCache.getThresholdModelTrainingData();
+        AtomicInteger size = taskCache.getThresholdModelTrainingDataSize();
+        int dataPointsAdded = Math.min(data.length, THRESHOLD_MODEL_TRAINING_SIZE - size.get());
+        System.arraycopy(data, 0, thresholdModelTrainingData, size.get(), dataPointsAdded);
+        return size.addAndGet(dataPointsAdded);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+     *
+     * @param adTask AD task
+     */
+    public synchronized void put(ADTask adTask) {
+        String taskId = adTask.getTaskId();
+        if (contains(taskId)) {
+            throw new IllegalArgumentException("AD task is already running");
+        }
+        if (containsTaskOfDetector(adTask.getDetectorId())) {
+            throw new IllegalArgumentException("There is one task executing for detector");
```
weicongs-amazon, line 70: so only one task is allowed per detector, right?

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+     * Get batch task cache. If task doesn't exist in cache, will throw
+     * {@link java.lang.IllegalArgumentException}
+     *
+     * @param taskId task id
+     * @return AD batch task cache
+     */
+    private ADBatchTaskCache getBatchTaskCache(String taskId) {
+        if (!contains(taskId)) {
+            throw new IllegalArgumentException("AD task not in cache");
+        }
```
weicongs-amazon, line 196: do we really need this if this is a private method?

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+     */
+    public void setThresholdModelTrained(String taskId, boolean trained) {
+        ADBatchTaskCache taskCache = getBatchTaskCache(taskId);
+        taskCache.setThresholdModelTrained(trained);
+        if (trained) {
+            int size = taskCache.getThresholdModelTrainingData().size();
+            long cacheSize = trainingDataMemorySize(size);
+            taskCache.getThresholdModelTrainingData().clear();
+            taskCache.getCacheMemorySize().getAndAdd(-cacheSize);
+            memoryTracker.releaseMemory(cacheSize, false, HISTORICAL_SINGLE_ENTITY_DETECTOR);
```
weicongs-amazon, line 152: why do we need clear this up since it's pretty small?

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
```diff
@@ -0,0 +1,131 @@
...
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings;
+import com.amazon.randomcutforest.RandomCutForest;
+
+/**
+ * AD batch task cache which will hold RCF, threshold model, shingle and training data.
+ */
+public class ADBatchTaskCache {
+    private final String detectorId;
+    private RandomCutForest rcfModel;
```
weicongs-amazon, line 45: why do we need put rcf model here? I remember there is rcf model caching in the AD entity result execution.  Are we trying to build a new workflow for this one?

weicongs-amazon (commented):
not sure how this task will be used.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Yes, user can only run one task for one detector. If a detector has one running task, user can't edit/delete/start the detector unless they stop the detector.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: This is to reuse the task existence checking logic, otherwise, need to write duplicate code in other methods.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
ylwu-amzn: For realtime high cardinality(HC) detector, we will cache entity's RCF model in `EntityModel`.  
In our first release, historical detector doesn't support high cardinality. We have two options for the cache change. 

1. Repurpose current cache to support task. The pros is we can reuse some existing code, the cons is we need to change current detector cache logic which may impact realtime detector run.

2. Create separate cache for AD task. So we don't need to worry about the old detector cache design. In next phase, we will not differentiate realtime and historical detectors, so we will create AD tasks for all detectors. Then we can deprecate the old cache mechanism and move to task cache easily. That will be clean and easy to migrate. Of course, the cons will be we will have some duplicate code: cache on detector or entity level, and cache on AD task level.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: It looks good to not clear the training data memory as it's pretty small per detector. But we will let user define max executing historical detectors per node via cluster setting. If there are many detectors executing, the cache size may be not small any more. And the training data size is 1000 currently, if we change it to a bigger value in future, the cache will increase as well. These training data is only to train the threshold model. After model trained, they will not be used any more. So we'd better clean up this cache to release resource.

weicongs-amazon: My understanding is that the model itself contributes much bigger than this training dataset. It's ok to keep this right now since the logic isn't very complicated. My original thought is that this optimization doesn't bring enough benefits to us comparing to added complexity into the codes. Without it, we can keep the cachedMemorySize readonly.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
weicongs-amazon: my point is if this is a really useful exception. I assume this won't happen at all. If this happens, what's the logic like?  Anyway, this doesn't hurt anything.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
weicongs-amazon: thanks. It's good with the second option. btw, is this cachedTask including the requirement of realtime detector, or we just have these two now, and consolidate these later by refactoring?

weicongs-amazon (approved):
LGTM

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
ylwu-amzn: It's too early to include the realtime detector's requirements in this task cache design. Prefer to refactor when we finalize the UX and product features.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Yeah, good question, in normal workflow this exception will not happen. It may happen for some race condition case. For example, user stops a running detector on AD Kibana plugin, then AD ES plugin will try to stop the running AD task in background. It's possible that the frontend polls the running task's state just after the task removed from the cache. Then the task will not be found.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
```diff
@@ -0,0 +1,131 @@
...
+import com.amazon.opendistroforelasticsearch.ad.ml.ThresholdingModel;
+import com.amazon.opendistroforelasticsearch.ad.model.ADTask;
+import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings;
+import com.amazon.randomcutforest.RandomCutForest;
+
+/**
+ * AD batch task cache which will hold RCF, threshold model, shingle and training data.
+ */
+public class ADBatchTaskCache {
```
kaituo, line 43: This class is a combination of 4 functions:
- data structure for models.  You can modify EntityModel to cover that.  Current EntityModel does not support shingle.  Once it supports that, it will need your Deque<Map.Entry<Long, Optional<double[]>>> shingle.
- functions to create empty models.  We create models during training.  Do we really need to create the empty one here?  It sounds strange that a cache is responsible for training or creating models. Can we let the caller pass in the trained objects in?
- cancel/invalidate the cache.
- track cache size

I am thinking whether we can define the class as:

class ADBatchTaskCache {
    private EntityModel ...
    private String cancelReason;
    private AtomicLong cacheMemorySize = new AtomicLong(0);
    ...
    public ADBatchTaskCache(EntityModel..) {..}

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    /**
+     * Get batch task cache. If task doesn't exist in cache, will throw
+     * {@link java.lang.IllegalArgumentException}
+     *
+     * @param taskId task id
+     * @return AD batch task cache
+     */
+    private ADBatchTaskCache getBatchTaskCache(String taskId) {
```
kaituo, line 193: runtime exception means sth you cannot recover.  Is getting something not existing not recoverable? Is it better to return an optional.empty when it is not there.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    /**
+     * Remove task from cache.
+     *
+     * @param taskId AD task id
+     */
+    public void remove(String taskId) {
+        if (contains(taskId)) {
+            memoryTracker.releaseMemory(getBatchTaskCache(taskId).getCacheMemorySize().get(), false, HISTORICAL_SINGLE_ENTITY_DETECTOR);
```
kaituo, line 218: 2nd parameter should be true.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
weicongs-amazon: Same thought, thanks

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
weicongs-amazon: I think if that's a valid scenario, it's better to fix it in our logic, instead of throwing exception.

ylwu-amzn: Yes, for this case, will catch the exception in TaskManger and return readable error message, will not throw exception directly to frontend.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
```diff
@@ -0,0 +1,131 @@
...
+    private final String detectorId;
+    private RandomCutForest rcfModel;
+    private ThresholdingModel thresholdModel;
+    private boolean thresholdModelTrained;
+    private Deque<Map.Entry<Long, Optional<double[]>>> shingle;
+    private List<Double> thresholdModelTrainingData;
+    private AtomicBoolean cancelled = new AtomicBoolean(false);
+    private AtomicLong cacheMemorySize = new AtomicLong(0);
+    private String cancelReason;
+    private String cancelledBy;
```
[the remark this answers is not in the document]
ylwu-amzn: For historical detector, it may run for a long time if the historical date range is long. If user starts a historical detector by mistake, they should stop them before the task reach the historical end time. When user stop a historical detector, AD ES plugin will cancel the running task in background. The cancel task is an async process, so we need to cache the cancel reason and cancelled by in cache until the task cancelled successfully.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+    /**
+     * Estimate max memory usage of model training data.
+     * The training data is double and will cache in {@link java.util.ArrayList}.
+     * Check {@link ADBatchTaskCache#getThresholdModelTrainingData()}
+     *
+     * @param size training data point count
+     * @return how many bytes will consume
+     */
+    public long trainingDataMemorySize(int size) {
+        return 24 * size;
```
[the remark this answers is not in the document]
ylwu-amzn: Training data will be stored in List<Double>, that will be 24 for each item. Will change this when we change to float. Actually we need to change all related places, not just this one.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+        taskCaches = new ConcurrentHashMap<>();
+        this.memoryTracker = memoryTracker;
+    }
+
+    /**
+     * Put AD task into cache.
+     *
+     * @param adTask AD task
+     */
+    public synchronized void put(ADTask adTask) {
```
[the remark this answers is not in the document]
ylwu-amzn: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+        memoryTracker.consumeMemory(neededCacheSize, false, HISTORICAL_SINGLE_ENTITY_DETECTOR);
+        ADBatchTaskCache taskCache = new ADBatchTaskCache(adTask);
+        taskCache.getCacheMemorySize().set(neededCacheSize);
+        taskCaches.put(taskId, taskCache);
+    }
+
+    /**
+     * check if current running batch task on current node exceeds
+     * max running task limitation.
+     */
```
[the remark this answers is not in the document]
ylwu-amzn: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
ylwu-amzn: Good catch, will fix

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+        return getBatchTaskCache(taskId).isThresholdModelTrained();
+    }
+
+    /**
+     * Set threshold model trained or not.
+     *
+     * @param taskId task id
+     * @param trained threshold model trained or not
+     */
+    public void setThresholdModelTrained(String taskId, boolean trained) {
```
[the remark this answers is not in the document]
ylwu-amzn: Correct, only the same thread in which task running can call this method

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
ylwu-amzn: I'd rather not involve EntityModel here, as historical detector is just for realtime detector. We will move to universal flow in next phase. Check response to Weicong's question: https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/337#discussion_r546886774

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: done

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: This is to make the error handling easier in `TaskManager` (I will send out another PR soon). In task manager, we will catch all exceptions, differentiate customer input error and server error in one place. We definitely can return `opotional.empty` or `null` to tell `TaskManger` that this task is not found in cache. But we need extra logic to handle such not found exceptions and in task runner, we need to check task existence for every pieces then remember to return it when the task not found. So I prefer to throw exception here and catch exceptions in one place, rather than check the returned result and check if it's `Opitonal.empty` or `null` in multiple places.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
kaituo: Double is of 8 bytes.  Where does the rest of 16 come from?  I want to make the change easier and the code easier to read as others will have the same questions as me.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
kaituo: Can we guarantee that?  You might want to put synchronized around the method.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+     * Estimate max memory usage of shingle data.
+     * Based on the test, one shingle data point consumes about 96 bytes.
+     * The shingle data is stored in {@link java.util.Deque}
+     * Check {@link ADBatchTaskCache#getShingle()}
+     *
+     * @param shingleSize shingle data point count
+     * @return how many bytes will consume
+     */
+    public long shingleMemorySize(int shingleSize) {
+        return 96 * shingleSize;
```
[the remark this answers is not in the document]
kaituo: you might miss some of the comments.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
kaituo: EntityModel is not specific to HC detectors.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * Cancel AD task.
+     *
+     * @param taskId AD task id
+     * @param reason why need to cancel task
+     * @param userName user name
+     */
+    public void cancel(String taskId, String reason, String userName) {
+        getBatchTaskCache(taskId).cancel(reason, userName);
```
kaituo, line 236: When you cancel a task, should we release memory involved?

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
ylwu-amzn: Will move this method to protected. Prefer not to add synchronized to this method currently as this method will not be called by other places. Other developers may confused by adding unnecessary synchronized.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Refer to this reply https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/337#discussion_r546949902, 
> The cancel task is an async process, so we need to cache the cancel reason and cancelled by in cache until the task cancelled successfully.

The memory will not be released until the task cancelled successfully.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskCache.java
ylwu-amzn: This is some tradeoff, you can check this reply https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/337#discussion_r546886774

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
kaituo: Can you explain these in comments?  In a similar place where Weicong commented, it is uncommon to write get/put by throwing exceptions in this way.  Others will want to know this when they read/modify these methods.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
ylwu-amzn: Good catch, forgot to change this part, will change the calculation.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
ylwu-amzn: Sure, will add more comments to explain.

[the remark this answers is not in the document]
ylwu-amzn: > Double is of 8 bytes.

double is 8 bytes, `Double` object will consumes 16 bytes more.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+        String taskId = adTask.getTaskId();
+        if (contains(taskId)) {
+            throw new IllegalArgumentException("AD task is already running");
+        }
+        if (containsTaskOfDetector(adTask.getDetectorId())) {
+            throw new IllegalArgumentException("There is one task executing for detector");
+        }
+        checkRunningTaskLimit();
+        long neededCacheSize = calculateADTaskCacheSize(adTask);
+        if (!memoryTracker.canAllocate(neededCacheSize)) {
```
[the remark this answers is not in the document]
kaituo: you missed my comment

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,306 @@
...
+            throw new LimitExceededException("No enough memory to run detector");
+        }
+        memoryTracker.consumeMemory(neededCacheSize, false, HISTORICAL_SINGLE_ENTITY_DETECTOR);
```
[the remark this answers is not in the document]
kaituo: you missed my comment

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,325 @@
...
+        return getBatchTaskCache(taskId).isThresholdModelTrained();
+    }
+
+    /**
+     * Set threshold model trained or not.
+     *
+     * @param taskId task id
+     * @param trained threshold model trained or not
+     */
+    protected void setThresholdModelTrained(String taskId, boolean trained) {
```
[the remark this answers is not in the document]
ylwu-amzn: For historical detector, we have all of the data, so we should use these data as much as possible to train and predict better. In realtime detector, we have read some(512) historical sampled data to train model. In historical detector, we use more(1000) data points to train model. For both realtime and historical detectors, the RCF&Threshold model will be continuously trained with following data after cold start. Threshold model will be trained first, then start to output reliable results. So we need to know whether the threshold model is trained or not.

From ML team's suggestion that we should use enough data to train RCF model and they suggest 1000 should be good enough. We may tune this value after performance test and more data quality checking.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,325 @@
...
+    /**
+     * Estimate max memory usage of model training data.
+     * The training data is double and will cache in double array.
+     * One double consumes 8 bytes.
+     *
+     * @param size training data point count
+     * @return how many bytes will consume
+     */
+    public long trainingDataMemorySize(int size) {
+        return 8 * size;
```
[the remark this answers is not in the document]
ylwu-amzn: ok, will add it. 

Not sure why, I can't reply your comments https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/337#discussion_r547615424 , so I just reply it here

> you missed my comment

Working on three branches for historical detector,  feature query validation and this PR. I guess I put the fix into a wrong branches. Will address this comment. Thanks.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
ylwu-amzn: This cache will be released when task cancelled, failed or finished normally. I thought this small PR will make the review easier , but seems that make things complicated and cause a lot of questions which can be answered by related code. Will send out a bigger and more self-explained PR next time. BTW, you can refer to our detail design doc for the whole picture.

Not sure why, I totally missed this comment. Will address this.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Oh, I didn't see this comment either as it's resolved before I addressed this. Will add some comments in java doc.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
```diff
@@ -0,0 +1,329 @@
...
+    public boolean containsTaskOfDetector(String detectorId) {
+        return taskCaches.values().stream().filter(v -> Objects.equals(detectorId, v.getDetectorId())).findAny().isPresent();
+    }
+
+    /**
+     * Get batch task cache. If task doesn't exist in cache, will throw
+     * {@link java.lang.IllegalArgumentException}
+     * We throw exception rather than return {@code Optional.empty} or null
+     * here, so don't need to check task existence by writing duplicate null
+     * checking code. All AD task exceptions will be handled in AD task manager.
```
kaituo, line 207: I still suggest we remove the exceptions and use Optional or null. In Java, the convention is to always check conditions whenever possible and not to use exceptions for flow control.  Conditional check is a jump in the byte code while the exception handling is much more complex.  When an exception occurs inside a Java method, the method creates an Exception object and passes the Exception object to the JVM (in Java term, the method "throw" an Exception). The Exception object contains the type of the exception, and the state of the program when the exception occurs. The JVM is responsible for finding an exception handler to process the Exception object. It searches backward through the call stack until it finds a matching exception handler for that particular class of Exception object (in Java term, it is called "catch" the Exception). If the JVM cannot find a matching exception handler in all the methods in the call stack, it terminates the program.

If you don't want to do it, I am fine as well.  Just a note this is not a good practice.

Ref: https://stackoverflow.com/questions/8161042/why-use-an-exception-instead-of-if-else

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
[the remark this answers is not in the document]
kaituo: Understood.  1000 is a trade-off between accuracy and usability. In real time, 128 is the threshold to emit results.

kaituo (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskCacheManager.java
ylwu-amzn: Totally agree to not use exception for flow control. Here the exception is not to control the flow, but rather terminate the task run flow, just like any AD result action to terminate detector job run by throwing exceptions. In next PR I may change the method name to reduce the confusion.

kaituo: got it.  Please change the method name or add comment to make it clearer.

Issue #339, opened 2020-12-23
[BUG] Profile API returns responses when there are still asyn requests running 


kaituo:
**Describe the bug**
A lot of channel closed warning for profile API in ES log.

```
[2020-11-03T17:48:08,373][WARN ][r.suppressed ] [20728416ac03e0c7fe711affc3da20d5] path: /_opendistro/_anomaly_detection/detectors/w3zMe3UBEMQTObK1dLdD/_profile/init_progress,state,error, params: {detectorID=w3zMe3UBEMQTObK1dLdD, type=init_progress,state,error}
java.lang.IllegalStateException: Channel is already closed
    at org.elasticsearch.rest.RestController$ResourceHandlingHttpChannel.close(RestController.java:511) ~[elasticsearch-7.9.1.jar:7.9.1-SNAPSHOT]
    at org.elasticsearch.rest.RestController$ResourceHandlingHttpChannel.sendResponse(RestController.java:504) ~[elasticsearch-7.9.1.jar:7.9.1-SNAPSHOT]
    at org.elasticsearch.rest.action.RestResponseListener.processResponse(RestResponseListener.java:37) ~[elasticsearch-7.9.1.jar:7.9.1-SNAPSHOT]
    at org.elasticsearch.rest.action.RestActionListener.onResponse(RestActionListener.java:47) [elasticsearch-7.9.1.jar:7.9.1-SNAPSHOT]
    at org.elasticsearch.action.support.TransportAction$1.onResponse(TransportAction.java:90) [elasticsearch-7.9.1.jar:7.9.1-SNAPSHOT]
    at org.elasticsearch.action.support.TransportAction$1.onResponse(TransportAction.java:84) [elasticsearch-7.9.1.jar:7.9.1-SNAPSHOT]
    at com.amazon.opendistroforelasticsearch.ad.transport.GetAnomalyDetectorTransportAction$2.accept(GetAnomalyDetectorTransportAction.java:259) [opendistro-anomaly-detection-1.11.0.0.jar:1.11.0.0]
    at com.amazon.opendistroforelasticsearch.ad.transport.GetAnomalyDetectorTransportAction$2.accept(GetAnomalyDetectorTransportAction.java:256) [opendistro-anomaly-detection-1.11.0.0.jar:1.11.0.0]
[...]

PR #340, opened 2020-12-23
Fix the profile API returns prematurely.

kaituo:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/339

*Description of changes:*
MultiResponsesDelegateActionListener helps send multiple requests asynchronously and return one final response altogether. While waiting for all inflight requests, the method respondImmediately and failImmediately can stop waiting and return immediately. While these two methods are convenient, it is easy to misuse them and cause bugs (see https://github.com/opendistro-for-elasticsearch/anomaly-detection/issues/339 for example). This PR removes the method respondImmediately and failImmediately and refactor profile runner to avoid using them.

This PR also stops printing out the unknown entity state since it is not useful.

Testing done:
1. Added unit tests to verify the bug fix.
2. Manual tests to run profile calls for single-stream and multi-entity detectors for different phases of the detector lifecycle (disabled, init, running). Verified profile results make sense.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
@@ -99,7 +99,6 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
             listener.onFailure(new InvalidParameterException(CommonErrorMessages.EMPTY_PROFILES_COLLECT));
             return;
         }
-
         calculateTotalResponsesToWait(detectorId, profilesToCollect, listener);
     }
 
@@ -118,10 +117,38 @@ public class AnomalyDetectorProfileRunner extends AbstractProfileRunner {
                 ) {
                     ensureExpectedToken(XContentParser.Token.START_OBJECT, xContentParser.nextToken(), xContentParser);
                     AnomalyDetector detector = AnomalyDetector.parse(xContentParser, detectorId);
+
+                    prepareProfile(detector, listener, profilesToCollect);
+                } catch (Exception e) {
+                    listener.onFailure(new RuntimeException(CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG + detectorId, e));
+                }
+            } else {
+                listener.onFailure(new RuntimeException(CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG + detectorId));
+            }
+        }, exception -> listener.onFailure(new RuntimeException(CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG + detectorId, exception))));
+    }
+
+    private void prepareProfile(
+        AnomalyDetector detector,
+        ActionListener<DetectorProfile> listener,
+        Set<DetectorProfileName> profilesToCollect
+    ) {
+        String detectorId = detector.getDetectorId();
+        GetRequest getRequest = new GetRequest(ANOMALY_DETECTOR_JOB_INDEX, detectorId);
+        client.get(getRequest, ActionListener.wrap(getResponse -> {
+            if (getResponse != null && getResponse.isExists()) {
+                try (
+                    XContentParser parser = XContentType.JSON
+                        .xContent()
+                        .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, getResponse.getSourceAsString())
+                ) {
+                    ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
+                    AnomalyDetectorJob job = AnomalyDetectorJob.parse(parser);
+                    long enabledTimeMs = job.getEnabledTime().toEpochMilli();
+
                     boolean isMultiEntityDetector = detector.isMultientityDetector();
 
                     int totalResponsesToWait = 0;
-
                     if (profilesToCollect.contains(DetectorProfileName.ERROR)) {
                         totalResponsesToWait++;
                     }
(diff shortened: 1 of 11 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -261,20 +258,19 @@ private void profileEntityStats(MultiResponsesDelegateActionListener<DetectorPro
                 DetectorProfile.Builder profileBuilder = new DetectorProfile.Builder();
                 DetectorProfile profile = profileBuilder.totalEntities(value).build();
                 listener.onResponse(profile);
-            }, searchException -> { listener.failImmediately(CommonErrorMessages.FAIL_TO_GET_TOTAL_ENTITIES + detector.getDetectorId()); })
-            );
+            }, searchException -> {
+                logger.warn(CommonErrorMessages.FAIL_TO_GET_TOTAL_ENTITIES + detector.getDetectorId());
```
yizheliu-amazon, line 262: why use `logger.warn` instead of `logger.error`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/EntityProfileRunner.java
```diff
@@ -184,6 +164,25 @@ private void getJob(
...
+                        totalResponsesToWait++;
+                    }
+                    if (profilesToCollect.contains(EntityProfileName.MODELS)) {
+                        totalResponsesToWait++;
+                    }
+                    MultiResponsesDelegateActionListener<EntityProfile> delegateListener =
+                        new MultiResponsesDelegateActionListener<EntityProfile>(
+                            listener,
+                            totalResponsesToWait,
+                            "Fail to fetch profile for " + entityValue + " of detector " + detectorId,
```
yizheliu-amazon, line 182: may replace with `FAIL_FETCH_ERR_MSG`

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/EntityProfile.java
```diff
@@ -214,7 +214,7 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         if (modelProfile != null) {
             builder.field(CommonName.MODEL, modelProfile);
         }
-        if (state != null) {
+        if (state != EntityState.UNKNOWN) {
```
yizheliu-amazon, line 217: is it possible for state to be `null`? Do we still need to keep checking if `state is null`?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: It is possible that users haven't created their source index yet.  Since it is a user error, I make it warn instead of error.

src/main/java/com/amazon/opendistroforelasticsearch/ad/EntityProfileRunner.java
kaituo: replaced

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/EntityProfile.java
kaituo: It is possible.  Added back.

yizheliu-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
```diff
@@ -118,10 +117,38 @@ private void calculateTotalResponsesToWait(
                 ) {
                     ensureExpectedToken(XContentParser.Token.START_OBJECT, xContentParser.nextToken(), xContentParser);
                     AnomalyDetector detector = AnomalyDetector.parse(xContentParser, detectorId);
+
+                    prepareProfile(detector, listener, profilesToCollect);
+                } catch (Exception e) {
+                    listener.onFailure(new RuntimeException(CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG + detectorId, e));
+                }
+            } else {
+                listener.onFailure(new RuntimeException(CommonErrorMessages.FAIL_TO_FIND_DETECTOR_MSG + detectorId));
```
ylwu-amzn, line 126: Why not use `AnomalyDetectionException` here?

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/EntityProfile.java
```diff
@@ -214,7 +214,7 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         if (modelProfile != null) {
             builder.field(CommonName.MODEL, modelProfile);
         }
-        if (state != null) {
+        if (state != null && state != EntityState.UNKNOWN) {
```
ylwu-amzn, line 217: When will the entity state be `UNKNOWN `

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: I can do it.  Most of transport APIs use AnomalyDetectionException (except the recent ones added by Sarat).  When I reviewed Sarat's PRs, I thought about pointing it out, but didn't because changing to use AnomalyDetectionException does not add too much benefit except that we have standardized the exceptions we throw.  Our public APIs do not standardize the exception it throws back to the user.   Take profile API as an example, sometimes it throws IOException, sometimes NullPointerException, and sometimes RuntimeException.  Do you think we should change all of public APIs and transport APIs to use AnomalyDetectionException?  The benefit is that we can have a standard wrapper exception to throw.  The drawback is that this might be another PR due to the large changes.

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/EntityProfile.java
kaituo: When we cannot get an entity's state.  For example, a user wants to profile an entity's state.  If we cannot find the associated detector's job, I will mark the state as unknown since I find no way to prove it is in INIT or RUNNING.

public enum EntityState {
    UNKNOWN,
    INIT,
    RUNNING
}

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
ylwu-amzn: Currently we only catch exceptions in AD realtime job and count in failure stats. We may need to count all exceptions from other places, then we need to wrap the exception.
No so urgent. We can fix it later.

ylwu-amzn (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorProfileRunner.java
kaituo: Make sense.

PR #341, opened 2020-12-24
filter out exceptions which should not be counted in failure stats

ylwu-amzn:
*Issue #, if available:*

## Description of changes:

1. Catch wrong feature query exception and stop detector
2. Some exceptions are caused by customer input, which not related with the code logic. Will filter out such exceptions so we can get the failure stats which caused by server error.

## Test
1. `./gradlew build`
2. `./gradlew integTest -PnumNodes=3 `
3. Start local cluster and test

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #347, opened 2020-12-29
backport change: filter out exceptions which should not be counted in failure stats

ylwu-amzn:
*Issue #, if available:*

Backport this change: https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/341

## Description of changes:

1. Catch wrong feature query exception and stop detector
2. Some exceptions are caused by customer input, which not related with the code logic. Will filter out such exceptions so we can get the failure stats which caused by server error.

## Test
1. `./gradlew build`
2. `./gradlew integTest -PnumNodes=3 `
3. Start local cluster and test

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportActionTests.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportActionTests.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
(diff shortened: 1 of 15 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportActionTests.java
```diff
@@ -0,0 +1,174 @@
...
+        super.setUp();
+        testIndex = "test_data";
+        testDataTimeStamp = Instant.now();
+        start = testDataTimeStamp.minus(10, ChronoUnit.MINUTES).toEpochMilli();
+        end = testDataTimeStamp.plus(10, ChronoUnit.MINUTES).toEpochMilli();
+        ingestTestData();
+    }
+
+    private void ingestTestData() throws IOException {
+        String mappings = "{\"properties\":{\"timestamp\":{\"type\":\"date\",\"format\":\"strict_date_time||epoch_millis\"},"
```
kaituo, line 59: Do we have tests for HC detectors?

ylwu-amzn: Tested both single entity and HC detectors on test cluster. But not include HC detectors IT cases. Will send out another PR to add HC detector IT cases.

yizheliu-amazon (approved):

src/test/java/com/amazon/opendistroforelasticsearch/ad/transport/AnomalyResultTransportActionTests.java
ylwu-amzn: Add more IT cases for HC detectors in this PR https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/348, will backport once the PR approved.

kaituo (approved):

PR #355, opened 2021-01-05
start historical detector

ylwu-amzn:
*Issue #, if available:*

## Description of changes:
If a detector has detection date range, we call it "historical detector".

Use the same start detector API to start both realtime and historical detector. 
1. We don't change realtime detector execution logic.
2. For historical detector, we will start an AD task and return the task id to client, then dispatch the task to a node with least load. The AD task will be split into smaller pieces, one piece has 1000 detection intervals by default, user can tune the piece size dynamically. AD task runner will execute the pieces sequentially. 

### AD task states
ad_workbench_task_lifecycle (2) (1)

## Test
1. `./gradlew build`

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -470,7 +470,7 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
             )
             .put(
                 StatNames.ANOMALY_DETECTION_STATE_STATUS.getName(),
-                new ADStat<>(true, new IndexStatusSupplier(indexUtils, DetectorInternalState.DETECTOR_STATE_INDEX))
+                new ADStat<>(true, new IndexStatusSupplier(indexUtils, CommonName.DETECTION_STATE_INDEX))
             )
             .put(StatNames.DETECTOR_COUNT.getName(), new ADStat<>(true, new SettableSupplier()))
             .put(StatNames.HISTORICAL_SINGLE_ENTITY_DETECTOR_COUNT.getName(), new ADStat<>(true, new SettableSupplier()))
@@ -706,7 +706,7 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
                         new SystemIndexDescriptor(AnomalyDetector.ANOMALY_DETECTORS_INDEX, "detector definition"),
                         new SystemIndexDescriptor(AnomalyDetectorJob.ANOMALY_DETECTOR_JOB_INDEX, "detector job"),
                         new SystemIndexDescriptor(CommonName.CHECKPOINT_INDEX_NAME, "model checkpoint"),
-                        new SystemIndexDescriptor(DetectorInternalState.DETECTOR_STATE_INDEX, "detector information like total rcf updates")
+                        new SystemIndexDescriptor(CommonName.DETECTION_STATE_INDEX, "detector information like total rcf updates")
                     )
             );
     }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
@@ -107,6 +107,13 @@ import com.amazon.opendistroforelasticsearch.ad.stats.suppliers.CounterSupplier;
 import com.amazon.opendistroforelasticsearch.ad.stats.suppliers.IndexStatusSupplier;
 import com.amazon.opendistroforelasticsearch.ad.stats.suppliers.ModelsOnNodeSupplier;
 import com.amazon.opendistroforelasticsearch.ad.stats.suppliers.SettableSupplier;
+import com.amazon.opendistroforelasticsearch.ad.task.ADBatchTaskRunner;
+import com.amazon.opendistroforelasticsearch.ad.task.ADTaskCacheManager;
+import com.amazon.opendistroforelasticsearch.ad.task.ADTaskManager;
+import com.amazon.opendistroforelasticsearch.ad.transport.ADBatchAnomalyResultAction;
+import com.amazon.opendistroforelasticsearch.ad.transport.ADBatchAnomalyResultTransportAction;
+import com.amazon.opendistroforelasticsearch.ad.transport.ADBatchTaskRemoteExecutionAction;
+import com.amazon.opendistroforelasticsearch.ad.transport.ADBatchTaskRemoteExecutionTransportAction;
 import com.amazon.opendistroforelasticsearch.ad.transport.ADResultBulkAction;
 import com.amazon.opendistroforelasticsearch.ad.transport.ADResultBulkTransportAction;
 import com.amazon.opendistroforelasticsearch.ad.transport.ADStatsNodesAction;
@@ -148,6 +155,7 @@ import com.amazon.opendistroforelasticsearch.ad.transport.StopDetectorTransportA
 import com.amazon.opendistroforelasticsearch.ad.transport.ThresholdResultAction;
 import com.amazon.opendistroforelasticsearch.ad.transport.ThresholdResultTransportAction;
 import com.amazon.opendistroforelasticsearch.ad.transport.handler.AnomalyIndexHandler;
+import com.amazon.opendistroforelasticsearch.ad.transport.handler.AnomalyResultBulkIndexHandler;
 import com.amazon.opendistroforelasticsearch.ad.transport.handler.DetectionStateHandler;
 import com.amazon.opendistroforelasticsearch.ad.transport.handler.MultiEntityResultHandler;
 import com.amazon.opendistroforelasticsearch.ad.util.ClientUtil;
@@ -173,7 +181,9 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
 
     public static final String AD_BASE_URI = "/_opendistro/_anomaly_detection";
     public static final String AD_BASE_DETECTORS_URI = AD_BASE_URI + "/detectors";
+    public static final String AD_THREAD_POOL_PREFIX = "opendistro.ad.";
     public static final String AD_THREAD_POOL_NAME = "ad-threadpool";
+    public static final String AD_BATCH_TASK_THREAD_POOL_NAME = "ad-batch-task-threadpool";
     public static final String AD_JOB_TYPE = "opendistro_anomaly_detector";
     private static Gson gson;
     private AnomalyDetectionIndices anomalyDetectionIndices;
@@ -186,6 +196,9 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
     private DiscoveryNodeFilterer nodeFilter;
     private IndexUtils indexUtils;
     private DetectionStateHandler detectorStateHandler;
+    private ADTaskCacheManager adTaskCacheManager;
+    private ADTaskManager adTaskManager;
+    private ADBatchTaskRunner adBatchTaskRunner;
 
     static {
         SpecialPermission.check();
@@ -473,7 +486,7 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
             client,
             settings,
             threadPool,
-            ThrowingConsumerWrapper.throwingConsumerWrapper(anomalyDetectionIndices::initDetectorStateIndex),
+            ThrowingConsumerWrapper.throwingConsumerWrapper(anomalyDetectionIndices::initDetectionStateIndex),
             anomalyDetectionIndices::doesDetectorStateIndexExist,
             this.clientUtil,
             this.indexUtils,
@@ -491,6 +504,35 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
             this.indexUtils,
             clusterService,
             stateManager
+        );
+
+        adTaskCacheManager = new ADTaskCacheManager(settings, clusterService, memoryTracker);
+        adTaskManager = new ADTaskManager(settings, clusterService, client, xContentRegistry, anomalyDetectionIndices);
+        AnomalyResultBulkIndexHandler anomalyResultBulkIndexHandler = new AnomalyResultBulkIndexHandler(
+            client,
+            settings,
+            threadPool,
+            ThrowingConsumerWrapper.throwingConsumerWrapper(anomalyDetectionIndices::initAnomalyResultIndexDirectly),
+            anomalyDetectionIndices::doesAnomalyResultIndexExist,
+            this.clientUtil,
+            this.indexUtils,
+            clusterService,
+            anomalyDetectionIndices
+        );
+        adBatchTaskRunner = new ADBatchTaskRunner(
+            settings,
+            threadPool,
+            clusterService,
+            client,
+            nodeFilter,
+            indexNameExpressionResolver,
+            adCircuitBreakerService,
+            featureManager,
+            adTaskManager,
+            anomalyDetectionIndices,
+            adStats,
+            anomalyResultBulkIndexHandler,
+            adTaskCacheManager
         );
 
         // return objects used by Guice to inject dependencies for e.g.,
@@ -517,7 +559,9 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
                 multiEntityResultHandler,
                 checkpoint,
                 modelPartitioner,
-                cacheProvider
+                cacheProvider,
+                adTaskManager,
+                adBatchTaskRunner
             );
     }
 
@@ -532,14 +576,21 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
 
     @Override
     public List<ExecutorBuilder<?>> getExecutorBuilders(Settings settings) {
-        return Collections
-            .singletonList(
+        return ImmutableList
+            .of(
                 new FixedExecutorBuilder(
                     settings,
                     AD_THREAD_POOL_NAME,
                     Math.max(1, EsExecutors.allocatedProcessors(settings) / 4),
                     AnomalyDetectorSettings.AD_THEAD_POOL_QUEUE_SIZE,
-                    "opendistro.ad." + AD_THREAD_POOL_NAME
+                    AD_THREAD_POOL_PREFIX + AD_THREAD_POOL_NAME
+                ),
+                new FixedExecutorBuilder(
+                    settings,
+                    AD_BATCH_TASK_THREAD_POOL_NAME,
+                    Math.max(1, EsExecutors.allocatedProcessors(settings) / 8),
+                    AnomalyDetectorSettings.AD_THEAD_POOL_QUEUE_SIZE,
+                    AD_THREAD_POOL_PREFIX + AD_BATCH_TASK_THREAD_POOL_NAME
                 )
             );
     }
@@ -571,7 +622,10 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
                 AnomalyDetectorSettings.MAX_PRIMARY_SHARDS,
                 AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES,
                 AnomalyDetectorSettings.MAX_CACHE_MISS_HANDLING_PER_SECOND,
-                AnomalyDetectorSettings.MAX_BATCH_TASK_PER_NODE
+                AnomalyDetectorSettings.MAX_BATCH_TASK_PER_NODE,
+                AnomalyDetectorSettings.BATCH_TASK_PIECE_INTERVAL_SECONDS,
+                AnomalyDetectorSettings.MAX_OLD_AD_TASK_DOCS_PER_DETECTOR,
+                AnomalyDetectorSettings.BATCH_TASK_PIECE_SIZE
             );
         return unmodifiableList(Stream.concat(enabledSetting.stream(), systemSetting.stream()).collect(Collectors.toList()));
     }
@@ -613,7 +667,9 @@ public class AnomalyDetectorPlugin extends Plugin implements ActionPlugin, Scrip
                 new ActionHandler<>(ADResultBulkAction.INSTANCE, ADResultBulkTransportAction.class),
                 new ActionHandler<>(EntityResultAction.INSTANCE, EntityResultTransportAction.class),
                 new ActionHandler<>(EntityProfileAction.INSTANCE, EntityProfileTransportAction.class),
-                new ActionHandler<>(SearchAnomalyDetectorInfoAction.INSTANCE, SearchAnomalyDetectorInfoTransportAction.class)
+                new ActionHandler<>(SearchAnomalyDetectorInfoAction.INSTANCE, SearchAnomalyDetectorInfoTransportAction.class),
+                new ActionHandler<>(ADBatchAnomalyResultAction.INSTANCE, ADBatchAnomalyResultTransportAction.class),
+                new ActionHandler<>(ADBatchTaskRemoteExecutionAction.INSTANCE, ADBatchTaskRemoteExecutionTransportAction.class)
             );
     }
 
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/ADTaskCancelledException.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/ADTaskCancelledException.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.ad.common.exception;
+
+public class ADTaskCancelledException extends AnomalyDetectionException {
+    private String cancelledBy;
+
+    public ADTaskCancelledException(String msg, String user) {
+        super(msg);
+        this.cancelledBy = user;
+        this.countedInStats(false);
+    }
+
+    public String getCancelledBy() {
+        return cancelledBy;
+    }
+}
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -23,6 +23,7 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.AbstractMap;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -698,26 +699,16 @@ public class FeatureManager implements CleanState {
         }
     }
 
-    public SinglePointFeatures getShingledFeature(
+    public SinglePointFeatures getShingledFeatureForHistoricalDetector(
         AnomalyDetector detector,
         Deque<Entry<Long, Optional<double[]>>> shingle,
-        Map<Long, Optional<double[]>> dataPoints,
+        Optional<double[]> dataPoint,
         long endTime
     ) {
-        long maxTimeDifference = detector.getDetectorIntervalInMilliseconds() / 2;
-        Map<Long, Entry<Long, Optional<double[]>>> featuresMap = getNearbyPointsForShingle(detector, shingle, endTime, maxTimeDifference)
-            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
-        List<Entry<Long, Long>> missingRanges = getMissingRangesInShingle(detector, featuresMap, endTime);
-        missingRanges.stream().forEach(r -> {
-            if (dataPoints.containsKey(r.getKey())) {
-                featuresMap.put(r.getValue(), new SimpleImmutableEntry<>(r.getValue(), dataPoints.get(r.getKey())));
-            }
-        });
-        shingle.clear();
-
-        getFullShingleEndTimes(endTime, detector.getDetectorIntervalInMilliseconds(), detector.getShingleSize())
-            .mapToObj(time -> featuresMap.getOrDefault(time, new SimpleImmutableEntry<>(time, Optional.empty())))
-            .forEach(e -> shingle.add(e));
+        while (shingle.size() >= detector.getShingleSize()) {
+            shingle.poll();
+        }
+        shingle.add(new AbstractMap.SimpleEntry<>(endTime, dataPoint));
 
         return getProcessedFeatures(shingle, detector, endTime);
     }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
@@ -700,7 +700,7 @@ public class FeatureManager implements CleanState {
         }
     }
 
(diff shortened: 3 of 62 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/common/exception/ADTaskCancelledException.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
ylwu-amzn, line 2: Not sure why, `./gradlew spotlessApply` will reset copyright year as 2020. If change to 2021, `./gradlew build` will fail. Will investigate more.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,63 @@ public int getShingleSize(String detectorId) {
             return -1;
         }
     }
+
+    public void getFeatureDataPoints(
```
yizheliu-amazon, line 703: maybe name it `getFeatureDataPointsByBatch`, `getFeatureDataPoints` sounds like it can be used for any general case.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
```diff
@@ -0,0 +1,444 @@
...
+            } catch (Exception e) {
+                String message = "Failed to start anomaly detector";
+                logger.error(message, e);
+                listener.onFailure(new ElasticsearchStatusException(message, RestStatus.INTERNAL_SERVER_ERROR));
+            }
+        }, exception -> listener.onFailure(exception)));
+    }
+
+    private String validateDetector(AnomalyDetector detector) {
+        if (detector.getFeatureAttributes().size() == 0) {
```
yizheliu-amazon, line 179: is it possible that `getFeatureAttributes()` is null?

src/test/java/com/amazon/opendistroforelasticsearch/ad/plugin/MockReindexPlugin.java
```diff
@@ -0,0 +1,182 @@
...
+import com.google.common.collect.ImmutableList;
+
+public class MockReindexPlugin extends Plugin implements ActionPlugin {
+
+    @Override
+    public List<ActionHandler<? extends ActionRequest, ? extends ActionResponse>> getActions() {
+        return Arrays
+            .asList(
+                new ActionHandler<>(UpdateByQueryAction.INSTANCE, MockTransportUpdateByQueryAction.class),
+                new ActionHandler<>(DeleteByQueryAction.INSTANCE, MockTransportDeleteByQueryAction.class)
```
yizheliu-amazon, line 62: question: how is `MockTransportUpdateByQueryAction ` and `MockTransportDeleteByQueryAction ` being used in test?

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,718 @@
...
+                        .getStatsMap()
+                        .get(AD_EXECUTING_BATCH_TASK_COUNT.getName()))
+                            .compareTo((Long) r2.getStatsMap().get(AD_EXECUTING_BATCH_TASK_COUNT.getName()))
+                )
+                .collect(Collectors.toList());
+
+            if (candidateNodeResponse.size() == 1) {
+                listener.onResponse(candidateNodeResponse.get(0).getNode());
+            } else {
+                // if multiple nodes have same running task count, choose the one with least JVM heap usage.
```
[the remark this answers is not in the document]
ylwu-amzn: JVM heap usage is not so stable, it may change quickly. Some node may have lower JVM heap usage, but it may increase rapidly and historical detector will run for a relatively longer time (10 minutes SLA for 1 month data). It's hard to predict the future JVM heap usage. If a node's JVM heap usage not exceed our limit(85%), then we consider it can run new AD task. We will not dispatch AD task to node which has higher heap usage than 85%. So we'd better distribute AD tasks evenly on data nodes.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,718 @@
...
+
+        client.execute(ADStatsNodesAction.INSTANCE, adStatsRequest, ActionListener.wrap(adStatsResponse -> {
+            List<ADStatsNodeResponse> candidateNodeResponse = adStatsResponse
+                .getNodes()
+                .stream()
+                .filter(stat -> (long) stat.getStatsMap().get(JVM_HEAP_USAGE.getName()) < DEFAULT_JVM_HEAP_USAGE_THRESHOLD)
+                .collect(Collectors.toList());
+
+            if (candidateNodeResponse.size() == 0) {
+                String errorMessage = "All nodes' memory usage exceeds limitation. No eligible node to run detector "
```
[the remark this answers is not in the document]
ylwu-amzn: Good point, will add

src/main/java/com/amazon/opendistroforelasticsearch/ad/indices/AnomalyDetectionIndices.java
```diff
@@ -376,26 +376,33 @@ public void initAnomalyResultIndexDirectly(ActionListener<CreateIndexResponse> a
...
      * @param actionListener action called after create index
-     * @throws IOException IOException from {@link AnomalyDetectionIndices#getDetectorStateMappings}
      */
-    public void initDetectorStateIndex(ActionListener<CreateIndexResponse> actionListener) throws IOException {
-        CreateIndexRequest request = new CreateIndexRequest(DetectorInternalState.DETECTOR_STATE_INDEX)
-            .mapping(AnomalyDetector.TYPE, getDetectorStateMappings(), XContentType.JSON);
-        adminClient.indices().create(request, markMappingUpToDate(ADIndex.STATE, actionListener));
+    public void initDetectionStateIndex(ActionListener<CreateIndexResponse> actionListener) {
+        try {
+            CreateIndexRequest request = new CreateIndexRequest(DetectorInternalState.DETECTOR_STATE_INDEX)
```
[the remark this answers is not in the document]
ylwu-amzn: `DetectorInternalState ` is only for detector's internal state, I think it's ok to keep as is. As `ADTask` and `DetectorInternalState` both defines constants for detection index, will move it to `CommonName`.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,63 @@ public int getShingleSize(String detectorId) {
...
+            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+        List<Entry<Long, Long>> missingRanges = getMissingRangesInShingle(detector, featuresMap, endTime);
+        missingRanges.stream().forEach(r -> {
+            if (dataPoints.containsKey(r.getKey())) {
+                featuresMap.put(r.getValue(), new SimpleImmutableEntry<>(r.getValue(), dataPoints.get(r.getKey())));
+            }
+        });
+        shingle.clear();
+
+        shingle.clear();
```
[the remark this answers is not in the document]
ylwu-amzn: good catch, will remove one.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
ylwu-amzn: sounds a better name, will change

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
ylwu-amzn: Good catch, it's possible. Will add null check.

src/test/java/com/amazon/opendistroforelasticsearch/ad/plugin/MockReindexPlugin.java
ylwu-amzn: `MockReindexPlugin.java` will be loaded by `AnomalyDetectorJobTransportActionTests#getMockPlugins`, then these two transport actions will be registered into test cluster. When we use `UpdateByQueryAction` and `DeleteByQueryAction` in `ADTaskManager`, these two transport actions will be called.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,718 @@
...
+    }
+
+    private ActionListener<String> internalBatchTaskListener(ADTask adTask) {
+        String taskId = adTask.getTaskId();
+        ActionListener<String> listener = ActionListener.wrap(response -> {
+            // If batch task finished normally, remove task from cache and decrease executing task count by 1.
+            adTaskCacheManager.remove(taskId);
+            adStats.getStat(AD_EXECUTING_BATCH_TASK_COUNT.getName()).decrement();
+        }, e -> {
+            // If batch task failed, remove task from cache and decrease executing task count by 1.
```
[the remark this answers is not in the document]
ylwu-amzn: Exception will be logged in `adTaskManager.handleADTaskException(adTask, e)`. 
To make debug easier, I will add a debug level log here.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,718 @@
...
+                                dataEndTime = dataEndTime - dataEndTime % interval;
+                                if ((dataEndTime - dataStartTime) < THRESHOLD_MODEL_TRAINING_SIZE * interval) {
+                                    listener
+                                        .onFailure(
+                                            new AnomalyDetectionException("There is no enough data to train model").countedInStats(false)
+                                        );
+                                    return;
+                                }
+                                long expectedPieceEndTime = dataStartTime + pieceSize * interval;
+                                long firstPieceEndTime = expectedPieceEndTime > dataEndTime ? dataEndTime : expectedPieceEndTime;
```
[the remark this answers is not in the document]
ylwu-amzn: Good point, will change

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
```diff
@@ -211,7 +212,7 @@ public AnomalyDetector(
         this.description = description;
         this.timeField = timeField;
         this.indices = indices;
-        this.featureAttributes = features;
+        this.featureAttributes = features == null ? ImmutableList.of() : ImmutableList.copyOf(features);
```
yizheliu-amazon, line 215: minor: just want to double confirm if [this line](https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java#L235) will make `featureAttributes ` null?

yizheliu-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/model/AnomalyDetector.java
ylwu-amzn: If `featureAttributes ` is null, that line will throw exception. Actually, we can't even write such detector to stream, will throw null pointer exception.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,736 @@
...
+                            ADBatchTaskRemoteExecutionAction.NAME,
+                            new ADBatchAnomalyResultRequest(adTask),
+                            option,
+                            new ActionListenerResponseHandler<>(delegatedListener, ADBatchAnomalyResultResponse::new)
+                        );
+                }
+            }, e -> delegatedListener.onFailure(e))), e -> delegatedListener.onFailure(e)));
+    }
+
+    private void getNodeStats(ADTask adTask, ActionListener<DiscoveryNode> listener) {
```
weicongs-amazon, line 222: This method is to find the right execution node with node stats. it's better to change the method name.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,736 @@
...
+            }
+            candidateNodeResponse = candidateNodeResponse
+                .stream()
+                .sorted(
+                    (ADStatsNodeResponse r1, ADStatsNodeResponse r2) -> ((Long) r1
+                        .getStatsMap()
+                        .get(AD_EXECUTING_BATCH_TASK_COUNT.getName()))
+                            .compareTo((Long) r2.getStatsMap().get(AD_EXECUTING_BATCH_TASK_COUNT.getName()))
+                )
+                .collect(Collectors.toList());
```
weicongs-amazon, line 262: minor: in theory, this sorting isn't needed :). still ok with it since this list size won't be very large

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,736 @@
...
+                listener.onResponse(candidateNodeResponse.get(0).getNode());
+            } else {
+                // if multiple nodes have same running task count, choose the one with least JVM heap usage.
+                Long minTaskCount = (Long) candidateNodeResponse.get(0).getStatsMap().get(AD_EXECUTING_BATCH_TASK_COUNT.getName());
+                Optional<ADStatsNodeResponse> first = candidateNodeResponse
+                    .stream()
+                    .filter(c -> minTaskCount.equals(c.getStatsMap().get(AD_EXECUTING_BATCH_TASK_COUNT.getName())))
+                    .sorted(
+                        (ADStatsNodeResponse r1, ADStatsNodeResponse r2) -> ((Long) r1.getStatsMap().get(JVM_HEAP_USAGE.getName()))
+                            .compareTo((Long) r2.getStatsMap().get(JVM_HEAP_USAGE.getName()))
```
weicongs-amazon, line 274: we  chan change the comparator in the first sorting to avoid this filter and sorting

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,736 @@
...
+        // track AD executing batch task and total batch task execution count
+        adStats.getStat(AD_EXECUTING_BATCH_TASK_COUNT.getName()).increment();
+        adStats.getStat(StatNames.AD_TOTAL_BATCH_TASK_EXECUTION_COUNT.getName()).increment();
+
+        // put AD task into cache
+        adTaskCacheManager.put(adTask);
+
+        // start to run first piece
+        Instant executeStartTime = Instant.now();
+        runFirstPiece(adTask, executeStartTime, internalListener);
```
weicongs-amazon, line 350: Is there a way to make the workflow more clear? I think most of people will have the question of how the other pieces are run since the internalListener will check the whole task status. To get the idea, have to dive deep into the codes. Actually It's  in the getFeatureData method, which is hard to get it with the method name.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,736 @@
...
+                    }
+                } else {
+                    grade = threshold.grade(score);
+                    confidence = threshold.confidence();
+                    if (score > 0) {
+                        threshold.update(score);
+                    }
+                }
+
+                AnomalyResult anomalyResult = new AnomalyResult(
```
weicongs-amazon, line 606: here,  all of anomaly results will be saved, right?

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,736 @@
...
+            long expectedPieceEndTime = pieceStartTime + pieceSize * interval;
+            long pieceEndTime = expectedPieceEndTime > dataEndTime ? dataEndTime : expectedPieceEndTime;
+            int i = 0;
+            while (i < pieceIntervalSeconds) {
+                // check if task cancelled every second, so frontend can get STOPPED state
+                // in 1 second once task cancelled.
+                checkIfADTaskCancelled(taskId);
+                rateLimiter.acquire(1);
+                i++;
+            }
```
weicongs-amazon, line 665: why is this needed for next piece running?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,62 @@ public int getShingleSize(String detectorId) {
...
+
+    public void getFeatureDataPointsByBatch(
+        AnomalyDetector detector,
+        long startTime,
+        long endTime,
+        ActionListener<Map<Long, Optional<double[]>>> listener
+    ) {
+        try {
+            searchFeatureDao.getFeaturesForPeriodByBatch(detector, startTime, endTime, ActionListener.wrap(points -> {
+                logger.info("features size: {}", points.size());
```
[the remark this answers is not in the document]
ylwu-amzn: oh, yeah, will change to debug level

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,62 @@ public int getShingleSize(String detectorId) {
...
+                listener.onResponse(points);
+            }, listener::onFailure));
+        } catch (Exception e) {
+            logger.error("Failed to get features for detector: " + detector.getDetectorId());
```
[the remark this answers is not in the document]
ylwu-amzn: Good catch. We can have 2 options for this exception: 
Option1. It's possible that there are some bad data during some period, so we just ignore this piece and go to the next piece. So we can detect other good pieces, but user may not be able to know this error.  We will return this error and log in task `error` field. Then go to the next piece. There may be many failed pieces, we should limit max errors we can store.

Option2. Just stop the historical detector run if current piece fails. 

The option2 is easier and take less effort, but not good to tolerate bad data. Option 1 is better to handle bad data and always do the best to detect all good data. Will use option 1 to address your comments. Any suggestion?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,62 @@ public int getShingleSize(String detectorId) {
...
+        long maxTimeDifference = detector.getDetectorIntervalInMilliseconds() / 2;
+        Map<Long, Entry<Long, Optional<double[]>>> featuresMap = getNearbyPointsForShingle(detector, shingle, endTime, maxTimeDifference)
+            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+        List<Entry<Long, Long>> missingRanges = getMissingRangesInShingle(detector, featuresMap, endTime);
+        missingRanges.stream().forEach(r -> {
+            if (dataPoints.containsKey(r.getKey())) {
+                featuresMap.put(r.getValue(), new SimpleImmutableEntry<>(r.getValue(), dataPoints.get(r.getKey())));
+            }
+        });
+        shingle.clear();
```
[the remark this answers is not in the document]
ylwu-amzn: The input shingle is cached and will be reused at other places like the task profile API will read this cached shingle to get realtime shingle size.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: Sure, how about rename it as "dispatchTask" ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: If user stop historical detector (stop detector change will be in next PR), we will set its AD task in cache as cancelled. These lines of code is to check if AD task cancelled or not, if the task is cancelled, we will not run the next piece as `checkIfADTaskCancelled ` will throw exception. 

The piece interval is 5seconds by default, but user can tune it to a bigger value, like 60seconds, we don't want user to wait for 60seconds on AD Kibana after clicking the stop button. So this while block will check task cancelled or not every second to make sure we can set the task as cancelled in index as early as possible.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: Yes, will save all anomaly results

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
```diff
@@ -0,0 +1,718 @@
...
+                    );
+                }
+            } catch (Exception e) {
+                listener.onFailure(e);
+            }
+        }, exception -> {
+            logger.error("Fail to execute onFeatureResponseLocalRCF", exception);
+            listener.onFailure(exception);
+        });
+        ThreadedActionListener threadedActionListener = new ThreadedActionListener<>(
```
kaituo, line 524: you miss generic type for ThreadedActionListener

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
```diff
@@ -532,14 +576,21 @@ protected Clock getClock() {
...
+            .of(
                 new FixedExecutorBuilder(
                     settings,
                     AD_THREAD_POOL_NAME,
                     Math.max(1, EsExecutors.allocatedProcessors(settings) / 4),
                     AnomalyDetectorSettings.AD_THEAD_POOL_QUEUE_SIZE,
-                    "opendistro.ad." + AD_THREAD_POOL_NAME
+                    AD_THREAD_POOL_PREFIX + AD_THREAD_POOL_NAME
+                ),
+                new FixedExecutorBuilder(
```
kaituo, line 588: Can we use ScalingExecutor with minimum zero cores since not everyone uses historical analysis?

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -594,7 +603,68 @@ public static boolean checkFilterByBackendRoles(User requestedUser, ActionListen
...
+
+    /**
+     * Generate batch query request for feature aggregation on given date range.
+     *
+     * @param detector anomaly detector
+     * @param startTime start time
+     * @param endTime end time
+     * @param xContentRegistry content registry
+     * @return search source builder
+     * @throws IOException throw IO exception if fail to parse feature aggregation
```
kaituo, line 618: Could you add AnomalyDetectionException to the comment as well?

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,63 @@ public int getShingleSize(String detectorId) {
...
+    }
+
+    public SinglePointFeatures getShingledFeature(
+        AnomalyDetector detector,
+        Deque<Entry<Long, Optional<double[]>>> shingle,
+        Map<Long, Optional<double[]>> dataPoints,
+        long endTime
+    ) {
+        long maxTimeDifference = detector.getDetectorIntervalInMilliseconds() / 2;
+        Map<Long, Entry<Long, Optional<double[]>>> featuresMap = getNearbyPointsForShingle(detector, shingle, endTime, maxTimeDifference)
```
kaituo, line 726: Why do we need to call getNearbyPointsForShingle which helps real time detectors to deal with uneven arrival of requests?  You run in batches and your timestamp within an interval is fixed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,63 @@ public int getShingleSize(String detectorId) {
...
+            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+        List<Entry<Long, Long>> missingRanges = getMissingRangesInShingle(detector, featuresMap, endTime);
```
kaituo, line 728: not sure if the workflow is needed.  You have dataPoints and shingle.  For each interval, you deque once and enqueue once.  Then you call getProcessedFeatures to interpolate.

kaituo (commented):
Only finished one class FeatureManager.  Will continue reviewing.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: We need to put different code logic in listener callback methods to achieve nonblocking call. So there will be multiple nested layers. There is a logical loop here: run first -> piece -> get feature data -> detect anomaly -> run next piece -> get feature data -> detector anomaly -> ...-> all pieces end. As we discussed, will add some TODO here and refactor later.

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
ylwu-amzn: Good point, will change

src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
ylwu-amzn: sure

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
ylwu-amzn: This is to handle sparse data, this method is used for imputing missing points in the shingle with neighboring points

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: sure, will add it

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
```diff
@@ -699,4 +699,63 @@ public int getShingleSize(String detectorId) {
...
+
+        shingle.clear();
+        getFullShingleEndTimes(endTime, detector.getDetectorIntervalInMilliseconds(), detector.getShingleSize())
+            .mapToObj(time -> featuresMap.getOrDefault(time, new SimpleImmutableEntry<>(time, Optional.empty())))
+            .forEach(e -> shingle.add(e));
+
+        return getProcessedFeatures(shingle, detector, endTime);
+    }
+
+    private SinglePointFeatures getProcessedFeatures(
```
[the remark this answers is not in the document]
ylwu-amzn: Yeah, correct. If reuse existing `getProcessedFeatures` method, need to pass in listener.  Actually, we can refactor current realtime detector code to call this new method and remove the old one, no need to pass listener around here. I will add a TODO currently and do the refactor together (as you can see, weicong has some suggestion to refactor too), to avoid bringing in more and more unrelated changed to this PR which is already big enough. BTW, sorry about the big PR.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: Sure will change the first comparator to include this one

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADBatchTaskRunner.java
ylwu-amzn: address with another comment https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/355#discussion_r553010777

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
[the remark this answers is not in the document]
ylwu-amzn: Only two places are using the old method, changed to new method, build can pass.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
ylwu-amzn: Not quite get you. Can you elaborate?

src/main/java/com/amazon/opendistroforelasticsearch/ad/AnomalyDetectorPlugin.java
kaituo: right, ScalingExecutor is more suitable for realtime detectors.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: In getNearbyPointsForShingle, the imputing distance is half of the interval, which does not apply to your case.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
kaituo: I meant your workflow can be:
1) you know the next timestamp t
2) check if dataPoints has a timestamp t
3) if yes, deque the shingle and enque the data point at 2)
4) if no, try interpolation.

The realtime part of the code is hard to understand and not necessary for your use case.

weicongs-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
weicongs-amazon: Do we really need this logic for historical detector?

ylwu-amzn: This is from product/user experience consideration, we will not differentiate the historical and realtime detection, will keep the model/algorithm consistent.

ylwu-amzn: We don't need to impute missing data which caused by run time jitter for historical detector, just need to impute the data hole in source data. 

Discussed with kaituo, will simplify the code for historical detector currently. For the single flow, we can create new function to handle both historical and realtime detection.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
ylwu-amzn: This comment discusses the same issue with this [one](https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/355/files#r553062489).

This function was designed to handle both realtime and historical detector. It looks complicated. Will simplify this function for historical detector. May create new function for single flow later.

src/main/java/com/amazon/opendistroforelasticsearch/ad/feature/FeatureManager.java
weicongs-amazon: seems that we have lots of ideas for the new universal workflow. Should we create a doc/issue to track them?

ylwu-amzn: Yeah, as universal flow is just started, I will doc these ideas on my notebook only now. Don't thinks it will benefit community to share such unconnected ideas on Github now as we don't even mention what universal flow is. Will create an RFC issue later once we finish research, and put those ideas to that Github issue. So user can know the big picture of background, our solution, then they can understand why we come up with these ideas.

kaituo:
Thanks for the change for FeatureManager, Yaliang.  I approve the change to that file.

PR #356, opened 2021-01-06
Adding role based filtering for Preview API

saratvemulapalli:
*Issue #195 *

*Description of changes:*
Added backend role based filtering for Preview API.
Added new integration tests to verify the behavior.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/PreviewAnomalyDetectorTransportAction.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/PreviewAnomalyDetectorTransportAction.java
@@ -15,7 +15,10 @@
 
 package com.amazon.opendistroforelasticsearch.ad.transport;
 
+import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES;
 import static com.amazon.opendistroforelasticsearch.ad.settings.AnomalyDetectorSettings.MAX_ANOMALY_FEATURES;
+import static com.amazon.opendistroforelasticsearch.ad.util.ParseUtils.getUserContext;
+import static com.amazon.opendistroforelasticsearch.ad.util.ParseUtils.resolveUserAndExecute;
 import static org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/PreviewAnomalyDetectorTransportAction.java
```diff
@@ -67,16 +74,39 @@ public PreviewAnomalyDetectorTransportAction(
...
         this.xContentRegistry = xContentRegistry;
         maxAnomalyFeatures = MAX_ANOMALY_FEATURES.get(settings);
         clusterService.getClusterSettings().addSettingsUpdateConsumer(MAX_ANOMALY_FEATURES, it -> maxAnomalyFeatures = it);
+        filterByEnabled = AnomalyDetectorSettings.FILTER_BY_BACKEND_ROLES.get(settings);
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(FILTER_BY_BACKEND_ROLES, it -> filterByEnabled = it);
     }
 
     @Override
     protected void doExecute(Task task, PreviewAnomalyDetectorRequest request, ActionListener<PreviewAnomalyDetectorResponse> listener) {
+        String detectorId = request.getDetectorId();
```
ylwu-amzn, line 89: Detector id could be null. No need to call `resolveUserAndExecute` if detector id is null.

saratvemulapalli: Thats a good point, sure.

kaituo (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/PreviewAnomalyDetectorTransportAction.java
ylwu-amzn: Sorry, I though the detector id was moved to request body. As we discussed, the detector id id part of request URL, we can suppose it always not null. You can keep the code as is.

ylwu-amzn (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/PreviewAnomalyDetectorTransportAction.java
saratvemulapalli: Had a quick chat with @ylwu-amzn. 
Detector ID will not be null its its part of the request path. 
Will leave this as it is to cover a case where user puts in a different detector ID which he does not own, we will be rejecting the request.

ylwu-amzn: Currently we have 2 steps: 
1.Configure data source basic info like index, interval, window delay, then save it to index, so we will get detector id 
2.Configure features, user can call preview API before feature changes saved to index.

In future, we may change to 1 step, configure data source basic info and feature on the same page, then user can call preview API before saving the detector to index, so we will not get detector id. Then we should move detector id into request body. We can refactor code when we make the change from frontend.

saratvemulapalli: Sure makes sense. Thats probably better for the customer.

PR #362, opened 2021-01-20
update/delete historical detector;search AD tasks

ylwu-amzn:
## Description of changes:
1. When user update historical detector, check if there is running task, and user can't switch between realtime and historical.
2. When user delete historical detector, check if there is running task.
3. Add. new search API to support searching AD tasks.

## Test
1. `./gradlew build`

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
@@ -237,14 +237,14 @@ public class IndexAnomalyDetectorActionHandler {
             }
 
             if (existingDetector.isRealTimeDetector()) {
-                validateHCDetector(existingDetector);
+                validateDetector(existingDetector);
             } else {
                 adTaskManager.getLatestADTask(detectorId, (adTask) -> {
                     if (adTask.isPresent() && !adTaskManager.isADTaskEnded(adTask.get())) {
                         // can't update detector if there is AD task running
                         listener.onFailure(new ElasticsearchStatusException("Detector is running", RestStatus.INTERNAL_SERVER_ERROR));
                     } else {
-                        // TODO: change to validateHCDetector method when we support HC historical detector
+                        // TODO: change to validateDetector method when we support HC historical detector
                         searchAdInputIndices(detectorId);
                     }
                 }, transportService, listener);
@@ -257,7 +257,7 @@ public class IndexAnomalyDetectorActionHandler {
 
     }
 
-    private void validateHCDetector(AnomalyDetector existingDetector) {
+    private void validateDetector(AnomalyDetector existingDetector) {
         if (!hasCategoryField(existingDetector) && hasCategoryField(this.anomalyDetector)) {
             validateAgainstExistingMultiEntityAnomalyDetector(detectorId);
         } else {
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
@@ -16,7 +16,6 @@
 package com.amazon.opendistroforelasticsearch.ad.rest.handler;
 
 import static com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector.ANOMALY_DETECTORS_INDEX;
-import static com.amazon.opendistroforelasticsearch.ad.util.ExceptionUtil.getShardsFailure;
 import static com.amazon.opendistroforelasticsearch.ad.util.RestHandlerUtils.XCONTENT_WITH_TYPE;
 import static org.elasticsearch.common.xcontent.XContentParserUtils.ensureExpectedToken;
 
@@ -45,6 +44,7 @@ import org.elasticsearch.action.search.SearchRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.action.support.IndicesOptions;
 import org.elasticsearch.action.support.WriteRequest;
+import org.elasticsearch.action.support.replication.ReplicationResponse;
 import org.elasticsearch.client.Client;
 import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.common.unit.TimeValue;
@@ -57,10 +57,12 @@ import org.elasticsearch.index.query.QueryBuilders;
 import org.elasticsearch.rest.RestRequest;
 import org.elasticsearch.rest.RestStatus;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
+import org.elasticsearch.transport.TransportService;
 
 import com.amazon.opendistroforelasticsearch.ad.constant.CommonName;
 import com.amazon.opendistroforelasticsearch.ad.indices.AnomalyDetectionIndices;
 import com.amazon.opendistroforelasticsearch.ad.model.AnomalyDetector;
+import com.amazon.opendistroforelasticsearch.ad.task.ADTaskManager;
 import com.amazon.opendistroforelasticsearch.ad.transport.IndexAnomalyDetectorResponse;
 import com.amazon.opendistroforelasticsearch.ad.util.RestHandlerUtils;
 import com.amazon.opendistroforelasticsearch.commons.authuser.User;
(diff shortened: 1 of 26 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
```diff
@@ -952,6 +952,26 @@ public ADTaskCancellationState cancelLocalTaskByDetectorId(String detectorId, St
...
+     */
+    public void deleteADTasks(String detectorId, Consumer consumer, ActionListener<DeleteResponse> listener) {
+        DeleteByQueryRequest request = new DeleteByQueryRequest(CommonName.DETECTION_STATE_INDEX);
+
+        BoolQueryBuilder query = new BoolQueryBuilder();
+        query.filter(new TermQueryBuilder(DETECTOR_ID_FIELD, detectorId));
+
+        request.setQuery(query);
+        client.execute(DeleteByQueryAction.INSTANCE, request, ActionListener.wrap(r -> {
+            logger.info("AD tasks deleted for detector {}", detectorId);
```
weicongs-amazon, line 970: just confirm: is this info or debug log? btw, is there a guideline for info, debug, warn logs in odfe?  I rarely see any info log in our repo.

weicongs-amazon (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
ylwu-amzn: **Debug level**: useful for developer to debug error/bug, but not helpful for other user when use this feature, or general user should not concern about these information. For example, the task coordinating node and work node load balancing logic should work correctly if no code bug, and general user should not concern about this. So we put these logic on debug level to avoid unnecessary logs. But if any code bug, developer can open debug level log and find more details.

**Info level**: useful/visible to user to know what steps executed and what's the state now. I think it is useful to let user know the task deleted successfully after they call the delete API, so put it on info level.

**Warn level**: unexpected case happen, but not an error or code bug, like the coordinating node left cluster, this is some tolerable case but not expected normal case, so will write warn log.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/DeleteAnomalyDetectorTransportAction.java
```diff
@@ -92,7 +98,28 @@ protected void doExecute(Task task, DeleteAnomalyDetectorRequest request, Action
...
+                                if (!adTaskManager.isADTaskEnded(adTask.get())) {
+                                    listener
+                                        .onFailure(
+                                            new ElasticsearchStatusException("Detector is running", RestStatus.INTERNAL_SERVER_ERROR)
+                                        );
+                                } else {
+                                    adTaskManager.deleteADTasks(detectorId, r -> deleteDetectorStateDoc(detectorId, listener), listener);
+                                }
+                            } else {
+                                deleteDetectorStateDoc(detectorId, listener);
```
kaituo, line 118: Here I assume one detector has one state doc.  After your changes, this is not true any more.  You need to do a delete by query.  Also, we won't have  real time and historical detector sharing the same detector id, right?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchADTasksTransportAction.java
```diff
@@ -0,0 +1,42 @@
...
+import org.elasticsearch.transport.TransportService;
+
+import com.amazon.opendistroforelasticsearch.ad.transport.handler.ADSearchHandler;
+
+public class SearchADTasksTransportAction extends HandledTransportAction<SearchRequest, SearchResponse> {
+    private ADSearchHandler searchHandler;
+
+    @Inject
+    public SearchADTasksTransportAction(TransportService transportService, ActionFilters actionFilters, ADSearchHandler searchHandler) {
+        super(SearchADTasksAction.NAME, transportService, actionFilters, SearchRequest::new);
```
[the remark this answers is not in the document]
ylwu-amzn: We do permission check on transport layer, if we merge all actions into one class, then we can't control permission separately.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/ADSearchHandler.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private void validateRole(SearchRequest request, User user, ActionListener<SearchResponse> listener) {
+        if (user == null) {
+            // Auth Header is empty when 1. Security is disabled. 2. When user is super-admin
+            // Proceed with search
+            searchDocs(request, listener);
+        } else if (!filterEnabled) {
+            // Security is enabled and filter is disabled
+            // Proceed with search as user is already authenticated to hit this API.
+            searchDocs(request, listener);
```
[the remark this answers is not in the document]
ylwu-amzn: Sure, will merge it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/handler/ADSearchHandler.java
```diff
@@ -0,0 +1,101 @@
...
+                listener.onResponse(searchResponse);
+            }
+
+            @Override
+            public void onFailure(Exception e) {
+                listener.onFailure(e);
+            }
+        });
+    }
+
```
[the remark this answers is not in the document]
ylwu-amzn: Good catch, will replace with input listener

src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/IndexAnomalyDetectorActionHandler.java
```diff
@@ -226,6 +257,14 @@ private void onGetAnomalyDetectorResponse(GetResponse response) {
 
     }
 
+    private void validateHCDetector(AnomalyDetector existingDetector) {
```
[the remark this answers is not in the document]
ylwu-amzn: sure

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
weicongs-amazon: thanks for the details. Totally agree. My question is more like: why there is no info log for start and stop detector in this class, but there is for deletion. When I check the elasticsearch.log I rarely see any info log. I thought there is general guideline for it, such as: no info log for normal indexing or normal operations. If there is, this also need be applied to our new ML plugin.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/DeleteAnomalyDetectorTransportAction.java
ylwu-amzn: The realtime detector task is not added yet. Will add task for realtime detector in universal flow project. Until this PR, the realtime detector still has one detection state.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchADTasksTransportAction.java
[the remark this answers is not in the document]
kaituo: Can we have an abstract class and the subclass just have constructors?  Or is it over-engineering?

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/DeleteAnomalyDetectorTransportAction.java
kaituo: I meant deleteDetectorStateDoc would only delete one doc whose id is the detector id.  Here you want to delete all state docs since multiple task docs share the same detector id.

src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
ylwu-amzn: For start detector, we have info log, check `ADTaskManager#runBatchResultAction` line 834, `logger.info("AD task {} of detector {} dispatched to {} node {}"`

For stop detector, we have warn log, check `ADTaskManager#handleADTaskException` line 876, `logger.warn("AD task cancelled: " + adTask.getTaskId());` , will add detector id in this log.

I don't think we have a special guideline for ES plugin log. Agree that we should not log too much for normal indexing or normal operations, but should have some info log for key steps such as create, start/stop. For the historical detector run, there will be too much log as we need to run many pieces, and will not be very helpful to general user, so just put on debug level. I think this is another consideration for log level to avoid too much log.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/SearchADTasksTransportAction.java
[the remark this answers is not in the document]
ylwu-amzn: Don't think it's so necessary to add an abstract class just to avoid one line of duplicate code `searchHandler.search(request, listener);` here. We can refactor later when more common logic need to abstracted.

src/main/java/com/amazon/opendistroforelasticsearch/ad/transport/DeleteAnomalyDetectorTransportAction.java
ylwu-amzn: Got it. Normally, if the latest task not found, there is not tasks for the historical detector, so no need to delete tasks with detector id. In case the latest task removed or updated as not latest, will change to delete task first. And will handle index not found exception.

kaituo (approved):

PR #372, opened 2021-01-26
remove spotless header file; remove copyright year in new files for h…

ylwu-amzn:
…istorical detector

## Description of changes:
1.Remove copyright year in new files for historical detector
2.Remove hard code spotless header file. 


build.gradle
```diff
@@ -451,7 +451,6 @@ spotless {
     java {
         removeUnusedImports()
         importOrder 'java', 'javax', 'org', 'com'
-        licenseHeaderFile 'spotless.license.java'
```
saratvemulapalli, line 454: Would this mean we would not mandate the license header for new files?
Can't we just remove the year?

ylwu-amzn: The license header is still mandatory. If license header is missing, can't pass gradle build. 
From Aloitta, we can remove copyright header for new files and keep old files as is. So I just removed copyright year in new files for AD historical detector in this PR. We can't specify a license header file without copyright year, otherwise the old files which has copyright year will fail gradle build.

yizheliu-amazon: Got it. Thanks. I guess AD Kibana should follow the same standard.

yizheliu-amazon (approved):

ohltyler (approved):

build.gradle
ohltyler: Created AD Kibana issue [here](https://github.com/opendistro-for-elasticsearch/anomaly-detection-kibana-plugin/issues/365). @ylwu-amzn is confirming on whether or not to change all older files to the new non-year format as well.

saratvemulapalli: Makes sense, thanks @ylwu-amzn !

saratvemulapalli (approved):

build.gradle
ohltyler: It is confirmed to leave the older files as they exist. Changes should only be made based on recently-created files (like historical-detector related code).

ylwu-amzn: From Alolita, “the open source community is used to having older formats in older code” we should not remove copyright year from old files. So we will just remove copyright year in new files from 2021

PR #377, opened 2021-01-29
validate detector only when start detector; fix flaky test case

ylwu-amzn:
*Issue #, if available:*

## Description of changes:
1. Only validate feature when start detector
2. Fix flaky test case

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
@@ -181,10 +181,13 @@ public class ADTaskManager {
         TransportService transportService,
         ActionListener<AnomalyDetectorJobResponse> listener
     ) {
-        getDetector(
-            detectorId,
-            (detector) -> handler.startAnomalyDetectorJob(detector), // run realtime detector
-            (detector) -> {
+        getDetector(detectorId, (detector) -> {
+            if (validateDetector(detector, listener)) {
+                // run realtime detector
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/task/ADTaskManager.java
```diff
@@ -293,12 +295,6 @@ public void stopDetector(
                 ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.nextToken(), parser);
                 AnomalyDetector detector = AnomalyDetector.parse(parser, response.getId(), response.getVersion());
 
-                String error = validateDetector(detector);
-                if (error != null) {
-                    listener.onFailure(new ElasticsearchStatusException(error, RestStatus.BAD_REQUEST));
-                    return;
-                }
-
```
ohltyler, line 301: Is this where the error was being propagated in the response? If so, just curious what is the reason for trying to stop a non-running task? I guess it makes sense if just for consistency to handle different cases, but just wondering if it's an unnecessary step.

ylwu-amzn: These code added for start historical detector change. Forgot to clean up it when made stop and delete historical detector change. No need to validate if detector has enabled feature when stop/delete detector.

ohltyler (approved):
Thanks for the fix!

saratvemulapalli (approved):
Thanks Yaliang for the change and the IT test to cover this.

PR #383, opened 2021-02-04
change the backend role filtering to keep consistent with alerting pl…

ylwu-amzn:
…ugin

## Description of changes:
Now we have two different logics for AD document level access control
1.For search APIs we will add user role filter query. When user is null, will add `user==null` filter, if user is not null, but backend role is null or empty will add `user != null and backend_roles == null or empty` query. So for user who has no backend role, they can still see their detectors after enabling backend role filtering

2.For other APIs like get/update/start/stop, we will check if the detector’s user backend role can match current thread user’s backend role. If either detector’s user backend role or current thread user backend role is null, we will throw no permission exception. That means if user doesn’t configure backend role, after enabling backend role filtering, they will not see their detectors.

For alerting plugin, if user's backend role is empty, the user can't see any monitors. User has to configure backend roles before enabling alerting backend role filtering. We will change to the same way of alerting plugin ([alerting plugin code link](https://github.com/opendistro-for-elasticsearch/alerting/blob/master/core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/elasticapi/ElasticExtensions.kt#L148)).

## Test
1. `./gradlew build`

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
@@ -50,7 +50,6 @@ import org.elasticsearch.common.xcontent.NamedXContentRegistry;
 import org.elasticsearch.common.xcontent.XContentParser;
 import org.elasticsearch.common.xcontent.XContentType;
 import org.elasticsearch.index.query.BoolQueryBuilder;
-import org.elasticsearch.index.query.ExistsQueryBuilder;
 import org.elasticsearch.index.query.NestedQueryBuilder;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
@@ -79,6 +78,7 @@ import com.amazon.opendistroforelasticsearch.ad.rest.handler.AnomalyDetectorFunc
 import com.amazon.opendistroforelasticsearch.ad.transport.GetAnomalyDetectorResponse;
 import com.amazon.opendistroforelasticsearch.commons.ConfigConstants;
 import com.amazon.opendistroforelasticsearch.commons.authuser.User;
+import com.google.common.collect.ImmutableList;
 
 /**
  * Parsing utility functions.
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/ad/util/ParseUtils.java
```diff
@@ -425,31 +425,17 @@ public static SearchSourceBuilder generateEntityColdStartQuery(
...
-
-            boolQueryBuilder.mustNot(nestedQueryBuilder);
-            boolQueryBuilder.must(userExistsNestedQueryBuilder);
-        } else {
-            // For normal case, user should have backend roles.
-            TermsQueryBuilder userRolesFilterQuery = QueryBuilders.termsQuery(userBackendRoleFieldName, user.getBackendRoles());
-            NestedQueryBuilder nestedQueryBuilder = new NestedQueryBuilder(userFieldName, userRolesFilterQuery, ScoreMode.None);
-            boolQueryBuilder.must(nestedQueryBuilder);
-        }
+        List<String> backendRoles = user.getBackendRoles() != null ? user.getBackendRoles() : ImmutableList.of();
```
ylwu-amzn, line 434: Currently AD plugin will parse user from thread context `OPENDISTRO_SECURITY_USER_INFO_THREAD_CONTEXT`, from the [User.parse](https://github.com/opendistro-for-elasticsearch/common-utils/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/commons/authuser/User.java#L181) code, the backend role should be empty list is no backend roles. We recheck here in case we support other case or the user parse method changes in future.

saratvemulapalli (approved):
The changes look good to me. 
I understand the timeline is tight for the 1.13 but can we add integration tests in https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/master/src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java for the search API?

Test cases to cover:
1. Search when backend role filter is enabled and user has no backend role
2. Search when backend role filter is enabled and user has a backend role
3. Search when backend role filter is disabled

ohltyler (approved):
Thanks for the changes

PR #385, opened 2021-02-10
fix failed integration cases

ylwu-amzn:
*Issue #, if available:*

## Description of changes:
Fix flaky integration test cases

## Test
1. Start local cluster with security plugin and run `./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster" -Dhttps=true -Dsecurity=true -Duser=admin -Dpassword=admin`
2. Start local cluster without security plugin and run `./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster" -Dhttps=false` and `./gradlew integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername="docker-cluster"`
3. `./gradlew build`
4. `./gradlew integTest -PnumNodes=3`

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/ad/ODFERestTestCase.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/ad/ODFERestTestCase.java
@@ -87,7 +87,7 @@ public abstract class ODFERestTestCase extends ESRestTestCase {
             // disable the warning exception for admin client since it's only used for cleanup.
             .put("strictDeprecationMode", false)
             .put("http.port", 9200)
-            .put(OPENDISTRO_SECURITY_SSL_HTTP_ENABLED, true)
+            .put(OPENDISTRO_SECURITY_SSL_HTTP_ENABLED, isHttps())
             .put(OPENDISTRO_SECURITY_SSL_HTTP_PEMCERT_FILEPATH, "sample.pem")
             .put(OPENDISTRO_SECURITY_SSL_HTTP_KEYSTORE_FILEPATH, "test-kirk.jks")
             .put(OPENDISTRO_SECURITY_SSL_HTTP_KEYSTORE_PASSWORD, "changeit")
--- a/src/test/java/com/amazon/opendistroforelasticsearch/ad/e2e/DetectionResultEvalutationIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/ad/e2e/DetectionResultEvalutationIT.java
@@ -15,6 +15,8 @@
 
 package com.amazon.opendistroforelasticsearch.ad.e2e;
 
+import static com.amazon.opendistroforelasticsearch.ad.TestHelpers.toHttpEntity;
+
 import java.io.File;
 import java.io.FileReader;
 import java.time.Instant;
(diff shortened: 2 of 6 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/ad/ODFERestTestCase.java
```diff
@@ -87,7 +87,7 @@ protected Settings restAdminSettings() {
             // disable the warning exception for admin client since it's only used for cleanup.
             .put("strictDeprecationMode", false)
```
weicongs-amazon, line 88: currently admin client is only used for cleanup. if we wanna change this, it's better to remove this line. Just have concern that we will use it in the wrong place.

src/test/java/com/amazon/opendistroforelasticsearch/ad/e2e/DetectionResultEvalutationIT.java
```diff
@@ -63,11 +71,10 @@ private void verifyAnomaly(
         List<JsonObject> data = getData(dataFileName);
         List<Entry<Instant, Instant>> anomalies = getAnomalyWindows(labelFileName);
 
-        indexTrainData(datasetName, data, trainTestSplit, client);
+        bulkIndexTrainData(datasetName, data, trainTestSplit, adminClient());
```
ylwu-amzn, line 74: Forgot to clean up this, will change to `client`. Same for line 77

src/test/java/com/amazon/opendistroforelasticsearch/ad/ODFERestTestCase.java
ylwu-amzn: I think you mean this [line](https://github.com/opendistro-for-elasticsearch/anomaly-detection/pull/385/files#diff-56ef50f457914d21259ff0d2b1ec65554ab516efc777a6def1ac7e705dd9dbaeR74) which is using adminClient. Forgot to clean up it, will change this to `client`

src/test/java/com/amazon/opendistroforelasticsearch/ad/rest/SecureADRestIT.java
```diff
@@ -45,194 +45,233 @@
...
-
-        createUser(dogUser, dogUser, new ArrayList<>(Arrays.asList()));
-        dogClient = new SecureRestClientBuilder(getClusterHosts().toArray(new HttpHost[0]), isHttps(), dogUser, dogUser)
-            .setSocketTimeout(60000)
-            .build();
-
-        createRoleMapping("anomaly_read_access", new ArrayList<>(Arrays.asList(bobUser)));
-        createRoleMapping("anomaly_full_access", new ArrayList<>(Arrays.asList(aliceUser, catUser, dogUser)));
-        createRoleMapping("index_all_access", new ArrayList<>(Arrays.asList(aliceUser, bobUser, catUser, dogUser)));
+        if (isHttps()) {
```
saratvemulapalli, line 48: Had an offline chat with @ylwu-amzn .
We intentionally wanted to run these tests if the https flag is set from the arguments.
We set this in build.gradle: https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/main/build.gradle#L168-L172
Probably as we discussed the cleaner way is to add another case of false to the build.gradle script.

ylwu-amzn: Exclude `SecureADRestIT` if https is false in case someone test with "-Dhttps=false"

saratvemulapalli (approved):

weicongs-amazon (approved):

PR #389, opened 2021-02-22
Extract a new class for entity frequenty tracking

kaituo:
*Issue #, if available:*

*Description of changes:*
HC detectors use a 1-pass algorithm for estimating heavy hitters in a stream. Our method maintains a time-decayed count for each entity, which allows us to compare the frequencies of entities from different detectors in the stream. To reuse the code in historical detectors, I created a new class PriorityTracker and moved all related logic there.  When an entity is hit, the caller can call PriorityTracker.updatePriority to update the entity's priority.  The callers can find the most frequently occurring entities in the stream using PriorityTracker.getTopNEntities.

This PR also adds tests for NodeStateManager.

Testing done:
1. manually tested basic workflow of HC detectors still works.
2. added new tests for PriorityTracker.


src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
```diff
@@ -0,0 +1,320 @@
...
+        long periods = (clock.instant().getEpochSecond() - landmarkEpoch) / intervalSecs;
+        return periods >> DECAY_CONSTANT;
+    }
+
+    /**
+     *
+     * @param n the number of entities to return.  Can be less than n if there are not enough entities stored.
+     * @return top entities in the descending order of priority
+     */
+    public List<String> getTopNEntities(int n) {
```
LiuJoyceC, line 312: Is this method going to be used elsewhere eventually? So far it's only called by tests.

LiuJoyceC (approved):
Other than a minor question I had about a currently unused new method, the refactor looks good.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
kaituo: yes, historical detectors are expected to call it.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
```diff
@@ -0,0 +1,320 @@
...
+        node.priority = getUpdatedPriority(node.priority);
+        this.priorityList.add(node);
+    }
+
+    /**
+     * Associate the specified priority with the entity model Id
+     * @param entityModelId Entity model Id
+     * @param priority priority
+     */
+    protected void addPriority(String entityModelId, float priority) {
```
ylwu-amzn, line 222: The initial `priority` of one entity should be 0?

ylwu-amzn: Why need to set the key as `entityModelId` ? Can we just put entity as key?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
```diff
@@ -0,0 +1,320 @@
...
+     * entities' priorities within the same detector.
+     *
+     * Each detector maintains an ordered map, filled by entities's accumulated sum of g(i−L),
+     * which is what this function computes.
+     *
+     * g(n) = e^{0.125n}.  i is current period. L is the landmark: period 0 when the
+     * detector is enabled. i - L measures the elapsed periods since detector starts.
+     * 0.125 is the decay constant.
+     *
+     * Since g(p−L) is changing and they are the same for all entities of the same detector,
```
ylwu-amzn, line 257: Seems all other places in this java doc are using `g(i - L)`,  change `g(p-L)` to `g(i - L)` ?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
kaituo: > The initial priority of one entity should be 0? I see the line 210 is using new PriorityNode(entityModelId, 0f)
Yes

> Why need to set the key as entityModelId ? Can we just put entity as key?
You asked me to use this name :) How about entityId?

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
kaituo: Good catch.  Changed.

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
ylwu-amzn: For realtime HC detector, it stores entity model id (detectorId + "_entity_" + entity ?). For historical HC detector, plan to use just entity value. As we discussed, as detector will have separate PriorityTracker, so we can just use entity, that can save some memory.

kaituo: " as detector will have separate PriorityTracker, so we can just use entity, that can save some memory."  Could you explain this sentence?  I don't understand.

ylwu-amzn (approved):

src/main/java/com/amazon/opendistroforelasticsearch/ad/caching/PriorityTracker.java
ylwu-amzn: For example, we store "<detector_id>+<entity_value>_entity" in cache, we can change to "<entity_value>" to avoid saving duplicate information of "<detector_id>" and "_entity" as the cache is on detector level.

